PR #4, opened 2018-06-24
Merge with elpa

matsl:
@rswgnu PTAL

The merge with elpa was more complicated than I first thought. Some files that were not touched in elpa but got silently merged in overriding the current version in savannah. Very strange but can probably be explained somehow. I have tried to fix these broken merges. All in all it became a lot a of files that were changed so I thought it could be a good idea to try to do a pull request review on this using github to try this out. 

So please have a look at this PR and let me know if it is a way of working that you think is worth using for future work as well.


matsl:
This PR was only a merge of the tip version on savannah with the tip on elpa. So there was no additional functionality. I think it is best to start the merge from scratch now when we have added more to the savannah tip. I will close this PR.

PR #7, opened 2019-03-02
Handle if hui-window is loaded before hui-mouse

matsl:
Hi Bob,

This is Stefans commit from elpa-hyperbole that is not touching on the XEmacs code. It is applied on top of current master so should apply cleanly.

* hmouse-drv.el: Just require `hui-window` without messing with
`hmouse-alist` any more.  Move compiler-silencing `defvar`s outside of
`eval-when-compile`.

* hui-mouse.el: Just use `require` instead of `load`.

* hui-window.el: Set hmouse-alist only after loading hui-mouse.
(hui-window--register): New function.
(hui-mouse): Use with-eval-after-load.

### DIFF
--- a/hmouse-drv.el
+++ b/hmouse-drv.el
@@ -4,7 +4,7 @@
 ;;
 ;; Orig-Date:    04-Feb-90
 ;;
-;; Copyright (C) 1989-2017  Free Software Foundation, Inc.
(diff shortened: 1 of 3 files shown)


rswgnu:
Got it.  I will try to merge today.  I just have to ensure that I also ensure compatibility with the smart-key-reliad function that I use for testing changes to the smart-key alists.

-- Bob

PR #13, opened 2019-08-20
Apply some easy to fix flycheck suggestions

matsl:
Use "? " at end of y-or-n-p
Remove "." at end of error messages
Use imperative voice in docs strings

This fixes the bulk of flycheck messages. There are things left to
correct but they will require more work and that will have to be dealt
with later.

@rswgnu  PTAL - here is the flycheck PR I was talking about.


rswgnu:
Looks good except for change from 2 spaces at the end of sentences to one.  Any way to get rid of that?

matsl:
From your comment I thought the change introduced two spaces at the end of sentences but I see now that those are already there. So the PR does not touch on the double space issue. So you are suggesting that we should update all doc strings to only have one space after a full stop?

I have no big issue with this except that it could go in its own PR not to complicate this.

According to https://www.gnu.org/software/emacs/manual/html_node/emacs/Sentences.html it is also not a recommended setting which makes me lean towards keeping it. Most other doc strings in Emacs would have the double space conventions as well so it would be consistent to keep it.

rswgnu:
No, it looked to me like it changed two spaces to one.  There definitely was some spacing change.  I will have to look again.  I did think it was strange.

-- Bob

matsl:
Hi Bob,

Ok. Great. I misunderstood.

Please have another look. I searched through the diff and found no place with that whitespace change. 

Yours

Robert Weiner <notifications@github.com> skrev: (22 augusti 2019 01:34:27 CEST)
>No, it looked to me like it changed two spaces to one.  There
>definitely was some spacing change.  I will have to look again.  I did
>think it was strange.>
>>
>-- Bob>
>>
>> On Aug 21, 2019, at 2:15 PM, Mats Lidell <notifications@github.com>
>wrote:>
>> >
>> From your comment I thought the change introduced two spaces at the
>end of sentences but I see now that those are already there.  So the PR
>does not touch on the double space issue. So you are suggesting that we
>should update all doc strings to only have one space after a full
>stop?>
>> >
>> I have no big issue with this except that it could go in its own PR
>not to complicate this.>
>> >
>> According to
>https://www.gnu.org/software/emacs/manual/html_node/emacs/Sentences.html
>it is also not a recommended setting which makes me lean towards
>keeping it. Most other doc strings in Emacs would have the double space
>conventions as well so it would be consistent to keep it.>
>> >
>> —>
>> You are receiving this because you were mentioned.>
>> Reply to this email directly, view it on GitHub, or mute the thread.>
>>
>>
>-- >
>You are receiving this because you authored the thread.>
>Reply to this email directly or view it on GitHub:>
>https://github.com/rswgnu/hyperbole/pull/13#issuecomment-523688709

-- 
## Mats

PR #20, opened 2020-01-11
Fail packaging when proper version number is missing

matsl:
@rswgnu The exit code was not handled properly in the Makefile. 

fgreps exit code is 1 when no files without matches are found which make considers as an error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -275,7 +275,7 @@ clean:
 version: doc
 	@ echo ""
 	@ echo "Any fgrep output means the version number has not been updated in that file."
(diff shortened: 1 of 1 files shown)


matsl:
Sorry. I must have been testing back an forth with different settings and pushed the wrong version. My bad.

If you see my original comment I account for the -L option. The change pushed did not check for the right return code though. The test should be like this: 

`fgrep -L $(HYPB_VERSION) Makefile HY-ABOUT HY-NEWS README.md hversion.el hyperbole.el man/hyperbole.texi man/version.texi; [ $$? -eq 1 ] || exit 1`

The reason for the extra check is that fgrep with the -L option returns the exit code 0 if there are files that does not match. In our case that is the error case.

Yours

PR #21, opened 2020-01-24
Use float for delay instead of looping

matsl:
I had to bump up the loop delay to 10000 to see any noticeable flashing on my machine. Looked for alternatives and found out that you can sit for a decimal time on the machines that supports waiting fractions of a second. Would not modern linux and windows machines do that?

@rswgnu PTAL


rswgnu:
Yes, that was just some really old code from when sit-for only took integers.

-- Bob

PR #22, opened 2020-01-30
Use new variable for the flash time to avoid old customization

matsl:
Hi Bob, When preparing this idea I saw your change with using the old variable and calculate the seconds to sleep. Might be a better approach but providing this anyway as food for thought.

### DIFF
--- a/hui-em-but.el
+++ b/hui-em-but.el
@@ -49,6 +49,11 @@
   :type '(integer :match (lambda (_widget value) (and (integerp value) (> value 0))))
   :group 'hyperbole-buttons)
 
+(defcustom hproperty:but-flash-time-seconds 0.05
+  "*Emacs button flash delay."
+  :type 'float
+  :group 'hyperbole-buttons)
+
 (defface hbut-flash
   '((((class color) (min-colors 88) (background light))
      :background "red3")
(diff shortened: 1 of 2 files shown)


rswgnu:
Thanks.  We will leave that pending and revisit across time.

-- Bob

rswgnu (approved):

matsl:
Added ChangeLog  to complete this PR.

hui-em-but.el
```diff
@@ -44,10 +44,16 @@
   :group 'hyperbole-buttons)
 
 (defcustom hproperty:but-flash-time 1000
-  "*Emacs button flash delay."
+  "*Emacs button flash delay. Use of this variable is removed. Use
+`hproperty:but-flash-time-seconds' instead for controlling the flash time."
   :type '(integer :match (lambda (_widget value) (and (integerp value) (> value 0))))
```
rswgnu, line 49: Look up deprecate variable in the Elisp manual for how we should do this.

rswgnu (approved):

hui-em-but.el
matsl: Uses now `make-obsolete-variable` instead.

rswgnu (approved):

PR #25, opened 2020-05-04
Only url encode the user input to the search

matsl:
## What

Only url encode the user input to the search

## Why

At least on Firefox using url encoding on the complete url does not work. It needs at least the '?' and '=' to not be url encoded. Just encoding the user entered search term feels safer.


matsl:
@rswgnu PTAL. This is a blocker for using the search functionality in the web search.

rswgnu:
PR applied.

-- Bob

PR #32, opened 2020-07-12
Use magit if available for git social button

matsl:
Hi,

This uses magit if available when navigating using a git social link. If you are a magit user this is likely what you want instead of entering dired.

We could of course combine this with customization if magit users would want to disable this behaviour. For now I think that is overkill.


matsl:
@rswgnu Changed according to review comments.

rswgnu:
Let's finalize on this tomorrow.

matsl:
@rswgnu Sorry but I've have not yet given up on this feature. Made it configurable and default to the standard behaviour hoping that it could be an acceptable compromise.

rswgnu:
I’ll look at integrating it tomorrow.  Thanks.

-- Bob

PR #33, opened 2020-07-25
Add support for dired-sidebar

matsl:
Add support for dired-sidebar.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,3 +1,7 @@
+2020-07-27  Mats Lidell  <matsl@gnu.org>
+
+* hui-dired-sidebar.el: Support for dired-sidebar added.
+
 2020-07-25  Bob Weiner  <rsw@gnu.org>
 
 * hib-kbd.el (kbd-key:execute-special-series): Added and used in kbd-key:act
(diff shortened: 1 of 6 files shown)


rswgnu:
Hi Mats:

Since we already support Treemacs, is there any significant advantage of dired-sidecar over Treemacs?  I read the blurb on it and tried it briefly but want to understand any advantages first.

To finish off this PR, we should have entries for it in Changes, HYNEWS, the files that summarizes mouse and keyboard commands and an update to the Smart Key appendix in the Hyperbole Manual.

matsl:
Different packages and different features. dired-sidebar is a more lightweight package than treemacs. It does not require you to add a folder as a project so it will open up the folder for the file your are visiting with no extra effort on your part. And it does not create any conflict with treemacs so can be used by those who prefer it etc.

rswgnu:
Sounds good.

-- Bob

matsl:
@rswgnu Docs updated.

rswgnu:
Will integrate.  Thanks.  Do you think you could get C-h h to autoload hyperbole.el within the next two days or so?

-- Bob

rswdp:
Nice job on the doc!

PR #34, opened 2020-07-26
Try to use one defface for flash and highlight

matsl:
This is work in progress. The defface used for flash is just copy of the standard highlight and should be changed to something more suitable. But IMHO this looks like something we could use to simplify the face code.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,11 @@
 * hypb-ert.el (hypb-ert-sym, hypb-ert-sel): Change 2nd args to
     strings to fix Emacs 27 byte-compile error.
 
+2021-05-01  Mats Lidell  <matsl@gnu.org>
+
+* hui-em-but.el (hbut-flash): Set defaults
+    (hproperty:color-list): Use defined-colors
+
 2021-04-30  Mats Lidell  <matsl@gnu.org>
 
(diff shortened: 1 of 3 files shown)


rswgnu:
Thanks, will get on it.

-- Bob

matsl:
Hi @rswgnu Combined some work I had on the flash and color code and updated this PR. I think it works for both getting match highlighting in rolo and button flashing. Please take a look.

rswgnu:
That's great.  Didn't know if we would get to this.  I'll take a look and
then we'll play with it.  -- Bob

rswgnu (approved):
That's a great reduction in code if everything works.  Have you cross-checked for any references to the symbols/names that are being deleted?  Approving assuming yes.

matsl:
I will double check that there are no references left to the deleted function. It was some time since it initially was prepared so my memory is weak regarding what checks I did. So I will double check.

matsl:
Double checked that there are no references left by running this:
```
grep hyrolo-initialize-highlight-face *.el
grep hproperty:background *.el
grep hproperty:foreground *.el
grep hproperty:item-highlight-color *.el
grep hproperty:but-color *.el
grep hproperty:set-flash-color *.el
grep hproperty:set-item-highlight *.el
grep hproperty:set-face-after-init *.el
```

With no output! So should be no problem with deleting those functions.

PR #39, opened 2020-10-04
Update hypb website repository from hyperbole src repositiory 

matsl:
## What

Use a separate Makefile for misc maintenance tasks not directly related to building Hyperbole

## Why

Try not to add more targets to the main Makefile since those should be related to users building and changing the software locally. Targets related to building the distribution and other maintenance related tasks can go into this Makefile instead eventually. 

This PR adds a first maintenance target for updating the Hyperbole site webpage workplace.


rswgnu (approved):
Looka fine.  Thanka.

PR #40, opened 2020-10-19
Use xdg-open to dispatch to user preferences for display alist

matsl:
Default is gimp and office extensions but can be customized.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,3 +1,10 @@
+2020-11-02  Mats Lidell  <matsl@gnu.org>
+
+* hpath.el (hpath:external-display-alist-mswindows):
+           (hpath:internal-display-alist): Add customization.
+           (hpath:external-display-alist-x): Add customization. Use
+    xdg-open to use users preferred app in window system.
+
 2020-10-25  Bob Weiner  <rsw@gnu.org>
 
 * hibtypes.el:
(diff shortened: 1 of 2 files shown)


rswgnu (commented):
Could you also make hpath:external-display-alist-mswindows and  hpath:internal-display-alist into defcustoms as well and submit together?

Does this commit handle all the cases that it replaces by using xdg--open as a default value that triggers for most suffixes?

matsl:
> Could you also make hpath:external-display-alist-mswindows and hpath:internal-display-alist into defcustoms as well and submit together?

Yes. I'll expand the PR to those as well.

> Does this commit handle all the cases that it replaces by using xdg--open as a default value that triggers for most suffixes?

No I have not had that ambition since many of the other extensions are handled today by Emacs itself. I have rather started from the programs I use, gimp and libreoffice, and think we can update the default as we go. An alternative is to use a very general regexp that matches all files with extensions. But then files like .txt etc gets handled by xdg-open which might not be what you want. So with the current design it is a balance.

rswgnu (changes requested):
Please add an entry in the Changes file to document this change (and all your changes) so we have a record of them; otherwise, we merge and lose track of them when not looking at the git log.  Change entries are typically more detailed than gitlog entries in my mind too.

Also, leave the old settings, e.g. from alist-x and alist-mswindows, in a comment, so I can go back, look at what was there and make adjustments.  Then, I'll be ready to merge.

rswgnu (approved):

PR #45, opened 2020-12-03
WIP - Update kexport to html

matsl:
## What

Update kexport to html

## Why

The kexport to html is not compliant with html5 and there are some bugs regarding export of links and key sequences. 

This is work in progress. The export works but the result looks bad since the cells are rendered as the table cells they are and the hyperbole.css used puts a frame around them etc. So more to do there if we want it to look pretty.

This does not yet address the request to control how the header looks and if it shall be exported at all.

### DIFF
--- a/kotl/kexport.el
+++ b/kotl/kexport.el
@@ -100,10 +100,11 @@ pattern may be:
 ;;; ************************************************************************
 
 ;;;###autoload
-(defun kexport:html (export-from output-to &optional soft-newlines-flag)
+(defun kexport:html (export-from output-to &optional soft-newlines-flag no-header)
   "Export a koutline buffer or file in EXPORT-FROM to html format in OUTPUT-TO.
 By default, this retains newlines within cells as they are.  With optional prefix arg, SOFT-NEWLINES-FLAG,
 hard newlines are not used.  Also converts Urls and Klinks into Html hyperlinks.
+Fourth arg NO-HEADER, if non-nil, export only the html body.
 STILL TODO:
   Make delimited pathnames into file links (but not if within klinks).
   Copy attributes stored in cell 0 and attributes from each cell."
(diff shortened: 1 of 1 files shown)


kotl/kexport.el
```diff
@@ -31,14 +31,8 @@
...
   "This is automatically set to the full pathname of the file presently being exported.")
 
-(defcustom kexport:html-body-attributes
-  "BGCOLOR=\"#FFFFFF\"" ;; white background
-  "*String of HTML attributes attached to the <BODY> tag of an HTML exported koutline file."
-  :type 'string
-  :group 'hyperbole-koutliner)
-
 (defcustom kexport:html-description
-  "Created by Hyperbole's outliner.\nSee \"(hyperbole)Koutliner\" for more information."
```
rswgnu, line 41: Leave the quotes here since a patch I will be applying to the Action Key will allow following Info references within backslash double quotes.

kotl/kexport.el
```diff
@@ -196,12 +189,10 @@ STILL TODO:
...
 	 (setq label (kcell-view:label))
-	 (princ (format "<a id=\"k%s\"></a>" label))
-	 (princ (format "<a id=\"k%s\"></a>\n" (kcell-view:idstamp)))
-	 (princ "<td width=2% valign=top><pre>\n")
+	 (princ (format "<span id=\"k%s\"></span>" label))
+	 (princ (format "<span id=\"k%s\"></span>" (kcell-view:idstamp)))
 	 (princ (format
-		 "<font %s>%s%s</font></pre></td>\n"
-		 kexport:label-html-font-attributes
+		 "<font>%s%s</font></pre></td>\n"
```
rswgnu, line 195: Will have to see what you do with the CSS to know if removing the font part works.  Will the CSS we use for the Hyperbole website be good enough or do we need to find an outline-specific CSS?  -- Bob

kotl/kexport.el
matsl: The problem is that content can't contain regular quotes to be valid html. Would have to be inserted as &quote to be there. But that way it would loose its meaning (if the intent is that someone who has the html in buffer can use the link.). I've included another solution that could make more sense I think if the meta info would be viewed in a browser.

matsl:
This version passes the validator.w3.org checks and allows the header to be dropped if wanted, see optional no-header parameter.

It does not do anything to how this is looks. That still remains to be done.

kotl/kexport.el
matsl: I have removed the <font> tag too since that is not html5. Can't say if we need a specific css for the outline but I would not be surprised.

matsl:
Close!

PR #48, opened 2021-02-28
Debug ert tests

matsl:
## What

Add  first unit tests  for defil and defal plus fix bug with replace-match not respecting case in replacement text.


matsl:
Replaced by #49

PR #51, opened 2021-03-06
Add more defal and defil tests

matsl:
## What

This PR adds some more tests for defil and defal. 

Also added test helper for more unit test like tests verifying that hpath:find is called with the right arguments so that we don't go the fill way to check that the proper file was loaded. We might consider going more in this direction though the integration tests that goes the full way is still not a bad idea I think. Look for `hbut-defil-it` and `hbut-defil` to see the difference.

Not sure but there seems to be some problems with cleaning up the defal and defil after a test case. I'm using  `ibtype:delete` now but not sure if that is the proper way to do that.

### DIFF
--- a/test/hibtypes-tests.el
+++ b/test/hibtypes-tests.el
@@ -118,7 +118,7 @@
 (ert-deftest ibtypes::ctags-vgrind-test ()
   (unwind-protect
       (with-temp-buffer
-        (insert "hy-test-helpers:consume-input-events hy-test-helpers.el 20\n")
+        (insert "hy-test-helpers:consume-input-events hy-test-helpers.el 21\n")
         (goto-char (point-min))
         (forward-char 4)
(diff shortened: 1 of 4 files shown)


test/hibtypes-tests.el
```diff
@@ -118,7 +118,7 @@
 (ert-deftest ibtypes::ctags-vgrind-test ()
   (unwind-protect
       (with-temp-buffer
-        (insert "hy-test-helpers:consume-input-events hy-test-helpers.el 20\n")
+        (insert "hy-test-helpers:consume-input-events hy-test-helpers.el 21\n")
```
rswgnu, line 121: Use ? character syntax, e.g. ?\C-u for a control-u character or ?\ for a space character.

rswgnu (approved):

PR #52, opened 2021-03-12
Add tests for ib with labels

matsl:
## What

Add tests for ib with labels

## Why

More tests are better. This also uses an old test case that was commented out due to some strange reason but seems to work now when the functionality is fixed. Thanks for that @rswgnu 

It should probably gone into separate PRs but it becomes administrative burden to split these small things so I group them into one.

### DIFF
--- a/test/hbut-tests.el
+++ b/test/hbut-tests.el
@@ -138,5 +138,21 @@ the button text"
         (should (looking-at "emacs-version")))
     (ibtype:delete 'ibtypes::defil-key)))
 
+;; Labels
+(ert-deftest hbut-ib-link-to-file-with-label ()
+  (with-temp-buffer
+    (insert "<[emacs]>: \"${hyperb:dir}/DEMO\"")
+    (goto-char 4)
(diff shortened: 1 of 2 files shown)


test/hbut-tests.el
```diff
@@ -144,5 +138,21 @@
...
 
+;; Labels
+(ert-deftest hbut-ib-link-to-file-with-label ()
+  (with-temp-buffer
+    (insert "<[emacs]>: \"${hyperb:dir}/DEMO\"")
+    (goto-char 4)
+    (hy-test-helpers:action-key-should-call-hpath:find (concat hyperb:dir "DEMO"))))
+
+(ert-deftest hbut-ib-url-with-label ()
+  "Should find link but failes with (user-error \"No link found\")"
```
rswgnu, line 149: Misspelled fails

rswgnu (approved):

test/hbut-tests.el
matsl: Thanks. Spelling corrected. No this still fails on latest master with the error "No link found".

PR #53, opened 2021-03-13
More tests for ib with labels

matsl:
Adds more test for implicit buttons with labels.

Uses a new helper: `with-simulated-input` found on melpa. Seems promising. Check it out.


rswgnu (approved):
Looks good.

PR #57, opened 2021-03-21
Exclude the test folder from the tar ball

matsl:
## What

This excludes the test folder from being inserted in the release tar ball as we talked about.

## Why?

Thinking about this now I'm not that sure it is a good idea. Why would we not want anyone who downloads the tar ball to have access to the test files? It is not that they take up a lot of space and they are not in the way for anyone who is not interested in them.


rswgnu:
If there are only a few test files, I am okay with leaving them in.

-- Bob

matsl:
Well not sure if they will be few but they will likely be rather small.

rswgnu (changes requested):
Let's leave the tests in and delete this PR.  We will want users to run them at times.

PR #59, opened 2021-04-02
Initial smart support for todotxt-mode

matsl:
## What

Initial smart support for todotxt-mode

## Why

Playing with supporting todotxt-mode in a simple manner.


rswgnu (changes requested):
Add a HY-NEWS entry for this and test runner action buttons.  Also need to document this in Smart Key appendix of Hyperbole Manual and the summary of Smart Keys, hkey-help.txt, or something like that.

rswgnu (approved):

rswgnu (approved):

PR #60, opened 2021-04-05
Use elc extension on hypb-org in ELC_COMPILE macro

matsl:
## What

Use elc extension on hypb-org in `ELC_COMPILE` macro

## Why

Wrong extension makes `make clean` remove `hypb-org.el`


rswgnu (changes requested):
I will push this change, so abandon this one, as we discussed.

PR #63, opened 2021-04-11
Add func to run all tests and make some test more robust

matsl:
## What

Add func to run all tests and make some test more robust as we discussed today. `hypb-run-all-tests`

Also makes some test more robust so that the tests can be run from any directory.

When run like this the tests can possibly break because of dependencies on the local environment and local user settings. That should not happen of course but easy to miss when developing on one machine. So if you spot anything let me know.


rswgnu (approved):
I would use expand-file-name rather than concat but otherwise is fine, so I'll approve now.

matsl:
Hi Bob,

Thanks for the `expand-file-name` suggestion. I have applied it on a few places but there could be more. Will have a look at that.

PR #64, opened 2021-04-15
Add pathname test cases

matsl:
## What

Add pathname test cases - as discussed.

There is bug with a pathname starting with a dash. It should load the file but it results in invalid function `hact`. You might want to fix that before we merge this, and we update the test case, or keep this and we fix the test case when the bug is fixed,

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,3 +1,13 @@
+2021-04-15  Mats Lidell  <matsl@gnu.org>
+
+* test/hibtypes-tests.el (ibtypes::pathname-anchor-test)
+    (ibtypes::pathname-anchor-line-test, ibtypes::pathname-line-column-test)
+    (ibtypes::pathname-load-path-line-column-test)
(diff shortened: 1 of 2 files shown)


rswgnu (approved):
This is a great start and really well-named/clear test cases.  However, a critical part of what we need is to know that the prioritized implicit button dispatch selects the expected ibtype when the Action Key is pressed on it.  So we need a way to test that in each case, writing the expected ibtype that you have already and comparing it to the one that fires.  We can add a little more instrumentation to the hkey-execute function so you can capture the 'pred' and 'pred-value' variables temporarily stored there.  Have a look and we can get into this on Sunday if need be.   I'll approve so you can merge if you want and then just update tests later when this is figured out.

matsl:
Yes. That makes sense. I will merge and use the test cases as base for creating the corresponding action-button test cases.

PR #67, opened 2021-04-17
Use ibtypes tests for action key test

matsl:
## What

- Used ibtypes test with action-key instead to check that proper ibtype is used.
- Moved hui tests to own file
- Some doc string updates

### DIFF
--- a/Changes
+++ b/Changes
@@ -20,10 +20,6 @@
 
 * test/hibtypes-tests.el (ibtypes::pathname-with-dash-loads-file-test):
     Bug fixed - Removed  expected failed.
-  test/hbut-tests.el (hbut-pathname-with-dash-loads-file-test): Remove
-    expected fail and added check for file loading
-    
-    Remove expected fail and added check for file loading
 * test/hbut-tests.el (hbut-pathname-with-dash-loads-file-test): Remove
     expected fail and added check for file loading
 
--- a/Changes
+++ b/Changes
@@ -20,9 +20,9 @@
 
 * test/hibtypes-tests.el (ibtypes::pathname-with-dash-loads-file-test):
     Bug fixed - Removed  expected failed.
-    (hbut-pathname-with-dash-loads-file-test): Bug fixed - Removed
-    expected failed.
-
+  test/hbut-tests.el (hbut-pathname-with-dash-loads-file-test): Remove
+    expected fail and added check for file loading
+    
 * test/hbut-tests.el (hbut-mail-address-test, hbut-pathname-test)
     (hbut-pathname-lisp-variable-test, hbut-pathname-env-variable-test)
     (hbut-pathname-emacs-lisp-file-test, hbut-pathname-anchor-test)
--- a/Changes
+++ b/Changes
@@ -1,25 +1,8 @@
 2021-04-18  Mats Lidell  <matsl@gnu.org>
 
-* test/hibtypes-tests.el (ibtypes::pathname-path-variable-test): Ert test
-    for path variables.
-
-* test/hpath-tests.el (hpath:path-at-point-in-path-variable-test)
-    (hpath:path-at-point-in-path-variable-shorter-than-three-colons-returns-nil-test):
-    Ert test for path variables.
-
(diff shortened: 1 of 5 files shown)


Changes
```diff
@@ -1,3 +1,25 @@
+2021-04-18  Mats Lidell  <matsl@gnu.org>
+
+* test/hbut-tests.el (hub-mail-address-test, hub-pathname-test)
```
rswgnu, line 3: What does 'hub' stand for or mean?

Nice overall expansion in test coverage!

test/demo-tests.el
```diff
@@ -238,7 +238,8 @@
 
 ;; org
 (ert-deftest demo-org-hide-header-test ()
-  "Should work with action-key but does not. Works with org modes hide function."
+  "Hide org mode header.
+Bug: Should work with action-key but does not.  Works with org modes hide function."
```
rswgnu, line 242: Add apostrophe: 'org mode's hide function'.

test/hbut-tests.el
```diff
@@ -165,35 +165,254 @@ the button text"
...
+
+(ert-deftest hub-pathname-emacs-lisp-file-test ()
+  (unwind-protect
+      (with-temp-buffer
+        (insert "\"hyperbole.el\"")
+        (goto-char 2)
+        (action-key)
+        (should (equal major-mode 'emacs-lisp-mode))
+        (should (string= "hyperbole.el" (buffer-name)))))
+  (kill-buffer "hyperbole.el"))
```
rswgnu, line 223: Here, should check that buffer-file-name is non-nil to ensure it found the file and didn't just create a buffer with that name.

test/hbut-tests.el
```diff
@@ -165,35 +165,254 @@ the button text"
...
+        (goto-char 2)
+        (action-key)
+        (should (string= "DEMO" (buffer-name)))
+        (should (= (line-number-at-pos) 3))
+        (should (= (current-column) 45)))
+    (kill-buffer "DEMO")))
+
+(ert-deftest hub-pathname-load-path-line-column-test ()
+  "Pathname with line and position specification."
+  (unwind-protect
```
rswgnu, line 264: Differentiate doc string from that of prior test.

rswgnu (changes requested):

Changes
matsl: :-) Massive typo! Will change to `hbut`

test/hbut-tests.el
```diff
@@ -273,13 +273,13 @@ Bug: With line spec looks in the wrong folder for the file?"
     (kill-buffer "hypb.el")))
 
 (ert-deftest hbut-pathname-with-dash-loads-file-test ()
-  "Pathname with dash loads file.
-Bug: Fails with 'Invalid function: hact'."
-  :expected-result :failed
+  "Pathname with dash loads file."
```
matsl, line 276: This case fails after Emacs does native byte compilation, I think!? It then complain about "Invalid function: hact'". So changing this to reflect the proper behaviour and leave the byte compilation issue to elsewhere. Running "make clean && make bin && make test-all" seems like a good way to run the tests not to be affected by native compilation.

matsl: Oh, seems still to be a problem with make-all but `"make clean && make bin && make test` seems to be stable anyway.

Must experiment more with turning off the lazy native compilation maybe.

rswgnu: hact is a macro; if it is not loaded before compilation, it is assumed to be a function but then when executed is not defined as a function, hence the error.  So we have to find where it is used but not required and either require it or do so only when compiling.

matsl: As we said in IRL the test case works with not byte compiled hyperbole and with byte compiled hyperbole (after `make bin`). When Emacs asynchronously compiles the byte code to native code the test case fails with the given error message,

rswgnu: Okay, just log to a bug list and we'll close this for now.

rswgnu (approved):

PR #69, opened 2021-04-18
Path variable tests

matsl:
## What

Path variable tests.

### DIFF
--- a/test/hbut-tests.el
+++ b/test/hbut-tests.el
@@ -194,5 +194,33 @@ the button text"
            (should (equal (car err) 'error))
            (should (string-search "ibutton at point already has a label" (cadr err)))))))))
 
+(ert-deftest hbut-pathname-path-variable-test ()
+  "Find file in path variable."
+  (with-temp-buffer
+    (insert "\"/var/lib:/bar:/tmp:/foo\"")
+    (goto-char 16)
+    (hy-test-helpers:action-key-should-call-hpath:find "/tmp")))
+
+(ert-deftest hbut-pathname-path-variable-with-two-colons-is-one-file-test ()
+  "Path variable with two colons is not seen as a path variable."
+  (with-temp-buffer
+    (insert "\"/var/lib:/bar:/tmp\"")
+    (goto-char 16)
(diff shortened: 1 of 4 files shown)


test/hbut-tests.el
```diff
@@ -194,5 +194,33 @@ the button text"
...
+
+(ert-deftest hbut-pathname-path-variable-with-two-colons-is-one-file-test ()
+  "Path variable with two colons is not seen as a path variable."
+  (with-temp-buffer
+    (insert "\"/var/lib:/bar:/tmp\"")
+    (goto-char 16)
+    (hy-test-helpers:action-key-should-call-hpath:find "/var/lib:/bar:/tmp")))
+
+(ert-deftest hbut-pathname-path-variable-with-three-colons-is-a-path-test ()
+  "Path variable with two colons is not seen as a path variable."
```
rswgnu, line 212: Fix doc string here to be three colons.

test/hbut-tests.el
```diff
@@ -194,5 +194,33 @@ the button text"
            (should (equal (car err) 'error))
            (should (string-search "ibutton at point already has a label" (cadr err)))))))))
 
+(ert-deftest hbut-pathname-path-variable-test ()
+  "Find file in path variable."
```
rswgnu, line 198: Given that there are different Hyperbole actions when working with a 'path variable name' and a 'path variable value', I find it helpful to use those two terms everywhere so you know what type you are working with.

rswgnu (changes requested):
Generally, looks good.  Will approve after small doc updates.  Need to test selecting paths with completions when selecting a path variable name at some point.  Thanks.

rswgnu (approved):

test/hbut-tests.el
matsl: Not sure about what is the best naming. I have always referred to PATH and its cousins as search paths or path lists. So for functions that actually take the `PATH` variable there `path variable` sounds good. Are you suggesting that for functions that takes the value that could come from a path variable we should call it `path variable value`? And not `path list` or `search path`? 

I'm happy to use whatever you suggest that is consistent with what we have in the code I just have not thought about it enough.

So in this case I should use the doc string `Find file in path variable value`?

PR #70, opened 2021-04-19
Break recursive dependency between hpath and hact

matsl:
## What

Break recursive dependency between `hpath` and `hact`

## Why

This seems to fix the native compilation issue. 

`hact.el` still has a dependency on `hpath.el`. There is one function from hpath that is used in hact. So don't know if this the final say or just a hint want seems to work and what is broken.

From a dependency point of view it looks strange to have this type of circular dependency. But don't know if there is some Emacs lisp voodoo that can be applied or if the modules should be arranged differently. Happy to learn how to  deal with this.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,3 +1,8 @@
+2021-04-19  Mats Lidell  <matsl@gnu.org>
+
+* hpath.el (hact): Add dependency on hact
+* hact.el (mapc): Remove dependency on hpath
+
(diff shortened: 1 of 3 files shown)


rswgnu (approved):

rswgnu:
I remember finding this issue a while back and working to resolve it but I guess I missed a final step.  Thanks.

PR #72, opened 2021-04-22
Add load modifier tests

matsl:
## What

Load modifier tests.


Changes
```diff
@@ -1,3 +1,13 @@
+2021-04-22  Mats Lidell  <matsl@gnu.org>
+
+* test/hpath-tests.el (hpath:load-modifier-loads-file)
+    (hpath:load-modifier-with-plain-file-loads-file-from-load-path): Load
```
rswgnu, line 4: Only a true Lisper could write a function name like that!

rswgnu (approved):

PR #73, opened 2021-04-25
Add org mode tests

matsl:
## What

More org mode related tests. Not complete yet. Did not get around to add tests for the new things you pushed earlier today. Most of these test were developed before your latest updates so could be a slight risk of them being wrong.

I also had to do some changes to hsys-org.el that you should take a look at.

But this is as good as it is going to bet today so I opened a PR for it for review. Will update with more tests later.

### DIFF
--- a/hsys-org.el
+++ b/hsys-org.el
@@ -102,7 +102,9 @@ or are looking for an Org link in another buffer type."
 	   (org-in-regexp org-link-bracket-re))
       (and (boundp 'org-bracket-link-regexp)
 	   (org-in-regexp org-bracket-link-regexp))
-      (hsys-org-face-at-p 'org-link)))
+      (and (boundp 'org-target-link-regexp)
+           (not (null org-target-link-regexp))
+           (org-in-regexp org-target-link-regexp))))
 
 ;; Assumes caller has already checked that the current buffer is in org-mode.
 (defun hsys-org-target-at-p ()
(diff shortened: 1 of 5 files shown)


hsys-org.el
```diff
@@ -102,7 +102,9 @@ or are looking for an Org link in another buffer type."
 	   (org-in-regexp org-link-bracket-re))
       (and (boundp 'org-bracket-link-regexp)
 	   (org-in-regexp org-bracket-link-regexp))
-      (hsys-org-face-at-p 'org-link)))
+      (and (boundp 'org-target-link-regexp)
+           (not (null org-target-link-regexp))
```
matsl, line 106: This uses `org-target-link-regexp` that matches the local link targets. This for so that we can use `org-in-regexp` to we get (start . end) for this case as well.

hsys-org.el
```diff
@@ -139,8 +141,8 @@ Assumes caller has already checked that the current buffer is in `org-mode'."
 	(goto-char (or (previous-single-property-change (point) 'face) (point-min))))
       (when (looking-at "<<<?")
 	(goto-char (match-end 0)))
-      (and (not (get-text-property (point) 'org-linked-text))
-	   (hsys-org-region-with-text-property-value (point) 'face)))))
+      (and (get-text-property (point) 'org-linked-text)
```
matsl, line 144: In my tests `get-text-property` this returns `t` if on a target link. Maybe org has changed if that property is set?

rswgnu (approved):

matsl:
Added some more tests to this PR. Take a look.

rswgnu (approved):
Looks good.

matsl:
Merged after rebase.

PR #76, opened 2021-05-02
Add hsys-org-enable-smart-keys tests

matsl:
## What

Adds test for `hsys-org-enable-smart-keys`

WIP: Still missing test for Smart Keys Context.


rswgnu (approved):

rswgnu (changes requested):
Needs to be updated to match latest code and then merged.  Would like to run these tests today if possible.

PR #77, opened 2021-05-02
Update face code

matsl:
## What

Use customize defface for highlighting ebuts. Some small tweaks to make it work for light and dark background (But could spend a lot more time on that since there are many moving parts there.)

Not sure a about the hypb-items highlighting since it looks (from a quick look) that the regular `highlight` face is used for that. Default for that is also `nil` so you have to toggle it and reload the file with ebuts to see highlighting. Should we change the default to t or is it to colorish as default?

In short would need more time to go over this for it to feel good going out in the release but I think it will provide the basic functionality.

### DIFF
--- a/hui-em-but.el
+++ b/hui-em-but.el
@@ -76,17 +76,10 @@
   :group 'hyperbole-buttons)
 
 (defface hbut-face
-  '((((class color) (min-colors 88) (background light))
-     :background "lightblue")
-    (((class color) (min-colors 88) (background dark))
-     :background "lightblue")
-    (((class color) (min-colors 16) (background light))
-     :background "lightblue")
-    (((class color) (in-colors 16) (background dark))
-     :background "lighblue")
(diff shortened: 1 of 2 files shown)


rswgnu:
Be easier if you show me on the highlight isssue or tell me what to look at vis-à-vis an example.￼

hui-em-but.el
```diff
@@ -75,6 +75,41 @@
...
+  :group 'hyperbole-buttons)
+
+(defface hbut-item-face
+  '((((class color) (min-colors 88) (background light))
+     :background "yellow")
+    (((class color) (min-colors 88) (background dark))
+     :background "yellow")
+    (((class color) (min-colors 16) (background light))
+     :background "yellow")
+    (((class color) (in-colors 16) (background dark))
```
rswgnu, line 99: Typo at min-colors

rswgnu (approved):

PR #78, opened 2021-05-09
Use write-file-functions

matsl:
## What

Use `write-file-functions`

## Why

`local-write-file-hooks` is obsolete

## WIP

_This is work in progress. I noticed that there are more use of obsolete functions and that they might require some more work to get right so will pause this PR for now to get back to it later._


rswgnu (approved):

matsl:
Will merge this since work has progressed in other PRs.

PR #79, opened 2021-05-09
Remove warning for unknown functions - part 1

matsl:
## What

Remove warning for unknown functions - part 1

## Why

Warnings should be removed. The part 1 is because this is what I managed to do today.


hargs.el
```diff
@@ -39,6 +39,11 @@
 (add-hook 'completion-setup-hook #'hargs:set-string-to-complete)
 (add-hook 'minibuffer-exit-hook  #'hargs:unset-string-to-complete)
 
+;;; ************************************************************************
+;;; Public declarations
+;;; ************************************************************************
+(declare-function monkey-filename "etx:monkey")
```
rswgnu, line 45: Typo: 'ext'.

hmouse-drv.el
```diff
@@ -109,6 +109,17 @@ Default is nil."
   "Used to pass the value of a selected region between a Smart Key depress and release.
 This permits the Smart Keys to behave as paste keys.")
 
+;;; ************************************************************************
+;;; Public declarations
+;;; ************************************************************************
+(declare-function br-in-view-window-p "ext:oobr")
+(declare-function br-next-listing-window  "ext:oobr")
+(declare-function br-to-view-window  "ext:oobr")
```
rswgnu, line 117: These should all be "ext:br" as they are in br.el of the OO-Browser.

hmouse-tag.el
```diff
@@ -128,6 +128,12 @@ Keyword matched is grouping 1.  Referent is grouping 2.")
   :type '(file :must-match t)
   :group 'hyperbole-commands)
 
+;;; ************************************************************************
+;;; Public declarations
+;;; ************************************************************************
+(declare-function br-edit "ext:oobr")
```
rswgnu, line 134: "ext:br"

hmouse-tag.el
```diff
@@ -128,6 +128,12 @@ Keyword matched is grouping 1.  Referent is grouping 2.")
...
+(declare-function br-edit-feature "ext:oobr")
```
rswgnu, line 135: "ext:br-ftr"

hui-mouse.el
```diff
@@ -45,6 +45,8 @@
   (require 'hmouse-tag))
 (require 'hsys-org)
 
+(require 'tar-mode)
```
rswgnu, line 48: (eval-when-compile (require 'tar-mode))

hbdata.el
```diff
@@ -50,6 +50,7 @@
 ;;; ************************************************************************
 
 (require 'hbmap)
+(require 'hgnus)
```
rswgnu, line 53: See if (require 'hmail) will work here instead, as it does not have all the dependencies of hgnus.el.  You are going to want (require 'hrmail) too at some point but that also has a lot of dependencies so I wouldn't add that right now but just declare the functions from there that you need.

rswgnu (changes requested):
Consider this approved after the changes requested inline are made, so you can push.  -- Bob

hbdata.el
rswgnu: Good question.  I would just leave them for now and lets get these added to the open Todo items and then I can look at it later.  Please merge if you feel ready.

matsl: I will merge and move on to part2 in a new PR! :-)

PR #82, opened 2021-05-12
Silence a diverse set of warnings

matsl:
## What

* Remove XEmacs functions
* Silence warning for filladapt-mode
* Surpress warnings for free vars

## Why

We need to silence more warnings :-)


hyrolo-menu.el
```diff
@@ -116,11 +116,8 @@
 ;;; This definition is used only by Emacs.
 (defun hyrolo-menubar-menu ()
   "Add a HyRolo menu to the rolo match buffer menubar."
-  (cond ((fboundp 'popup-mode-menu)
```
rswgnu, line 119: InfoDock is not the same as XEmacs.  Although it was built atop XEmacs, we may bring it back atop GNU Emacs at some point, so let's leave the InfoDock references and just remove the XEmacs ones, like zmacs-region-stays.  Any id- identifier is for InfoDock.

rswgnu (changes requested):

hyrolo-menu.el
matsl: OK. I will change this. But would not InfoDock built on top of GNU Emacs use the menu functions in GNU Emacs!?

I know about the id-indentifiers and have not tried to address those yet but they do contribute to the warnings so at some point we would want do fix that as well.

rswgnu (approved):

PR #84, opened 2021-05-14
Inline hproperty:but - only used in one place

matsl:
## What

Inline hproperty:but - only used in one place

## Why

Noticed a warning for this variable and saw it was only used in one place. Not sure if a mistake from earlier face refactoring or just that it is not needed. Take a look what you make of it.


rswgnu (approved):
Yes, I think it was a remnant from the great face redo.

PR #86, opened 2021-05-15
Use cl functions

matsl:
## What

Use the cl functions for copy-list and remprop

## Why

I can't see that copy-list and remprop is defined en emacs 27 nor emacs 28. I had to do this change in order to byte compile.


rswgnu (changes requested):
I have fixed this slightly differently in my next push, so let'd delete this PR.  It was helpful to point out the issues though.

matsl:
No problem. That is the purpose of the PR. An easy way to show what was required.

PR #90, opened 2021-05-16
Diverse changes to silence warnings

matsl:
## What

Diverse changes to silence warnings

## Why

More silencing of compiler warnings. Some including using dealing with changed and obsolete functions and variables.


hmouse-drv.el
```diff
@@ -831,7 +833,9 @@ frame instead."
      ((and (frame-parameter frame 'drag-with-mode-line)
            (window-at-side-p window 'bottom))
       ;; Drag frame when the window is on the bottom of its frame.
-      (mouse-drag-frame start-event 'move)))))
+      (if (fboundp 'mouse-drag-frame)
```
matsl, line 836: `mouse-drag-frame` is not available in Emacs 28.

hyperbole.el
```diff
@@ -680,7 +681,7 @@ This is used only when running from git source and not a package release."
   "Renerate Hyperbole *-autoloads.el files whether they already exist or not."
   (let* ((default-directory hyperb:dir)
 	 (backup-inhibited t)
-	 (find-file-hooks) ;; Prevent header insertion
+	 (find-file-hook) ;; Prevent header insertion
```
matsl, line 684: `find-file-hook` is the preferred name since Emasc 22.1 so should be safe to just use with no guard.

rswgnu (approved):

rswgnu (approved):

rswgnu (approved):
Nice job!

PR #92, opened 2021-05-17
Use compatibility macro for with-suppressed-warnings

matsl:
## What

Use compatibility macro for with-suppressed-warnings

## Why

It was introduced in Emacs 27 and Elpa uses Emacs 26.  Alternative to #91 maybe?

### DIFF
--- a/kotl/kotl-mode.el
+++ b/kotl/kotl-mode.el
@@ -18,12 +18,7 @@
 
 (eval-and-compile (mapc #'require '(cl-lib delsel hsettings hmail kfile kvspec kcell outline org-table kotl-orgtbl)))
 
-(defmacro hypb-with-suppressed-warnings (spec &rest body)
-  "Backwards compatibility macro."
-  (declare (debug (sexp &optional body)) (indent 1))
-  (if (fboundp 'with-suppressed-warnings)
-      `(with-suppressed-warnings ,spec ,@body)
-    `(with-no-warnings ,@body)))
+(require 'hypb)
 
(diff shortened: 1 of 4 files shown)


kotl/kotl-mode.el
```diff
@@ -18,6 +18,13 @@
 
 (eval-and-compile (mapc #'require '(cl-lib delsel hsettings hmail kfile kvspec kcell outline org-table kotl-orgtbl)))
 
+(defmacro hypb-with-suppressed-warnings (spec &rest body)
+  "Backwards compatibility macro."
+  (declare (debug (sexp &optional body)) (indent 1))
+  (if (fboundp 'with-suppressed-warnings)
+      `(with-suppressed-warnings ,spec ,@body)
+    `(with-no-warnings ,@body)))
+
```
rswgnu, line 27: This is the 2nd def of this macro in this commit.  I would change it to hypb:with-suppress-warnings and put it in the hypb.el file and then assure that that is required and you can utilize it wherever.

rswgnu (changes requested):

kotl/kotl-mode.el
matsl: Yes. This is WIP and wasn't sure where to put it since hyperbole.el does not require any other hypb-files. So wanted your wise suggestion. Will refactor this.

matsl:
I have refactored it. Don't know if it make sense for just to use in two places but could maybe be put to more use in the future.

rswgnu (approved):

PR #95, opened 2021-05-24
Add kotl-mode tests

matsl:
## What

Kotl-mode tests.

### DIFF
--- a/test/kotl-mode-tests.el
+++ b/test/kotl-mode-tests.el
@@ -82,6 +82,220 @@
    ;; Cleanup
    (set-buffer-modified-p nil)))
 
+(ert-deftest kotl-mode-extension-open-buffer-in-kotl-mode ()
+  "When a file with kotl extension is created it enters kotl mode."
+  (let ((kotl-file (make-temp-file "hypb" nil ".kotl" "hej")))
+    (unwind-protect
+        (progn
+          (find-file kotl-file)
+          (should (equal major-mode 'kotl-mode)))
+      (delete-file kotl-file))))
+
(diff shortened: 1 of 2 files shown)


test/kotl-mode-tests.el
```diff
@@ -82,6 +82,220 @@
...
+        (progn
+          (find-file kotl-file)
+          (kotl-mode:add-cell)
+
+          ;; Verify default label
+          (should (string= (kcell-view:label (point)) "2"))
+
+          ;; Verify idstamp label
+          (kvspec:activate "ben0")
+          (should (string= (kcell-view:label (point)) "02"))
```
rswgnu, line 142: (kcell-view:idstamp) returns the idstamp of the cell at point, so you can check its value without setting the labels to idstamp mode and this is a better check because the label could be wrong and potentially differ if it isn't updated, so maybe you want to check both.

test/kotl-mode-tests.el
```diff
@@ -82,6 +82,220 @@
...
+          ;; Verify default label
+          (should (string= (kcell-view:label (point)) "2"))
+
+          ;; Verify demote change label
+          (kotl-mode:demote-tree 0)
+          (should (string= (kcell-view:label (point)) "1a")))
+      (delete-file kotl-file))))
+
+(ert-deftest kotl-mode-views ()
+  "Kotl-mode views."
```
rswgnu, line 166: "Kotl-mode test label type activation."

rswgnu (approved):
Nice work.  Just a few suggestions then feel free to commit.

PR #96, opened 2021-05-28
Add ebut create tests

matsl:
## What

Add a few ebut create tests.

These use key-series to defined the buttons so requires interactive mode. Use `make test-all` to execute.


rswgnu:
Try the ebut:program function that can be run in batch mode, I believe.

-- Bob

rswgnu (approved):

PR #100, opened 2021-06-18
Fix even more warnings

matsl:
## What

Fix even more warnings. 

Tried to remove some more warnings but did also find some other problems. So contains also some replacements of functions that are deprecated but should be available from Emacs 27 (Only verified by by inspection.) and fixing a call with too few arguments.

## Why

We need to silence the output when compiling.

### DIFF
--- a/hsys-org.el
+++ b/hsys-org.el
@@ -83,10 +83,7 @@ with different settings of this option.  For example, a nil value makes
   "Follows an optional Org mode LINK to its target.
 If LINK is nil, follows any link at point.  Otherwise, triggers an error."
   (if (stringp link)
-      (cond ((fboundp #'org-link-open-from-string)
-	     (org-link-open-from-string link))
-            ((fboundp #'org-open-link-from-string)
-	     (org-open-link-from-string link))) ;; autoloaded
+      (org-link-open-from-string link)
     (org-open-at-point))) ;; autoloaded
 
 (defact org-internal-link-target (&optional link-target)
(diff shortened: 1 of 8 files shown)


hui-mouse.el
```diff
@@ -762,10 +771,10 @@ If assist-key is pressed:
 
   (interactive)
   (cond ((eobp) (calendar-cursor-to-nearest-date)
-	 (scroll-calendar-right-three-months 1))
+	 (calendar-scroll-right-three-months 1))
 	((< (current-column) 5) (calendar-cursor-to-nearest-date)
-	 (scroll-calendar-left-three-months 1))
-	(t (mark-diary-entries))))
+	 (calendar-scroll-left-three-months 1))
+	(t (diary-view-entries))))
```
rswgnu, line 777: Does this do the same thing as the mark function?  If not, need to update the doc string and manual entries.

rswgnu (changes requested):

hsys-org.el
```diff
@@ -83,10 +83,7 @@ with different settings of this option.  For example, a nil value makes
   "Follows an optional Org mode LINK to its target.
 If LINK is nil, follows any link at point.  Otherwise, triggers an error."
   (if (stringp link)
-      (cond ((fboundp #'org-link-open-from-string)
```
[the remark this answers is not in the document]
matsl: In my dockerized Emacs 27.1 the help message says:

```org-open-link-from-string is an alias for org-link-open-from-string.

(org-open-link-from-string S &optional ARG)

  This function is obsolete since Org 9.3;
  use `org-link-open-from-string' instead.

Open a link in the string S, as if it was in Org mode.
Optional argument is passed to `org-open-file' when S is
a "file" link.

[back]
```
So I feel we can be safe not to check for that.

hui-mouse.el
matsl: Thanks. Nice catch. Shall be `diary-mark-entries`. Fix is coming.

rswgnu: Sounds good for both your comments.

rswgnu (approved):

PR #104, opened 2021-06-20
Test installations using different sources

matsl:
## What

Test installations using different sources

## Description

This is work in progress for an install test. Current setup is to run the script `local-install-test.sh` with args `tarball`, `elpa` or `straight` to choose what install source to use. If the install is successful the unit test are executed in non interactive mode to verify the install further.

Don't know if it makes sense to have these as individual or a combined target in the main Makefile. Since the test is **post** install it might suite better just to run it at times manually or in an automated fashion when a new version of Hyperbole is uploaded in a pipeline.

Please take a look and tell me what you think.

### DIFF
--- a/install-test/local-install-test.sh
+++ b/install-test/local-install-test.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+echo *** Install test ***
+
+install_method="$1"
+
+app=/tmp/hypb-$$
+
+mkdir -p $app
+cp -a * $app
+export HOME=$app/$install_method
+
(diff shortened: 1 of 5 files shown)


install-test/local-install-test.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+echo *** Install test ***
+
+install_method="$1"
```
rswgnu, line 5: I'll probably have some comments after testing this but fine to merge now.

rswgnu (approved):

PR #105, opened 2021-07-03
Add ebut:program tests

matsl:
Add ebut:program tests.

### DIFF
--- a/test/hbut-tests.el
+++ b/test/hbut-tests.el
@@ -26,7 +26,7 @@
     (unwind-protect
         (progn
           (find-file file)
-          (ebut:program "label" 'actypes::link-to-directory "/tmp")
+          (ebut:program "label" 'link-to-directory "/tmp")
           (should (eq (hattr:get (hbut:at-p) 'actype) 'actypes::link-to-directory))
           (should (equal (hattr:get (hbut:at-p) 'args) '("/tmp")))
           (should (equal (hattr:get (hbut:at-p) 'lbl-key) "label")))
@@ -38,7 +38,7 @@
     (unwind-protect
         (progn
           (find-file file)
-          (ebut:program "label" 'actypes::link-to-directory temporary-file-directory)
+          (ebut:program "label" 'link-to-directory temporary-file-directory)
           (should (eq (hattr:get (hbut:at-p) 'actype) 'actypes::link-to-directory))
           (should (equal (hattr:get (hbut:at-p) 'args) (list temporary-file-directory)))
           (should (equal (hattr:get (hbut:at-p) 'lbl-key) "label")))
@@ -50,7 +50,7 @@
     (unwind-protect
         (progn
           (find-file file)
-          (ebut:program "label" 'actypes::exec-shell-cmd "ls /tmp")
+          (ebut:program "label" 'exec-shell-cmd "ls /tmp")
           (should (eq (hattr:get (hbut:at-p) 'actype) 'actypes::exec-shell-cmd))
           (should (equal (hattr:get (hbut:at-p) 'args) '("ls /tmp")))
           (should (equal (hattr:get (hbut:at-p) 'lbl-key) "label")))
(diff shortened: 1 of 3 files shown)


test/hypb-tests.el
```diff
@@ -89,5 +90,39 @@
     (should (equal (hypb:replace-match-string "\\`\\|x" "--xx--" "z")
                    "z--zz--"))))
 
+(ert-deftest hypb:program-create-but-in-buffer ()
+  "Create button with hypb:program in buffer.
+hbut:at-p does not recognise a button created within a buffer."
+  :expected-result :failed
+  (with-temp-buffer
+    (ebut:program "label" 'actypes::link-to-directory "/tmp")
+    (should (eq (hattr:get (hbut:at-p) 'actype) 'actypes::link-to-directory))
```
matsl, line 99: hbut:at-p does not return any properties when the ebut is stored in the buffer. i.e. not in .hypb but after ^L at the end of the file. Is that expected?

test/hypb-tests.el
```diff
@@ -89,5 +90,39 @@
...
+
+(ert-deftest hypb:program-create-link-to-file-line-and-column-but-in-file ()
+  "Create button that links to file with line and column with hypb:program in buffer."
+  (let ((test-file (make-temp-file "test-file")))
+    (unwind-protect
+        (progn
+          (find-file test-file)
+          (ebut:program "label" 'actypes::link-to-file-line-and-column test-file 2 3)
+          (should (eq (hattr:get (hbut:at-p) 'actype) 'actypes::link-to-file-line-and-column))
+          (should (equal (hattr:get (hbut:at-p) 'args) (list 'test-file 2 3)))
```
matsl, line 123: Is this how ebut:program is supposed to work? The arg test-file is not evaluated so the variable test-file is inserted into the ebut. Does not that make ebut:program hard to use?

test/hypb-tests.el
rswgnu: Sounds like a bug that should be fixed.

test/hypb-tests.el
rswgnu: I'll have to look at that.

matsl:
This should now make ert tests work again after ebut:program update. I had to use 'actype:: prefix though. Got undefined variable without it.

gbut was not as straight forward as I initially expected to will come back with that in another PR.

@rswgnu PR has been updated.

rswgnu:
Please resolve the conflicts here.

matsl:
@rswgnu Conflicts resolved.

test/hypb-tests.el
rswgnu: I just tested this with an explicit button created in an *unsent mail* buffer in Emacs 27.1 and it works fine.  Remember GNUS is not yet integrated with Hyperbole as a mail reader/sender, only as a news reader.  You'll have to give me a specific recipe for recreating this error to understand it.

test/hypb-tests.el
rswgnu: Looked at the code and tested and the args argument is evaluated now.  Maybe we had this as a macro but now this is a function so there should not be a problem.

test/hbut-tests.el
```diff
@@ -26,21 +26,21 @@
     (unwind-protect
         (progn
           (find-file file)
-          (ebut:program "label" 'link-to-directory "/tmp")
+          (ebut:program "label" 'actypes::link-to-directory "/tmp")
```
rswgnu, line 29: You can revert this change and other users where you add the "actypes::" prefix; it is not necessary.  I also have another improvement in my next push that will accept prettty much any format for the acton type.  After removing that prefix in ebut:program calls, you can squash and merge.

rswgnu (approved):

test/hypb-tests.el
matsl: Maybe it is an 28 problem. I will try to run it in 27.1 using docker to see if it works there.

The test case as such should work as a recipe. Create a buffer not attached to a file, run ebut-program there, inspect the result using hbut:at-p.

test/hypb-tests.el
matsl: The code snippet above is outdated and is left over from your comment 15 days ago. See the "outdated" mark. It was replaced after the macro change so line 123 about is not in the PR anymore like that. So yes this works as expected and the value of test-file is inserted in the button as expected.

test/hbut-tests.el
matsl: Yes, you are right. It works fine without that. Not sure why it stopped to work for me but now it seems fine again and I have reverted,

PR #106, opened 2021-07-03
Add hui:ebut-create and hui:ebut-modify tests

matsl:
Add hui:ebut-create and hui:ebut-modify tests.

The hui:ebut-modify test triggers an error. The label is inserted again although it is not changed. So it defined as an expected failure.

### DIFF
--- a/test/hui-tests.el
+++ b/test/hui-tests.el
@@ -17,6 +17,7 @@
 
 (require 'ert)
 (require 'with-simulated-input)
+(require 'hui)
 
 (load (expand-file-name "hy-test-helpers"
                         (file-name-directory (or load-file-name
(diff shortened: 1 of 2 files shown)


test/hui-tests.el
```diff
@@ -60,6 +61,38 @@
...
+          (hui:ebut-create)
+          (should (eq (hattr:get (hbut:at-p) 'actype) 'actypes::www-url))
+          (should (equal (hattr:get (hbut:at-p) 'args) '("www.hypb.org")))
+          (should (equal (hattr:get (hbut:at-p) 'lbl-key) "label")))
+        (with-simulated-input "RET RET RET RET"
+          (hui:ebut-modify "label")
+          (should (eq (hattr:get (hbut:at-p) 'actype) 'actypes::www-url))
+          (should (equal (hattr:get (hbut:at-p) 'args) '("www.hypb.org")))
+          (should (equal (hattr:get (hbut:at-p) 'lbl-key) "label"))
+          (should (string= "<(label)>" (buffer-string)))))
```
matsl, line 93: What happens is that the label is inserted again causing `<(label)><(label)>` to be shown in the buffer.

rswgnu: Let's look at that together.  I thought I had generally fixed that issue.

matsl: Yes. By having it as an ert test it is easy to run it using `make test` and  avoid interactions with completion packages. So it can either go in for an eventual fix or we keep the PR until the issue is fixed and the test case can be modified accordingly.

rswgnu (approved):

PR #110, opened 2021-08-04
Refactor hyperbole set functions - patch piece 3

matsl:
## What

The patch refactors use of the set functions in favor of using cl-lib functions that provides the same functionality. Some set functions remains though. So I'm not sure if this is a first step to remove set.el completely or if the remaining function is set.el still make sense to keep. Not sure about the role of set:equal-op in this scenario too since that one still remains.

A few other small unrelated but hopefully useful changes are kept in too to make the PR match Stefans patch better.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2023-01-07  Mats Lidell  <matsl@gnu.org>
+
+* Part three of patch from Stefan monnier. Thank you Stefan.
+  Contains:
+  - Replace use of functions from set.el with standard functions from
+    cl-lib.
+  - Use #' short hand
+
 2022-12-18  Mats Lidell  <matsl@gnu.org>
 
 * Makefile (ELC_COMPILE, ELC_KOTL): Use function to derive elc files.
--- a/hact.el
+++ b/hact.el
@@ -1,9 +1,9 @@
-;;; hact.el --- GNU Hyperbole button action handling  -*- lexical-binding: t; -let*-
+;;; hact.el --- GNU Hyperbole button action handling  -*- lexical-binding: t; -*-
 ;;
 ;; Author:       Bob Weiner
 ;;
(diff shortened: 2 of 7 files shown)


rswgnu:
Mats, please put a comment here explaining the status on this one.  Are we leaving it here for archival purposes?  Does it have any value now with intent to pursue it in the future, etc.?

matsl:
> Mats, please put a comment here explaining the status on this one. Are we leaving it here for archival purposes? Does it have any value now with intent to pursue it in the future, etc.?

I think we kept it as something we potentially would want to do but did not want to pursue at the time. Most important because we thought there were more important things to fix. Maybe time has come now to either drop it or act on it. I think there are multiple issues here (In my priority order):

1. `set` is a pretty generic name so there is a high risk of collision with other packages. If I would see set in some code I would expect it to be part of core Emacs. This could be resolved easily with renaming file and functions to something like `hypb-set` (or whatever prefix we should use for Hyperbole.) (This would make externals happy.)
2.  The set functionality duplicates functionality provided by the cl-lib package so could either be replaced by it (as the current patch is doing) or could be used as a new implementation to simplify things.
3. The way identity is used in the package is a nifty feature but do we need that? To me that complicates the package and a simple equals only implementation would be all we need and by that would be preferable IMHO. (I have not looked again now but I think we only use the equal version. Happy to be proven wrong here.)

matsl:
@rswgnu I'm closing this PR since it does not help us keeping it around. We know about the issues with the set library and we can as well use the closed PR for inspiration if we need that.

rswgnu:
Sounds good.On Feb 6, 2025, at 3:52 PM, Mats Lidell ***@***.***> wrote:﻿
@rswgnu I'm closing this PR since it does not help us keeping it around. We know about the issues with the set library and we can as well use the closed PR for inspiration if we need that.

PR #116, opened 2021-09-09
Pieces of patch set from Stefan

matsl:
## What

Pieces of the patch set from Stefan

These changes are more limited in scope so should be safer. For instance with these changes the tests are still passing.  So lexical code changes that breaks the tests have been left out. 

## Update

Removed the changes to  turn  hact into a defun and the two declare changes. PTAL.

### DIFF
--- a/hactypes.el
+++ b/hactypes.el
@@ -1,10 +1,10 @@
-;;; hactypes.el --- Default action types for GNU Hyperbole
+;;; hactypes.el --- Default action types for GNU Hyperbole  -*- lexical-binding: t; -*-
 ;;
 ;; Author:       Bob Weiner
 ;;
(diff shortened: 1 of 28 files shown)


hactypes.el
```diff
@@ -488,7 +488,7 @@ and its buffer must have a file attached."
        ;; is in progress, so ignore this for now.  -- RSW, 01-25-2020
 
        ;; When not on an ibut and moddifying the link, use existing arguments
```
rswgnu, line 490: Change spelling to 'modifying'.

rswgnu (changes requested):
Largely good and simple but want to a review a few of the macro, etc. changes with you tomorrow.

rswgnu (approved):

rswgnu (approved):

PR #111, opened 2021-08-04
Use interactive-form for action:commandp

matsl:
## What

Piece four of patch from Stefan Monnier. Thank you Stefan.
Contains:
 - hargs.el (hargs:action-get): Use interactive-form.
 - hact.el (action:commandp): Mark as obsolete.
   (action:create, action:params-emacs): Remove.
   (actype:interact): Use interactive-form.

Changes earlier party of this PR has been moved to #118

PR #118, opened 2021-09-12
Part of patch from Stefan

matsl:
## What

Parts from earlier PR broken out to make it simpler. See #111

This does not introduce any new errors in the test suite.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,7 +2,6 @@
 
 * Part of patch from Stefan. Thanks Stefan.
   hbut.el (hbut--inhibit-something): Add variable.
-  Use #' shorthand
   Other refactoring
   hgnus.el (news-setup-hook): Use add-hook.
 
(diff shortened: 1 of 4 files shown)


matsl:
Hi @rswgnu I attempted to bring this PR to life as well. Most of it was to my delight already fixed by the other work we have been doing so not much remains. But before just closing it in silence I thought it was worth a look anyway. So here it is. Passes all test cases. Not sure though we have anything that touches the `hbut--inhibit-something` variable that is suggested by Stefan instead of using the `hyperb:stack-frame` thing (which I haven't looked at all.) Maybe make sense if  `hbut--inhibit-something`  can be given a descriptive name of what is going on there. But maybe I have misunderstood the whole construct and Stefans suggestions is plain wrong?

matsl:
Closing this PR since half of this has been merge in PR #303 The other part will not be included now so closing to not have old PRs hanging around to long.

PR #122, opened 2021-09-17
Add test for split-cell

matsl:
## What

Add test for split-cell

## Why

This looks like a bug. When I run this I get

```
F kotl-mode-split-cell
    Kotl-mode split cell.
    (error "(kotl-mode:move-after): Can’t move tree <1> to within itself")
```
Can be recreated manually. Create a kotl-file with one cell with one word. Split the cell in the middle of the word. Try to demote the new cell that was created. Looking at the cell attributes it looks like the new cell is the same as the original cell where the split started.

### DIFF
--- a/kotl/kotl-mode.el
+++ b/kotl/kotl-mode.el
@@ -2545,7 +2545,7 @@ the current cell."
     ;; delete any preceding whitespace
     (skip-chars-backward " \t\n\r" start)
     (delete-region (max start (point)) (kcell-view:end-contents))
-    (kotl-mode:add-cell arg new-cell-contents (kcell-view:plist))))
+    (kotl-mode:add-cell arg new-cell-contents)))
 
 (defun kotl-mode:transpose-cells (arg)
   "Exchange current and previous visible cells, leaving point after both.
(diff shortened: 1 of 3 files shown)


rswgnu:
Thanks, I’ll work on it when I get time.

-- Bob

rswgnu (approved):

kotl/kotl-mode.el
```diff
@@ -2545,7 +2545,7 @@ the current cell."
     ;; delete any preceding whitespace
     (skip-chars-backward " \t\n\r" start)
     (delete-region (max start (point)) (kcell-view:end-contents))
-    (kotl-mode:add-cell arg new-cell-contents (kcell-view:plist))))
```
matsl, line 2548: Using the current plist here in the new cells seems suspicious. By removing that so that add-cell will create a new plist makes the test case work. Also verified manually that demote etc works for the new cell after this change.

matsl:
@rswgnu Changed how the new cells gets its plist. It seems to fix the test case. PTAL since I'm not sure if anything else is needed.

rswgnu:
New cell should get a copy if the plist aside from the properties added at i itial cell creation, which are cell specific,i.e. copy only user added properties, which typically is none/nil.

-- Bob

PR #127, opened 2021-10-31
Move backwards to previous cell when starting from invalid pos

matsl:
## What

Move backwards to previous cell when starting from invalid pos

## Why

It is more natural to move to the next cell if starting from between two cells, i.e. invalid position. Current behavior is to move to the previous valid cell and then jump to the previous cell. 

This PR counts moving to the previous cell as one step when starting from an invalid position. Test cases for this are added.

## Note

When moving backwards fails the position in the closest cell is kept leaving the point in a valid position. This is considered as a feature for now :smile:


rswgnu (approved):

rswgnu:
Nice.  Thanks.  -- Bob

PR #132, opened 2021-11-28
Broken test cases

matsl:
## What

Regression with latest changes on master. These test cases a broken after the latest changes on master. 

1.  ibut labels are not inserted at the right place. Seems point it not moved out of the ibut. Instead label are inserted where point is.
2. Action key on /tmp call hpath-find with /tmp/. This might be the expected behavior and test case should be updated?

## Note
Not filing as normal PR. Only for reporting this issue for discussion.


rswgnu:
Mats and I worked through this and resolved these issues today without the need for these changes.

PR #133, opened 2021-12-11
Add test for buttons in ls R listing

matsl:
## What 

Add test for buttons in ls R listing

## Why 

Tests are needed!

## Note

The `ls -R` listing seems to result in a one column only listing. We should have tests also where the listing produces multiple columns of files but this was so far I got with this today.

### DIFF
--- a/test/hpath-tests.el
+++ b/test/hpath-tests.el
@@ -18,6 +18,11 @@
 (require 'ert)
 (require 'hpath)
 
+(load (expand-file-name "hy-test-helpers"
+                        (file-name-directory (or load-file-name
+                                                 default-directory))))
+(declare-function hy-test-helpers:action-key-should-call-hpath:find "hy-test-helpers")
+
 (ert-deftest hpath:find-report-lisp-variable-path-name-when-not-exists ()
   "Test that hpath:find expands and returns filename when it is non-existent."
   (condition-case err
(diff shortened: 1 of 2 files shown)


test/hpath-tests.el
```diff
@@ -103,5 +108,25 @@
...
+(ert-deftest hpath:prepend-ls-directory-test ()
+  "Find file in ls -R listing."
+  (let ((shell-buffer "*hypb-test-shell-buffer*"))
+    (unwind-protect
+        (let ((explicit-shell-file-name "/usr/bin/sh")
+              (default-directory hyperb:dir))
+          (hypb-run-shell-test-command "ls -R" shell-buffer)
+          (dolist (file '("COPYING" "man/version.texi" "man/hkey-help.txt" "man/im/demo.png"))
+            (goto-char (point-min))
+            (search-forward (car (last (split-string file "/"))))
```
rswgnu, line 126: You need nil and t as 2nd and third args or this could potentially trigger an error.  Then you can use the bool return value to determine whether to continue or fail the test.

rswgnu (changes requested):

test/hpath-tests.el
matsl: Nice catch. Thanks. It is updated.

rswgnu (approved):

PR #134, opened 2021-12-13
Add package-lint target

matsl:
## What

This is a setup to run package-lint on Hyperbole.

## Why

Want to see what challenges we have if we would like to make Hyperbole package-lint clean. Provided here to share it and show the setup. Including example of running it.


matsl:
Run this command
`> make package-lint 2>&1 | grep -v \"hyperbole\" | tee package-lint.txt `
Grep it for removing all errors complaining about functions not having the package prefix "hyperbole". Around 2034 lines that is. :smile: 
The rest goes into the file below:
[package-lint.txt](https://github.com/rswgnu/hyperbole/files/7707082/package-lint.txt)

rswgnu:
Ok, I see what you mean.  Maybe Melpa is meant for simpler packages :-).  More incentive to finish the release.

-- Bob

matsl:
For this check the major difference we have is that we don't have a package unique prefix for all our files and functions. `hypb` comes first to my mind in that respect. Would be a major undertake to fix that of course. Maybe also points in the direction of having multiple packages like, `hypb` for the hyperbole core, `hypb-kotl` for kotl-mode and possibly others depending on how you would want to slice and dice it. Anyway - think about it.

rswgnu:
Yes, thinking about how we could break it into multiple packages, but definitely after V8 release.

-- Bob

rswgnu (approved):

matsl:
This is a new run of package-lint with filtered out all the warning|errors about files and functions not having the proper prefix and that should not be used to separate prefix from rest of function|variables|symbols. 

That is above 3000 lines of error messages. What remains are: 
[package-lint-2.txt](https://github.com/rswgnu/hyperbole/files/8901931/package-lint-2.txt)

Some of these looks like easy fixes. I might attempt to fix some of them.

rswgnu:
Github mobile won’t open your link.  Try changing it to a .org file and I’ll see if that works.  Thanks.  I am flying today, so just on my phone.

-- Bob

rswgnu (approved):

matsl:
*.org is not a known type. Trying with pdf!?
[output.pdf](https://github.com/rswgnu/hyperbole/files/8903157/output.pdf)

rswgnu:
Same problem.  I’ll just look at it later on a computer.  Thanks.

-- Bob

PR #137, opened 2021-12-28
Add test cases for kotl-mode:transpose-cells

matsl:
## What

Add test cases for kotl-mode:transpose-cells

## Why

Test case in 8.0.0 release todos. 

## Note

* Had to change copy-list to cl-copy-list.
* The resulting position of point after different transpositions does not feel logical. Would it be better if point always stayed with the cell that is moved or stays withing the cell that takes the place of the cell that is moved? The three test cases shows three different behaviors.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2021-12-28  Mats Lidell  <matsl@gnu.org>
+
+* kotl/kotl-mode.el (kotl-mode:exchange-cells): Use cl-copy-list.
+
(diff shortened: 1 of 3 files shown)


rswgnu:
> * The resulting position of point after different transpositions does not feel logical. Would it be better if point always stayed with the cell that is moved or stays withing the cell that takes the place of the cell that is moved? The three test cases shows three different behaviors.

I know point does not look right but it works like transpose lines, but for cells.  The first cell moved down, keeps moving forward with repeated uses.  -- Bob

rswgnu (approved):

PR #148, opened 2022-01-09
Add keybinding tests

matsl:


test/hyperbole-tests.el
```diff
@@ -0,0 +1,78 @@
...
+                                   ,binding))))
+        (verify-keybinding "\C-c@" #'hycontrol-windows-grid)
+        (verify-keybinding "\C-c\C-m" #'hui-select-thing)
+        (verify-keybinding "\C-c\\" #'hycontrol-enable-windows-mode)
+        (verify-keybinding "\C-c/" #'hui-search-web)
+        (verify-keybinding "\C-c." #'hui-select-goto-matching-delimiter)
+        (mapc (lambda (key) (verify-keybinding key #'hkey-either))
+              '("\M-\C-m"
+                ;; Todo: How to verify these bindings.
+                ;;"ESC RET" "M-RET" "ESC <return>" "M-<return>"
```
matsl, line 43: There seems to be some "translation" step involved here so these keys does not directly map to 'hkey-either. Could not figure out how this works and because of that don't know how to verify that the keys are correctly bound.  Ideas welcome here.

test/hyperbole-tests.el
```diff
@@ -0,0 +1,78 @@
...
+                ))
+        (verify-keybinding "\C-hA" #'hkey-help)
+        (verify-keybinding "\M-o" #'hkey-operate)))))
+
```
rswgnu, line 47: Why is this called kkey?

test/hyperbole-tests.el
```diff
@@ -0,0 +1,78 @@
...
+(ert-deftest hyperbole-kkey-init-controls-tests ()
```
rswgnu, line 48: Should be able to combine these two functions with the should and should-not, just passing either the value of hkey-init or the should/should-not function/macro.

rswgnu (changes requested):

test/hyperbole-tests.el
matsl: Fat fingers and bad sight probably. Fixed now.

rswgnu (approved):

test/hyperbole-tests.el
matsl: I gave it a try but I'm not sure it makes it easier to understand the tests. Both `should` and `cl-macrolet` being macros did not make it easier for me either! :smile: I will provide my attempt as a separate PR so you can have a look and will push this so it gets in.

PR #149, opened 2022-01-09
Combine key binding tests in one defun

matsl:
## What

Combine key binding tests in one defun.

## Why

Illustrates how this looks.

## Note

Feel free to suggest how this could be improved further.


rswgnu (approved):
Looks pretty good.

PR #153, opened 2022-01-16
Fix insertion of special char {

matsl:
## What

Fix insertion of special char {.

## Why

Problem reported by texi2any (GNU texinfo) 6.7.

## Note

Does not update hyperbole.html, hyperbole.info and hyperbole.pdf. Do we need to push those updates on all updates to the texi-file or just before a release?


rswgnu:
I build only on certain commits, when a draft is ready.  I think I'll fix these as the markup should be a little different but this is useful to see.  -- Bob

rswgnu (commented):
Reject; updates are in latest push so you can delete this one.

PR #156, opened 2022-01-28
Stefan part of patch 2022 01 applied

matsl:
## What

Stefan part of patch 2022 01 applied

## Why

Suggested improvements by Stefan. Thank you Stefan.

## Note

- Did not pick all changes but most of them. Not all was introduced verbatim but adjusted by me. So if good they are attributed to Stefan and if bad it is all my fault. 
- Changes were introduced as separate commits which might make the review easier.
- I did not do the name change on `hargs:reading-p` fully. Thinking we should think carefully on what prefix to use before doing an overall of the names.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,9 +1,17 @@
+2022-01-28  Mats Lidell  <matsl@gnu.org>
+
+* hargs.el (hargs:make-iform-vector): Converted to macro.
+    (hargs:iform-vector): Use defconst.
+    (hargs:iform-extensions-vector): Use defconst.
+    Patch from Stefan Monnier. Thank you Stefan.
+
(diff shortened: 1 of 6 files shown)


rswgnu:
Hi Mats:

I read all this yesterday and want to discuss with you.  We may want to apply only some of the commits and will figure it out together.

I assume the reason for adding the hargs macro is so the vector code can be byte-compiled.

matsl:
I separated the different parts of the PR into different commit for this reason.

I can't tell if there is a deeper reason for the macro cleanup. To me it looks like a cleanup that can be applied for that reason.

rswgnu:
Sounds good.

-- Bob

matsl:
This PR has been split and provided in multiple other PRs. Closing.

PR #159, opened 2022-01-29
Use 10 digit hash to document git based install (Was: Remove bad test case)

matsl:
## What

Use 10 digit git hash to document a git install and remove a bad designed test case

## Why

Use a 10 digit hash instead of a full git hash in the install documentation. 

Does also activate the unit test but with a test removed that was not working and got pushed before by mistake.

### DIFF
--- a/test/hypb-tests.el
+++ b/test/hypb-tests.el
@@ -99,8 +99,8 @@
   (let ((hyperb:dir "/a_git_folder"))
     (with-mock
       (mock (file-exists-p "/a_git_folder/.git") => t)
-      (mock (shell-command-to-string "git rev-parse HEAD") => "abcdefg")
-      (should (equal (hypb--installation-type) '("git" "abcdefg")))))
+      (mock (shell-command-to-string "git rev-parse HEAD") => "d43d05a0973e8adcbfdd8c85681dac5de669aaa9")
+      (should (equal (hypb--installation-type) '("git" "d43d05a097")))))
   (let ((hyperb:dir "/a_git_folder"))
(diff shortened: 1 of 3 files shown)


test/hypb-tests.el
```diff
@@ -92,26 +92,19 @@
...
-;;       (mock (file-exists-p "/a_git_folder/.git") => nil)
-;;       (should-not (hypb--installation-type)))))
+(ert-deftest hypb--installation-type-test ()
+  "Verify installation type alternatives."
+  (let ((hyperb:dir "/home/user/.emacs.d/elpa/hyperbole-8.0.0pre0.20220126.1138"))
+    (should (equal (hypb--installation-type) '("elpa-devel" "8.0.0pre0.20220126.1138"))))
+  (let ((hyperb:dir "/a_git_folder"))
+    (with-mock
+      (mock (file-exists-p "/a_git_folder/.git") => t)
+      (mock (shell-command-to-string "git rev-parse HEAD") => "abcdefg")
```
rswgnu, line 102: Does this work if installed from elpa or with straight?  Can you test it against all installation methods aside fromthe tar all?

rswgnu (changes requested):

test/hypb-tests.el
matsl: This is a unit test for the function that returns version info for a not official release such as an install from a git repo or from elpa-devel.

matsl:
Function is updates and test case changed to verify that. PTAL

rswgnu (approved):

PR #166, opened 2022-02-12
Silence warnings for unused lexical args and params

matsl:
## What

Silence warnings for unused lexical args and params.

## Why

Make the warnings easier to review by trying to remove them and by that highlight any hidden dependency etc.


rswgnu (commented):
Ensure there are tests for each function where a local variable is deleted.

matsl:
Will be replaced with multiple PRs each covering different part from this PR.

PR #169, opened 2022-02-17
Add test for kcell:ref-to-id

matsl:
## What

Add test for kcell:ref-to-id

## Why

Tests are good.

### DIFF
--- a/test/kcell-tests.el
+++ b/test/kcell-tests.el
@@ -1,0 +1,92 @@
+;;; kcell-tests.el --- Test for kcell        -*- lexical-binding: t; -*-
+;; usage:        GNU Emacs Lisp Library
+;; keywords:     test
+;;
+;; author:       Mats Lidell
+;; org:          Free Software Foundation, Inc.
+;; e-mail:       matsl@gnu.org
+;;
+;; orig-date:    16-Feb-22 at 23:28:49
+;; last-mod:     18-Feb-22 at 19:03:54 by Mats Lidell
+;;
+;; Copyright (C) 2021  Free Software Foundation, Inc.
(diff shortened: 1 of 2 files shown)


test/kcell-tests.el
```diff
@@ -0,0 +1,89 @@
...
+    ("1a|xyz" nil "ben" 3)
+    ("1a|xyz" t "ben" "03|xyz")
+    ("1.1|xyz" nil "ben." 3)
+    ("1.1|xyz" t "ben." "03|xyz")
+
+    ("1a=03|xyz" nil "ben" 3)
+    ("1a=03|xyz" t "ben" "03|xyz")
+    ("1.1=03|xyz" nil "ben." 3)
+    ("1.1=03|xyz" t "ben." "03|xyz")
+
```
rswgnu, line 60: - Should also add a test where relative id is wrong but idstamp is right, and ensure it resolves correctly.  If a klink within a koutline, then when activated, relative id should be updated to the proper value afterwards.

test/kcell-tests.el
```diff
@@ -0,0 +1,89 @@
...
+    (unwind-protect
+        (progn
+          (find-file kotl-file)
+          (kotl-mode:add-child)
+          (let ((failures (delq t (mapcar (lambda (x) (funcall #'kcell-tests--check-ref-to-id x)) kcell-tests--ref-to-id-tests))))
+            (if failures
+                (ert-fail (cons "These refs were not correctly converted to ids:" failures))
+              t)))
+      (delete-file kotl-file))))
+
```
rswgnu, line 87: Looks nice.

rswgnu (approved):

PR #171, opened 2022-02-19
Add target for interactive test run with summary output

matsl:
## What

Add target for interactive test run with summary output

## Why

Useful for running the interactive test runs in "batch mode". 

When running `make test-all-output` the run is terminated when all tests are done and the `*ert* buffer contents is appended to a file `ERT-OUTPUT`. That can look like this:

```
Selector: t
Passed:  220
Failed:  0
Skipped: 0
Total:   220/220

Started at:   2022-02-19 23:51:50+0100
Finished.
Finished at:  2022-02-19 23:52:08+0100

............................................................................................................................................................................................................................
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Author:       Bob Weiner
 #
 # Orig-Date:    15-Jun-94 at 03:42:38
-# Last-Mod:     13-Feb-22 at 10:35:20 by Bob Weiner
+# Last-Mod:     19-Feb-22 at 10:43:51 by Mats Lidell
 #
 # Copyright (C) 1994-2021  Free Software Foundation, Inc.
 # See the file HY-COPY for license information.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -450,6 +450,9 @@ else
 	$(EMACS) --quick $(PRELOADS) --eval "(load-file \"test/hy-test-dependencies.el\")" --eval "(let ((auto-save-default)) $(LOAD_TEST_ERT_FILES) (ert-run-tests-interactively t))"
 endif
 
+test-all-output:
+	$(EMACS) --quick $(PRELOADS) --eval "(load-file \"test/hy-test-dependencies.el\")" --eval "(let ((auto-save-default) (ert-quiet t)) $(LOAD_TEST_ERT_FILES) (ert-run-tests-interactively t) (with-current-buffer \"*ert*\" (append-to-file (point-min) (point-max) \"ERT-OUTPUT\")) (kill-emacs))"
```
rswgnu, line 454: What kind of different summary does this produce compared to 'test-all'?  Maybe show a snippet of the difference in the description.

rswgnu (approved):

Makefile
matsl: @rswgnu Description in the PR has been updated with how a test run looks. Was this the info you wanted and where you wanted it?

rswgnu: Ah, the difference is that the summary output is appended to a file.  The PR title made it sound like the summary output itself differed some how and that is what I did not understand with the first iteration.  Thanks.

matsl: Yes, Plus that the test run is `batch` like in that you do not have to terminate Emacs after a test run. This is so that it can be used in test automation to run all our tests even the interactive ones.

PR #172, opened 2022-02-27
Add target for generating web site

matsl:
## What

Add target for generating web site

## Why

Easy to generate the sources as a make target.

## Note

First step in making make targets for packaging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Author:       Bob Weiner
 #
 # Orig-Date:    15-Jun-94 at 03:42:38
-# Last-Mod:     19-Feb-22 at 10:43:51 by Mats Lidell
+# Last-Mod:     27-Feb-22 at 22:44:32 by Mats Lidell
 #
 # Copyright (C) 1994-2021  Free Software Foundation, Inc.
 # See the file HY-COPY for license information.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -146,6 +149,9 @@ pkg_hyperbole = $(pkg_dir)/hyperbole
 # Temp file to use to build .elc files.
 ELISP_TO_COMPILE = $(pkg_dir)/elc-${USER}
 
+# Path to dir where the web repository is located i.e. hypb:web-repo-location
+HYPB_WEB_REPO_LOCATION = "../hyweb/hyperbole/"
```
matsl, line 153: Since hyweb is a folder we use that is under source control using git it does not work well with having CVS there as well. Example: The CVS folder is personal. A more neutral name might be better but keeping this for now.

rswgnu: I think savannah requires use of CVS.  Do you have a suggestion of what to do?

rswgnu (approved):

rswgnu (approved):

Makefile
matsl: I suggest we use another location as default for the web site sources than inside hyweb since hyweb is under git. But this is good enough for now. I have changed my CVS/Root under hyweb in case I would push the website changes but will try to avoid to push that to hyweb. This is not critical since it is up to each developer how to organize their files. Just a slight inconvenience given by how we use these names.

PR #175, opened 2022-03-05
Remove kotl source buffers modified due to export

matsl:
## What

Remove kotl source buffers modified due to export

## Why

During export kotl files are modified but we do not what that  to be saved to disk by mistake. 

## Note

@rswgnu Could we do an export that will not change the source file?


rswgnu:
> @rswgnu Could we do an export that will not change the source file?

That is what it should do (not modify the source), so we'll have to make updates to remedy this.

rswgnu (approved):

PR #176, opened 2022-03-05
Use only files under source control for packaging

matsl:
## What

Use only files under source control for packaging.

## Why

Old packaging could include files that were present in users workspace.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Author:       Bob Weiner
 #
 # Orig-Date:    15-Jun-94 at 03:42:38
-# Last-Mod:     27-Feb-22 at 22:44:32 by Mats Lidell
+# Last-Mod:      9-Mar-22 at 21:25:34 by Mats Lidell
 #
 # Copyright (C) 1994-2021  Free Software Foundation, Inc.
 # See the file HY-COPY for license information.
@@ -377,7 +377,9 @@ release: package git-push $(pkg_dir)/hyperbole-$(HYPB_VERSION).tar.gz elpa ftp
 
 # Ensure local hyperbole directory is synchronized with master before building a release.
 git-pull:
-	git pull
+	git checkout master && git pull
+	git diff-index --quiet master
+
 git-push:
 	git push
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -410,20 +412,16 @@ $(pkg_dir)/hyperbole-$(HYPB_VERSION).tar.gz:
...
-	cd .. && COPYFILE_DISABLE=1 $(TAR) -clf $(pkg_dir)/h.tar hyperbole
-	cd $(pkg_dir) && COPYFILE_DISABLE=1 $(TAR) xvf h.tar && cd $(pkg_hyperbole) && $(MAKE) packageclean
-	cd $(pkg_hyperbole) && make autoloads && chmod 755 topwin.py && \
-	cd $(pkg_dir) && $(RM) h.tar; \
-	  mv $(pkg_hyperbole) $(pkg_hyperbole)-$(HYPB_VERSION) && \
-	  COPYFILE_DISABLE=1 $(TAR) -clf $(pkg_dir)/hyperbole-$(HYPB_VERSION).tar hyperbole-$(HYPB_VERSION)
-	$(INSTALL) HY-ABOUT HY-ANNOUNCE HY-NEWS HY-WHY.kotl INSTALL README README.md README.html $(pkg_dir)/; chmod 644 $(pkg_dir)/*.tar
+$(pkg_dir)/hyperbole-$(HYPB_VERSION).tar: version $(HYPERBOLE_FILES)
+	$(RM) -fr $(pkg_hyperbole) $(pkg_hyperbole)-$(HYPB_VERSION).tar
+	git archive --format=tar --prefix=hyperbole-$(HYPB_VERSION)/ HEAD | (cd $(pkg_dir) && tar xf -)
```
matsl, line 422: Uses `git archive` for not bringing any loose files into the package.

Makefile
```diff
@@ -410,20 +412,16 @@ $(pkg_dir)/hyperbole-$(HYPB_VERSION).tar.gz:
...
+	cd $(pkg_hyperbole)-$(HYPB_VERSION) && make autoloads && chmod 755 topwin.py
```
matsl, line 423: As before this generates the autoloads and set the right mode on topwin.py.
Q: Should this step also generate the docs? We now have info and pdf checked in which mean that if we forget to update those the wrong version would be packaged? But maybe we could verify that some other way?

rswgnu: You should be able to make 'doc' a dependency and so it is either run and up-to-date or it will run before packaging.  Add that to this PR.  I will approve assuming you are doing that before the merge.

rswgnu (approved):

Makefile
matsl: @rswgnu `doc` is already a dependency on the `package` target so that will already be satisfied. So not really needed to add this dependency here as well although it will not break anything!?

I guess what I'm not 100% happy about is that we have generated files, the`info`-file and `pdf`-file, checked in.  Not a big problem and not something I intend to solve with this PR.

rswgnu: I see your point.  I included them so when someone is browsing the source repo, they are avail without having to figure out how to build them properly.  We should be as turnkey as possible.

Bob

Makefile
rswgnu, line 422: I can't see how the new lines can handle evything the old lines did, but if you are comfortable with the tests you have done, let's give it a try.

Bob

rswgnu (approved):

Makefile
matsl: Fair enough. I see other packages doing it differently but I'm happy either way. So I'm keeping the behavior as is.

Makefile
matsl: I have compared it with what we have uploaded to ftp.gnu.org but apart from that I have done no other test. The comparison is only manual inspection since we then did not have completely the same file structure. What to do you think would be a good test? Try to use the package and install hyperbole from it? Right?

rswgnu:
Sounds good and then run the full test set.

-- Bob

matsl:
The test suite was OK after running an install from the tar ball in a docker environment.  

It did reveal a problem with two test cases that depends on the file `tutorial.el`. As it happens it did not exist in the docker image. Only the compiled file was there. i.e, `tutorial.elc` Maybe I should find another file candidate for those test cases. In all cases it is not a problem with the packaging of hyperbole. So will check :heavy_check_mark: this off as verified and  merge it.

PR #181, opened 2022-03-31
Suggestion for some changes

matsl:
## What

Suggestion for some changes.

## Why

Walked through the `FAST-DEMO` and found some things that might improve it or at least make us more aware of thing that might confuse users.

## Note

The part with kotl-link does not work for me. When I click on `<kotl/EXAMPLE.kotl, 3b10 |c2en>` I get:
```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  kcell:ref-to-id("3b10 |c2en" t)
  actypes::link-to-kotl("kotl/EXAMPLE.kotl, 3b10 |c2en")
  actype:act(link-to-kotl "kotl/EXAMPLE.kotl, 3b10 |c2en")
  klink:act("kotl/EXAMPLE.kotl, 3b10 |c2en" 9670)
  actype:act(klink:act "kotl/EXAMPLE.kotl, 3b10 |c2en" 9670)
  apply(actype:act klink:act ("kotl/EXAMPLE.kotl, 3b10 |c2en" 9670))
  hbut:act(nil)
  actype:act(hbut:act hbut:current)
  hui:hbut-operate(hbut:act "Activate Hyperbole button: " hbut:current)
  hui:hbut-act(hbut:current)
  hkey-execute(nil)
  action-key-internal()
  action-key()
  hkey-either(nil)
  funcall-interactively(hkey-either nil)
  call-interactively(hkey-either nil nil)
  command-execute(hkey-either)
```
If I change the reference to an id-stamp `<kotl/EXAMPLE.kotl, 032 |c2en>`

I get: `hypb:error: (kotl-mode:is-p): Command requires a valid Hyperbole koutline` suggesting that it requires to be within a kotl file!?

See further comments below.

### DIFF
--- a/FAST-DEMO
+++ b/FAST-DEMO
@@ -101,9 +101,9 @@
 
        Press {C-h A} when over any Hyperbole button for an explanation of what it does.
 
-    {M-x dired-other-window RET ${hyperb:dir}/*.org RET}
+    {M-x dired-other-window RET ${hyperb:dir}/*.el RET}
 
-       Hyperbole home directory Dired listing of Org files only
+       Hyperbole home directory Dired listing of Emacs Lisp files only
 
     {C-c @ 22 RET}
 
@@ -117,10 +117,11 @@
        Find the Hyperbole Koutliner source files that begin with 'kotl' and display
        them in an auto-sized grid of windows.
 
-    {C-u 0 C-c @ python-mode RET 33 RET}
+    {C-u 0 C-c @ emacs-lisp-mode RET 33 RET}
 
-       Display the 9 mostly recently used Python buffers in a 3x3 grid.  Other buffers
-       fill in the grid if there are not enough Python ones.
+       Display the 9 mostly recently used Emacs Lisp buffers in a 3x3
+       grid.  A BLANK buffer fills in the grid if there are not enough
+       Emacs Lisp ones.
 
     {C-x d ${hyperb:dir} RET}
 
@@ -130,9 +131,9 @@
        a window grid of those files only.  That works in Buffer Menu and IBuffer modes
        as well.
 
-    {C-x C-b C-x C-q M-x keep-lines RET .py$ RET C-x C-q}
+    {C-x C-b C-x C-q M-x keep-lines RET .el$ RET C-x C-q}
 
-       Display a buffer list of Python file buffers only
+       Display a buffer list of Emacs Lisp file buffers only
 
     {C-x C-b C-x C-q M-x keep-lines RET [\\/]$ RET C-x C-q}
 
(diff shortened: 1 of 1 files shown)


FAST-DEMO
```diff
@@ -101,9 +101,9 @@
 
        Press {C-h A} when over any Hyperbole button for an explanation of what it does.
 
-    {M-x dired-other-window RET ${hyperb:dir}/*.org RET}
+    {M-x dired-other-window RET ${hyperb:dir}/*.el RET}
```
matsl, line 104: There is no org-files in the root folder of Hyperbole so this command breaks.

- Suggested change to use the emacs lisp files instead but for fewer files we could use `.md` or `.otl`

FAST-DEMO
```diff
@@ -130,9 +131,9 @@
        a window grid of those files only.  That works in Buffer Menu and IBuffer modes
        as well.
 
-    {C-x C-b C-x C-q M-x keep-lines RET .py$ RET C-x C-q}
+    {C-x C-b C-x C-q M-x keep-lines RET .el$ RET C-x C-q}
```
matsl, line 134: Same here. What do you think of using Emacs Lisp files instead of Python and that we see to that those have been loaded into Emacs? I think following the `DEMO` does see to that.

FAST-DEMO
```diff
@@ -141,6 +142,16 @@
 
 * Posix/Linux Shell Command Implicit Buttons
 
+    The shell command part of the demo requires as shell that is
```
matsl, line 145: Maybe overkill but I use `fish` and then the `shell` part breaks. So either something according to the lines here or just a comment?

rswgnu, line 145: 'a' shell not 'as'

FAST-DEMO
```diff
@@ -141,6 +142,16 @@
...
+    compatible with bash.  If you are using another shell you can for
+    this part switch to bash,
+
+    { M-x set-variable RET shell-file-name RET "/usr/bin/bash" RET }
+
+       Set the name of the inferior shell to /usr/bin/bash. If you
+       need this the don't forget to reset it when you are ready with
+       the demo.
```
rswgnu, line 153: Proofread and fix language here.

rswgnu (changes requested):

FAST-DEMO
matsl: New version committed. PTAL.

FAST-DEMO
rswgnu: Yes.

FAST-DEMO
rswgnu: This change is fine.

FAST-DEMO
rswgnu: Fine, just update the PR.

FAST-DEMO
```diff
@@ -117,9 +117,9 @@
        Find the Hyperbole Koutliner source files that begin with 'kotl' and display
        them in an auto-sized grid of windows.
 
-    {C-u 0 C-c @ python-mode RET 33 RET}
+    {C-u 0 C-c @ emacs-lisp-mode RET 33 RET}
```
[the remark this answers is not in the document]
rswgnu: Fine.

rswgnu (approved):

PR #184, opened 2022-04-09
Add check and remove temp file

matsl:
## What

Add check and remove temp file

## Why

Improves test case slightly so it verifies that the item at level three looks correct and after termination of test case removes the temp file.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2022-04-09  Mats Lidell  <matsl@gnu.org>
+
+* test/hyrolo-tests.el (hyrolo-add-items-at-multiple-levels): Add check
+    and remove test temp file.
+
 2022-04-03  Kathy  <bk@water-imac.lan>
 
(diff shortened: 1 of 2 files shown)


ChangeLog
```diff
@@ -1,3 +1,8 @@
+2022-04-09  Mats Lidell  <matsl@gnu.org>
+
+* test/hyrolo-tests.el (hyrolo-add-items-at-multiple-levels): Add check
+    and remove test temp file.
+
 2022-04-03  Kathy  <bk@water-imac.lan>
```
matsl, line 6: @rswgnu Looks like you have been away from your computer so user name `Kathy` got inserted in the ChangeLog. Since that is likely you behind the keyboard you might want to update the Change Log entry accordingly. :smile:

rswgnu: I thought I had fixed that.  I'll have a look tomorrow.

rswgnu (approved):

PR #185, opened 2022-04-14
Add hui-select delimited-thing tests

matsl:
## What

Add hui-select delimited-thing and hui-select-thing tests.

### DIFF
--- a/test/hui-select-tests.el
+++ b/test/hui-select-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    14-Apr-22 at 23:45:52
-;; Last-Mod:     15-Apr-22 at 00:31:20 by Mats Lidell
+;; Last-Mod:     15-Apr-22 at 22:38:51 by Mats Lidell
 ;;
 ;; Copyright (C) 2021  Free Software Foundation, Inc.
 ;; See the "../HY-COPY" file for license information.
@@ -24,7 +24,7 @@
 (require 'ert)
 (require 'hui-select)
 
-(ert-deftest hui-select__at-delimited-thing-p ()
+(ert-deftest hui-select--at-delimited-thing-p ()
   "At delimited thing p returns type of thing."
   (with-temp-buffer
     (insert "(\"x\") ")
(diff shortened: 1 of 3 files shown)


test/hui-select-tests.el
```diff
@@ -0,0 +1,79 @@
...
+
+(ert-deftest hui-select__delimited-thing-ending-in-newline ()
+  "Delimited thing marks region of thing when next char after things is a newline."
+  (with-temp-buffer
+    (insert "(\"x\")\n")
+
+    ;; hui-select-sexp-start
+    (goto-char 1)
+    (should (hui-select-delimited-thing))
+    (should (string= (buffer-substring-no-properties (region-beginning) (region-end)) "(\"x\")\n"))))
```
matsl, line 76: @rswgnu Added this test case since the region includes the trailing newline. Should it? Compare to the the test case `hui-select__delimited-thing` where I had to add a space after the sexp for the match to work and that will not include the space.

test/hui-select-tests.el
```diff
@@ -0,0 +1,79 @@
...
+    (should (equal (hui-select-at-delimited-thing-p) 'hui-select-string))
+
+    ;; hui-select-sexp-end
+    (goto-char 5)
+    (should (equal (hui-select-at-delimited-thing-p) 'hui-select-sexp-end))))
+
+(ert-deftest hui-select__delimited-thing ()
+  "Delimited thing marks region of thing."
+  (with-temp-buffer
+    (insert "(\"x\") ")
```
matsl, line 55: @rswgnu Notice I need to add a space after the sexp for it to be recognized. Should not  the string `("x")` suffice for ("x") to be found when pointing at the first paren?

rswgnu: This will depend on the mode of the buffer and the character syntax of the paren.  In emacs-lisp-interaction mode for example, '(x)' is selected when on the opening brace without any trailing space in 27.1.

test/hui-select-tests.el
rswgnu: Yes, the newline is included so you can cut or copy a full line and then yank it and the newline comes with it, inserting the line properly.

rswgnu (approved):

PR #187, opened 2022-04-17
Remove xemacs compatibility code fragments

matsl:
## What

Remove `XEmacs` compatibility code fragments

## Why 

We don't support `XEmacs` so the compatibility code that is left gets in the way (clutters the code) and some times even causes warnings which we want to reduce to a minimum.

## Note

I took one step further here and since `XEmacs` in many cases is similar to `InfoDock` I removed that too. I know it is not 100% true and I also know you might want to bring `InfoDock` back to life some day. But we have version control to support that and I guess a future `InfoDock` would be based on a modern `GNU Emacs` so most of the old code would not be useful. If you do not agree please advice how I should deal with the parts where `InfoDock` is mentioned. (`InfoDock` can still be mentioned here and there since I have gone over the code looking for `XEmacs` references.)

**The PR can probably best be reviewed a commit at a time or a file at the time. Should be mostly the same except for the `ChangeLog` entries.**

All tests passes but I do not think our test covers most of the parts that has been touched so review with that in mind.

I have not touched `mouse-sh.el` since that looks really old and scary. Do you have some insights into what of that is really used and worth keeping.

There is also a part in `kotl-mode.el` where the naming of two functions are explained since there was a byte compiler error for those functions in `XEmacs`. I kept that as a reminder that these functions now could maybe be given their proper names instead. But that will be a longer PR so leaving that for later. On the other hand why not just keep the names and forget about the reason for them being a bit off!?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,11 @@
 
 * hact.el (hact): Use defun. Patch from Stefan Monnier. Thanks Stefan.
 
+* kotl/kmenu.el (id-menubar-set): Remove declared unused function.
+    (id-menubar-kotl): Remove unsued InfoDock menu.
+    (kotl-menubar-menu): Remove XEmacs compatibility code.
+    (kotl-mode-hook): Set hook unconditionally.
+
 * hyrolo-menu.el (id-menubar-hyrolo): Remove InfoDock menu.
     (id-menubar-set, mode-popup-menu): Remove unused public declarations.
     (hyrolo-menubar-menu): Remove XEmacs compatibility code.
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,8 @@
 
 2022-04-18  Mats Lidell  <matsl@gnu.org>
 
+* hsettings.el: Remove XEmacs from comment.
+
 * kotl/kotl-mode.el (kotl-mode:fill-paragraph-or-region):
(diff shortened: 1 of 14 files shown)


rswgnu:
I think we should delete only XEmacs clauses but leave InfoDock ones, as things like InfoDock menus will probably be used in the future.  If this is rather complicated, we can table this for a future release.  Thanks, Bob

matsl:
> delete only XEmacs clauses but leave InfoDock ones

Are you referring to code selected by `(featurep 'infodock)` or in general where `InfoDock` is mentioned?  Keeping the menus I can understand. They stand out as different implementation. I would prefer using source control for keeping that but it is more at the level of a nit for me so I will not complain to much if we keep it on the master branch. :smile:

matsl:
Hi @rswgnu, looked over the old PRs and found this. Can you take a look again?  

It is rebased and ChangeLog updated to reflect that changes looks like they are done now. I think it should be OK. Passes all tests but I'm afraid might not touch on many test cases ... 

From a quick look it seems like the Infodock menues are not touched in this PR so it is preserving those for potential future use.

rswgnu:
Sounds like good things to discuss at our next meeting.

-- Bob

rswgnu (approved):

PR #189, opened 2022-04-18
Make release updates

matsl:
## What

Make release updates

## Why

Maybe something like this could be used. I have not included generation of the webpage yet. Not fully tested either since I don't want to start tagging the release and do ftp uploads and mocking these steps will still not fully verify things will work.

What do you think of trying out this together and see where things break? We can then see what makes sense to add to the process and what is best to have as separate steps?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Author:       Bob Weiner
 #
 # Orig-Date:    15-Jun-94 at 03:42:38
-# Last-Mod:     18-Apr-22 at 23:00:52 by Mats Lidell
+# Last-Mod:     18-Apr-22 at 23:46:38 by Mats Lidell
 #
 # Copyright (C) 1994-2021  Free Software Foundation, Inc.
 # See the file HY-COPY for license information.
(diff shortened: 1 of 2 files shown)


rswgnu:
Sure, sounds good.  -- Bob

Makefile
```diff
@@ -392,16 +389,6 @@ git-pull:
 git-push:
 	git push
 
-# Once the release version number is updated in hyperbole.el and the release is pushed to
```
rswgnu, line 395: Please leave these comments somewhere in the Makefile so we know that Elpa builds are automatic.

rswgnu (changes requested):

Makefile
```diff
@@ -377,30 +374,26 @@ website:
...
+# Generate and distribute a Hyperbole release to ftp.gnu.org.
 # One step in this is to generate an autoloads file for the Koutliner, kotl/kotl-autoloads.el.
-release: package git-push $(pkg_dir)/hyperbole-$(HYPB_VERSION).tar.gz elpa ftp
-	@ echo; echo "Hyperbole $(HYPB_VERSION) released to elpa and ftp.gnu.org successfully."
+release: package $(pkg_dir)/hyperbole-$(HYPB_VERSION).tar.gz ftp git-tag-release
+	@ echo; echo "Hyperbole $(HYPB_VERSION) released to ftp.gnu.org successfully."
 
 # Ensure local hyperbole directory is synchronized with master before building a release.
 git-pull:
+	echo "If this step failes check your work directory for not commited changes"
```
rswgnu, line 386: Replace with 'fails' and 'committed'.

rswgnu (changes requested):

Makefile
matsl: Added it at the top of the Makefile. Is that OK ro do you want it closer to the pkg and release targets?

rswgnu: Is ok.

rswgnu (approved):

PR #190, opened 2022-04-23
Add hui:delimited-selectable-thing tests

matsl:
## What

Add hui:delimited-selectable-thing tests

### DIFF
--- a/test/hui-tests.el
+++ b/test/hui-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    30-Jan-21 at 12:00:00
-;; Last-Mod:     20-Feb-22 at 22:22:23 by Bob Weiner
+;; Last-Mod:     23-Apr-22 at 11:10:58 by Mats Lidell
 ;;
 ;; Copyright (C) 2021-2022  Free Software Foundation, Inc.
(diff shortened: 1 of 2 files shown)


test/hui-tests.el
```diff
@@ -239,6 +239,85 @@ Ensure modifying the button but keeping the label does not create a double label
...
+  (let ((file (make-temp-file "hypb" nil ".txt")))
+    (unwind-protect
+        (progn
+          (find-file file)
+          (insert "<[lnk]>: " file "\n")
+          (beginning-of-buffer)
+          (should (equal (hui:delimited-selectable-thing) file)))
+      (delete-file file))))
+
+(ert-deftest hui--delimited-selectable-thing--in-ebut-return-nil ()
```
matsl, line 275: @rswgnu Is it expected that an ebut is not copied?

rswgnu: Right now it should just copy the text of the button, not the button itself.  Maybe we will improve upon that do e the road.

Bo

rswgnu (approved):

test/hui-tests.el
matsl: But as you see below on line `283` it does not do that at the moment. `Nil` is returned. So this is a bug? If that is the case shall we try to fix it (and update the test case after that) before we merge?

PR #197, opened 2022-06-05
Add target for native byte compilation

matsl:
## What

Add target for native byte compilation

## Why

It is practical to be able to force native byte compilation so that we know unit tests will pick up the native compiled code and not depend on async native compilation in the background.


rswgnu (approved):
Great.  Will be very helpful.

PR #199, opened 2022-06-14
Fix some package lint messages

matsl:
## What

Fix some package lint messages.

## Why

Some or the reaming issues looked small so I gave it a try. 

Some are really hard, impossible!?, to fix since we change function in other packages. I supposed that would require some help from the packages we modify to support some hooks or similar for those changes to be possible cleanly!?  

## Note

Each type of change is in a separate commit except for the ChangeLog and copyright year update that tend to be afterthoughts in my case and comes together as the last commit.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,17 @@
+2022-06-14  Mats Lidell  <matsl@gnu.org>
+
+* hload-path.el (hload-path--make-directory-autoloads)
+    (hload-path--internal-make-directory-autoloads): Use defalias for
+    compatibility function.
+    (hyperb:generate-autoloads): Use it.
+
+* hibtypes.el (cl-lib): Require cl-lib to remove package-lint warning.
+
+* hyperbole.el: Remove defalias for compatilility with obsolete functions
+    since Emacs 25.1 and outline-flag-region available since 2003.
+
+* hui-em-but.el (hproperty:good-colors): package-lint code formating.
+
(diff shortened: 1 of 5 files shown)


ChangeLog
```diff
@@ -1,3 +1,17 @@
+2022-06-14  Mats Lidell  <matsl@gnu.org>
+
+* hload-path.el (hload-path--make-directory-autoloads)
+    (hload-path--internal-make-directory-autoloads): Use defalias for
+    compatibility function.
+    (hyperb:generate-autoloads): Use it.
+
+* hibtypes.el (cl-lib): Require cl-lib to remove package-lint warning.
+
+* hyperbole.el: Remove defalias for compatilility with obsolete functions
```
rswgnu, line 10: Compatibility misspelling.

ChangeLog
```diff
@@ -1,3 +1,17 @@
...
+    since Emacs 25.1 and outline-flag-region available since 2003.
+
+* hui-em-but.el (hproperty:good-colors): package-lint code formating.
```
rswgnu, line 13: Formatting

rswgnu (approved):

ChangeLog
matsl: :scream:

ChangeLog
matsl: I must start with having the spell checker active in the commit window. :pray:

PR #200, opened 2022-06-15
Use hload-path--make-directory-autoloads in Makefile

matsl:
## What 

Use hload-path--make-directory-autoloads in Makefile.

## Why

I missed that it was used from the Makefile so will not work with Emacs pre 28. This ought to do the trick but is only verified with Emacs 28! ;-)


matsl:
Since it breaks the Elpa develop build I'm going to merge it now but sharing here for transparency.

rswgnu (commented):
Got it.

PR #201, opened 2022-06-19
Allow match on different apropos formats

matsl:
## What

Allow match on different apropos formats

## Why

Apropos can have white space or none between command and first parentheses.

My apropos looks like this:
```
❯ apropos grep
bzegrep (1)          - search possibly bzip2 compressed files for a regular expression
bzfgrep (1)          - search possibly bzip2 compressed files for a regular expression
bzgrep (1)           - search possibly bzip2 compressed files for a regular expression
[...]
```
In is original form the test loops over the buffer until it is a match or until 5 seconds has passed. In the later case the test case fails. For each loop it tries to get more output from the process. So even if the process is slow and does not produce any output for the match the loop should ensure that it finally gets there. 

I'm not sure about how the `accept-process-output` really works but since you changed the pattern match could it not just be that your version of apropos does not have white space after the command, like this?
```
❯ apropos grep
bzegrep(1)          - search possibly bzip2 compressed files for a regular expression
bzfgrep(1)          - search possibly bzip2 compressed files for a regular expression
bzgrep(1)           - search possibly bzip2 compressed files for a regular expression
[...]
```
And what you saw was the test case just checking the wrong pattern over an over again? If you try this version does it fail on your machine?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2022-07-19  Mats Lidell  <matsl@gnu.org>
+
+* test/demo-tests.el (fast-demo-key-series-shell-apropos): Add optional
+    whitespace after command to accomodate for different versions on
+    apropos.
+
 2022-07-15  Mats Lidell  <matsl@gnu.org>
 
 * test/hypb-tests.el: Remove test cases for hypb:replace-match-string.
--- a/test/demo-tests.el
+++ b/test/demo-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    30-Jan-21 at 12:00:00
-;; Last-Mod:     19-Jun-22 at 20:25:19 by Bob Weiner
+;; Last-Mod:     16-Jul-22 at 22:39:30 by Mats Lidell
 ;;
 ;; Copyright (C) 2021  Free Software Foundation, Inc.
 ;; See the "HY-COPY" file for license information.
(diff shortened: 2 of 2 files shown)


rswgnu:
Mine does not have whitespace.  I’m fine with jusy making the whitespace match optional in the regex.

-- Bob

rswgnu:
Please make the PR against my latest version and just address this single issue.  Thanks.

rswgnu (changes requested):
See prior comment.

rswgnu:
I pushed a simple change that handles an optional space before the section number in parens.  Try it out and then abandon this pull request if good.

rswgnu:
I changed it be ause the test was failing for me and waiting for the 5 second timeout.  The command output would not appear until some kind of other wait was done in Emacs 27.1.  This was not true of other tests.

-- Bob

matsl:
Was it not failing because the regular expression was wrong for your apropos so you got into the timeout? With the right regexp the test case should work for both of us. I tried the timeout case, but just for emacs-29 so far, and it reports the failure fine in the *ert* output!? No strange report at al.

The construct with timeout for the failure together with `accept-process-output` was stolen from `Emacs` and the tramp test suite. It seems to be the pattern used for waiting for some output and fail on a timer if that never occurs.

Anyway, I'll mark this as `question` fro now and we can come back to it in our next talk just to sort out what happens. Your version is fine. Only problem I have with it is that it goes against the comment you gave me when you taught me about `accept-process-output`. :smile:

rswgnu:
I tested this out and Mats version works better, so I'm approving this.

matsl:
Thanks @rswgnu for evaluating the different versions.

rswgnu:
Yes, let me check on it and we’ll resolve together.  Thanks.  I got most of my A-Z menu idea going today but will be awhile before I push.

-- Bob

PR #203, opened 2022-07-09
Example of doc changes to be within 80 char limit

matsl:
## What 

Example of doc changes to be within 80 char limit.

## Why

Looking at how it would be to go down the route to fix the 80 char limit. You suggested that I would do a couple of docs to see what it means. I did slightly more but before doing all I would want to share it for feedback. It is really a lot to do but it triggers some good discussion points, see below.

Some things I noticed while doing this work:
* elisp mode defaults to fill column of 65 chars which is a bit surprising. We should probably have a style guide so that our doc strings look similar. Tried to look for a general style guide for Emacs about that but did not find it but maybe there is. I have for this work kept the default when formatting the comment body. I have tried to do that in as few cases as possible as that creates a bigger diff.
* Some first line comments are really close to 80 chars so simple things like dropping a `the` or similar suffice to fix those.
* Some first line comments mentions Hyperbole, or other facts that should be clear for the context, and can be fixed by removing these.
* Some first line comments tries to capture everything about a function. Both the major functionality but also all alternatives. This might be the ones that would be hardest to change, but to be honest, I find them sometimes hard to read and understand. So forcing us to describe it within 80 chars would have the good side effect of making them more readable since long sentences are harder to read.
* Some functions have slightly misleading names. Example of that are the `-p` (predicate functions) where the comment line has to compensate for that by explain more what the function does. A good function name could make the task of the first comment line smaller and by that possible to capture in the less space.
* Functions with a lot of alternatives, controlled by optional arguments, might benefit from being split into multiple functions making each function both easier to test and to describe?
* Some techniques I have used:
   * Simply write a shorter first comment line and kept the original comment line in the comment body.
   * Cut the comment line in two. Moving the second part of it into the comment body.
   * Cut our part of the doc in the first line and then write a longer part describing the cut out piece in the comment body.
* My flycheck warned about single quotes in some doc strings so I fixed those as well. Not sure that is necessary. It just happened when I was at it. So feel free to ignore those changes.

### DIFF
--- a/hact.el
+++ b/hact.el
@@ -49,11 +49,12 @@ The type categories are either 'actypes or 'ibtypes.")
   (plist-get symtable 'name))
 
 (defsubst symtable:select (type-category)
-  "Inline the return of the symtable for TYPE-CATEGORY, one of 'actypes or 'ibtypes."
+  "Inline the return of the symtable for TYPE-CATEGORY.
+TYPE-CATEGORY is one of `actypes' or `ibtypes'."
   (plist-get symtable:category-plist type-category))
 
 (defun    symtable:operate (operation symbol-or-name symtable)
-  "Call hash-table function OPERATION with Hyperbole SYMBOL-OR-NAME as key upon SYMTABLE.
+  "Call hash-table function OPERATION with SYMBOL-OR-NAME as key upon SYMTABLE.
 Trigger an error if SYMBOL-OR-NAME cannot be mapped to an existing Elisp
 symbol or if SYMTABLE is invalid."
   (let ((name (cond ((stringp symbol-or-name)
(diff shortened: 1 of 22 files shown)


rswgnu:
See comments above so far to integrate.  I have more to review.

matsl:
Hi @rswgnu, I can't see any comments I'm afraid. Could it be that you have started a review and that is it not completed yet? I normally don't use the review feature but write single comments so I'm not sure how the review thing works. I think you need to finish it before the comments becomes available to me! (or something like that. I really should learn how that can be used. Might be a good thing.)

rswgnu:
I think you are right.  I’ll figure out a solution and let’s move to the smaller chunks so I can approve them faster.

-- Bob

matsl:
@rswgnu So your suggestion is that we shall close this PR and that I should come back with smaller PRs?

rswgnu (changes requested):
This has the comments I made with updates I made but just for the first several entries.  I do suggest you make these changes, close this PR, and then submit smaller, chunked ones that I can process each day for you.  Thanks, Bob

rswgnu:
Yes, I just submitted my partial review with the updates to make before you
do the 'chunked' PRs.

Cheers,

Bob

rswgnu:
Thanks.  Will look in detail at this letter but excited you have worked on
this.

Just FYI, Hyperbole SYMBOL-OR-NAME means not just an Elisp symbol or
symbol-name but one created in Hyperbole's own namespaces like actypes::
and ibtypes::, so this is not equivalent to saying just SYMBOL-OR-NAME.
Most of the other initial changes I looked at are fine.

Cheers,

Bob

PR #206, opened 2022-07-12
Rename global var to remove shadowing warnings

matsl:
## What

Rename global var to remove shadowing warnings.

## Why

By renaming the global variable to `kview-label-sep-len` the functionality is kept intact. The shadowing warnings are by this removed. It remains to analyze if there is any code that relies on dynamic binding. 

Note that this PR should not make the situation worse or even change. It just makes it clear where the global variable is used. With lexical scoping the value passed in to functions will not affect functions using the global variable.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2022-07-16  Mats Lidell  <matsl@gnu.org>
+
+* kotl/kvspec.el (kvspec:show-lines-this-cell): Use kview-label-sep-len.
+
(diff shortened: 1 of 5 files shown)


rswgnu:
Let's talk about this one tomorrow.  I think it should be resolved the opposite way with the function parameters renamed and the global staying the same since that is the one that people might set by name in their own Lisp files.

matsl:
For the discussion marked `label-sep-len` as obsolete and replaced local occurrences with `lbl-sep-len`.

rswgnu (approved):

PR #209, opened 2022-07-14
Patch by Stefan

matsl:
## What

Stefans patch.

Removes some code that digs in the guts of functions, replacing it
with cleaner `advice-add`.

It also adds various FIXMEs about advice which should be
removed/replaced (it's OK as a temporary measure but it should come
*together* with a better long-term solution).

### DIFF
--- a/hrmail.el
+++ b/hrmail.el
@@ -5,7 +5,7 @@
 ;; Orig-Date:     9-May-91 at 04:22:02
 ;; Last-Mod:      5-Jun-22 at 17:59:19 by Bob Weiner
 ;;
-;; Copyright (C) 1991-2016  Free Software Foundation, Inc.
+;; Copyright (C) 1991-2022  Free Software Foundation, Inc.
 ;; See the "HY-COPY" file for license information.
 ;;
 ;; This file is part of GNU Hyperbole.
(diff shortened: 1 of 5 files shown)


hypb.el
```diff
@@ -888,20 +793,6 @@ If FILE is not an absolute path, expand it relative to `hyperb:dir'."
...
-    (while (>= (setq i (1- i)) 0)
-      (setq constant (aref constant-vector i))
-      (cond ((eq constant sym-to-replace)
-	     (aset constant-vector i replace-with-sym))
-	    ((and (fboundp 'compiled-function-p)
-		  (compiled-function-p constant))
-	     (hypb:function-symbol-replace
-	      constant sym-to-replace replace-with-sym))))))
-
 (defun hypb:insert-hyperbole-banner ()
```
rswgnu, line 905: Check that all the removed fun tions above are no longer called anywhere.

Add Changelog entries for all major file-level changes.

hrmail.el
```diff
@@ -320,31 +326,27 @@ see the documentation of `rmail-resend'."
...
-			       (hproperty:but-create))))
+;; Redefine version of 'rmail-get-new-mail' from "rmail.el" to highlight
+;; Hyperbole buttons when possible.
+;;
+
+(defun hrmail--show-msg-and-buttons (&rest _)
+  (if (fboundp 'hproperty:but-create)
+      (progn (widen) (hproperty:but-create)
+	     (rmail-show-message))))
+(if (boundp 'rmail-get-new-mail-post-hook) ;FIXME: Doesn't exist.  XEmacs?
```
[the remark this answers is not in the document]
matsl: Not sure I understand. Does it mean that we need to see to that this works or that it must be kept? 

Related: I can't find `rmail-get-new-mail-post-hook` neither in Emacs nor XEmacs in current and historic data I have access to.

hypb.el
matsl: Verified that all deleted functions are by this PR not use somewhere else.

Noticed while looking at Stefans mail again that ChangeLog entries were provided already by Stefan so I reused those. (Wasted time by providing my own ChangeLog entries. Can be viewed in separate commit. :smile: )

hrmail.el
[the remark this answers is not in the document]
rswgnu: It may be something I added amwhen I redid rmail and rmailsum years ago but not sure if it was merged to mainline.

hypb.el
rswgnu: Ok, thanks.

rswgnu (approved):

PR #218, opened 2022-07-20
Doc strings within 80 char limit for hhist.el

matsl:
## What

Doc strings within 80 char limit for hhist.el

## Why

Extracted from longer PR so review will be easier.


hhist.el
```diff
@@ -25,8 +25,9 @@
 ;;; ************************************************************************
 
 (defun hhist:add (elt)
-  "Add ELT to hyper-history list if not the same as current or prior location (frame configuration).
-ELT must have been created via a call to 'hhist:element' prior to
+  "Add ELT to hyper-history list.
+Do not add if not the same as current or prior location (frame configuration).
```
rswgnu, line 29: Do not add if the same as current or prior location (frame configuration).

rswgnu (changes requested):

rswgnu (approved):

PR #220, opened 2022-07-20
Doc strings within 80 char limit for hib-kbd.el

matsl:
## What

Doc strings within 80 char limit for hib-kbd.el

## Why


hib-kbd.el
```diff
@@ -327,7 +337,8 @@ keyboad input queue, as if they had been typed by the user."
 	(t (error "(kbd-key:normalize): requires a string argument, not `%s'" key-series))))
 
 (defun kbd-key:remove-delimiters (str start-delim end-delim)
-  "Return STR sans START-DELIM and END-DELIM (strings) iff it starts and ends with these delimiters."
+  "Return STR sans START-DELIM and END-DELIM (strings).
+Return iff it starts and ends with these delimiters."
```
rswgnu, line 341: Return nil is STR does not start and end with the given delimiters."

hib-kbd.el
```diff
@@ -447,12 +458,13 @@ For an approximate inverse of this, see `key-description'."
       res)))
 
 (defun kbd-key:extended-command-p (key-series)
-  "Return non-nil if the string KEY-SERIES is a normalized extended command invocation, i.e. M-x command."
+  "Return non-nil if the KEY-SERIES is a normalized extended command invocation.
+A command of the form M-x command."
```
rswgnu, line 462: That is, 'M-x command'.

hib-kbd.el
```diff
@@ -447,12 +458,13 @@ For an approximate inverse of this, see `key-description'."
...
   (when (stringp key-series)
     (string-match kbd-key:extended-command-prefix key-series)))
 
 (defun kbd-key:hyperbole-hycontrol-key-p (key-series)
-  "Return t if normalized, non-nil KEY-SERIES is given when in a HyControl mode, else nil.
+  "True if normalized, non-nil KEY-SERIES is given when in a HyControl mode.
```
rswgnu, line 467: "Return t if KEY-SERIES is normalized, non-nil and in HyControl mode, else nil.
Allow for multiple key sequences strung together."

rswgnu (changes requested):

hib-kbd.el
matsl: FYI: checkdocs complains: ```Keycode` M-x embedded in doc string.  Use \\<mapvar> & \\[command] instead```
Not sure how to apply that so keeping the suggested text for now. Will not cause a warning when byte compiling.

rswgnu (approved):

PR #222, opened 2022-07-21
For Mats only: Broken dev branch to debug

rswgnu:
Hi Mats:

I added ibut:create in hbut.el and other changes.  If you run the test suite, you'll see many implicit button tests now fail.  Some of the new logic is overwriting the actype when the hbut:current properties are set for the button at point.  The actype is set to nil and then nothing works.  I have not been able to find where this is happening so why don't you give it a look.
Thanks.


ChangeLog
```diff
@@ -1,4 +1,11 @@
 2022-07-19  Mats Lidell  <matsl@gnu.org>
+* test/demo-tests.el (fast-demo-key-series-shell-pushd-hyperb-dir,
+                      fast-demo-key-series-shell-grep,
+                      fast-demo-key-series-shell-apropos):
+    Use 'sh' instead of 'bash' for wider applicability.
```
matsl, line 20: We use pushd in a test case which is not available in standard bash. If we want to use /bin/sh we need to adjust that test case.

hbut.el
```diff
@@ -1583,6 +1542,78 @@ associated arguments from the button."
...
+	  (hattr:set 'hbut:current 'lbl-end lbl-end))
+	(hattr:set 'hbut:current 'categ is-type)
+	(when lbl-key
+	  (hattr:set 'hbut:current 'lbl-key lbl-key))
+	(hattr:set 'hbut:current 'loc (or loc (save-excursion
+						(hbut:key-src 'full))))
+	(when action
+	  (hattr:set 'hbut:current 'action action))
+	(or args
+	    (hattr:get 'hbut:current 'args)
```
matsl, line 1602: Sharing my initial findings although they are not complete. (This was tricky to debug!) The test on args looks suspicious. It is not there in the code before refactoring. Seems to fix a few test cases anyway but not sure if this is the problem. 
```suggestion
	(or (hattr:get 'hbut:current 'args)
```

rswgnu: This pointed me in the right direction.  I applied a new commit and fixed the bulk of the problem as a result.  Now we just have to debug remaining issues in 8 test cases.  The above update has been applied separately so can be removed after a merge with the current branch.

matsl: Great! Good news.

PR #225, opened 2022-07-21
Doc strings within 80 char limit for kfile.el

matsl:
## What

Doc strings within 80 char limit for kfile.el.

## Why


kotl/kfile.el
```diff
@@ -66,7 +66,9 @@ Return the new kview."
 
 ;;;###autoload
 (defun kfile:is-p ()
-  "Iff current buffer contains an unformatted or formatted koutline, return file format version string, else nil."
+  "Iff current buffer contain a koutline, return file format version string.
+If not a koutline return nil.  This works both for unformatted
+and formatted koutlines."
```
rswgnu, line 71: "Iff current buffer contains a koutline, return file format version string.
If not a koutline, return nil.  This works both for unformatted
and formatted koutlines."

rswgnu (changes requested):

rswgnu (approved):

PR #229, opened 2022-07-21
Doc strings within 80 char limit for klink.el

matsl:
## What

Doc strings within 80 char limit for klink.el.

## Why


kotl/klink.el
```diff
@@ -85,9 +85,12 @@
 ;;; ************************************************************************
 
 (defun klink:absolute (label-and-pos)
-  "With point in a klink's source buffer and LABEL-AND-POS a list of (klink-label, klink-start, klink-end) including delimiters, return an absolute klink string.
-Klink is of the form: \"<absolute-file-name, cell-ref>\".
-See documentation for `kcell:ref-to-id' for valid cell-ref formats."
+  "Return an absolute klink string.
```
rswgnu, line 88:
```suggestion
  "Return an absolute klink string from LABEL-AND-POS list.
```

kotl/klink.el
```diff
@@ -85,9 +85,12 @@
...
+With point in a klink's source buffer and LABEL-AND-POS a list
+of (klink-label, klink-start, klink-end) including delimiters,
+return an absolute klink string.  Klink is of the form:
```
rswgnu, line 91:
```suggestion
return an absolute klink string.  Klink returned is of the form:
```

rswgnu (changes requested):

rswgnu (approved):

PR #232, opened 2022-07-21
Doc strings within 80 char limit for kproperty.el

matsl:
## What

Doc strings within 80 char limit for kproperty.el.

## Why


kotl/kproperty.el
```diff
@@ -37,17 +37,18 @@
...
   (kproperty:map (lambda (start _end) start) property value))
 
 (defalias 'kproperty:get 'get-text-property)
 
 (defun kproperty:map (function property value)
-  "Apply FUNCTION to each character with PROPERTY `eq' to VALUE in the current buffer.
-FUNCTION is called with the start and end points of the text span with the matching PROPERTY
-and with point at the start."
+  "Apply FUNCTION to each character with PROPERTY VALUE in the current buffer.
+FUNCTION is called with the start and end points of the text span
```
rswgnu, line 50:
```suggestion
FUNCTION is called with the start and end points of the text span,
```

kotl/kproperty.el
```diff
@@ -63,7 +64,8 @@ and with point at the start."
 (defalias 'kproperty:next-single-change 'next-single-property-change)
 
 (defun kproperty:position (property value)
-  "Return the non-narrowed buffer position of the first kcell with PROPERTY set to VALUE, else nil.
+  "Return the non-narrowed buffer position of the first kcell with PROPERTY VALUE.
+If no kcell with PROPERTY VALUE return nil.
```
rswgnu, line 68:
```suggestion
If no kcell with PROPERTY VALUE, return nil.
```

kotl/kproperty.el
```diff
@@ -37,17 +37,18 @@
 		 plist))
 
 (defun kproperty:all-positions (property value)
-  "Return a list of all non-narrowed buffer positions of kcells with PROPERTY set to VALUE, else nil.
+  "Return a list of all non-narrowed positions of kcells with PROPERTY VALUE.
+If no kcells with PROPERTY VALUE return nil.
```
rswgnu, line 41:
```suggestion
If no kcells with PROPERTY VALUE, return nil.
```

rswgnu (changes requested):

rswgnu (approved):

PR #233, opened 2022-07-21
Doc strings within 80 char limit for kview.el

matsl:
## What

Doc strings within 80 char limit for kview.el.

## Why

### DIFF
--- a/kotl/kview.el
+++ b/kotl/kview.el
@@ -51,7 +51,7 @@ Labels are padded with spaces on the left.  Default value is 4."
   :group 'hyperbole-koutliner)
 
 (defcustom kview:default-label-separator ". "
-  "*Characters to insert between label and contents of a koutline cell.
+  "*Default characters to insert between label and contents of a kcell.
 Default value is \". \"."
   :type 'string
   :group 'hyperbole-koutliner)
--- a/kotl/kview.el
+++ b/kotl/kview.el
@@ -1201,8 +1201,8 @@ Value may be the character immediately after point."
     (error "(kview:set-buffer): Invalid kview argument")))
 
 (defun kview:set-cells-status (kview start end cell-status-list)
-  "Use CELL-STATUS-LIST to restore each cell's status.
-Set status in the current buffer's KVIEW, between START and END.
+  "Restore each cell's status in current buffer's KVIEW between START and END.
+Set each cell status from CELL-STATUS-LIST.
 Status is: 0 if all the lines of the cell are visible and it has no
 hidden branches; a positive count of the lines displayed in the cell
 if it has no hidden branches; otherwise, a negative count of the lines
--- a/kotl/kview.el
+++ b/kotl/kview.el
@@ -727,7 +727,7 @@ this function is called."
 
 (defun kview:first-invisible-point (&optional pos)
   "Return the first point that is followed by an invisible character.
-Start from point or optional POS.  If none are found return the
+Start from point or optional POS.  If none are found, return the
 end point of the cell contents.
 Value may be the character immediately after point."
   (unless pos
--- a/kotl/kview.el
+++ b/kotl/kview.el
@@ -60,7 +60,7 @@ Default value is \". \"."
   "Koutline view `outline-regexp' value that handles all label formats.")
 
 (defcustom kview:default-label-type 'alpha
-  "*Default label-type to use for new koutlines.  Default value is `alpha.
+  "*Default label-type to use for new koutlines.  Default value is 'alpha.
 It must be one of the following symbols:
   alpha           for `1b3' full alphanumeric labels
   id              for `027' permanent idstamp labels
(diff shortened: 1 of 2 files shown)


rswgnu (changes requested):

kotl/kview.el
```diff
@@ -51,13 +51,13 @@ Labels are padded with spaces on the left.  Default value is 4."
...
 (defcustom kview:default-label-separator ". "
-  "*Default string of characters to insert between label and contents of a koutline cell.
+  "*Characters to insert between label and contents of a koutline cell.
 Default value is \". \"."
   :type 'string
   :group 'hyperbole-koutliner)
 
 (defcustom kview:default-label-type 'alpha
-  "*Default label-type to use for new koutlines.  Default value is 'alpha.
+  "*Default label-type to use for new koutlines.  Default value is `alpha'.
```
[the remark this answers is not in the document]
matsl: I can revert this. The quotes is not a `docstring` length issue so probably something `flycheck` reported and I tried to fix. We can defer that to later since I think there are more warnings of this problem and we should fix them in the same way.

kotl/kview.el
```diff
@@ -718,7 +726,9 @@ this function is called."
     (set-marker opoint nil)))
 
 (defun kview:first-invisible-point (&optional pos)
-  "Return the first point following optional POS that is followed by an invisible character, else the end point of the cell contents.
+  "Return the first point that is followed by an invisible character.
+Start from point or optional POS.  If none are found return the
+end point of the cell contents.
```
rswgnu, line 731:
```suggestion
  "Return the first point that is followed by an invisible character.
Start from point or optional POS.  If none are found, return the
end point of the cell contents.
```

kotl/kview.el
```diff
@@ -633,9 +639,11 @@ level."
...
 			 label-min-width blank-lines levels-to-show lines-to-show)
   "Return a new kview for BUFFER-NAME.
-Optional ID-COUNTER is the maximum permanent id previously given out in this
-outline.  Optional LABEL-TYPE, LEVEL-INDENT, LABEL-SEPARATOR, LABEL-MIN-WIDTH,
-BLANK-LINES, LEVELS-TO-SHOW, and LINES-TO-SHOW may also be given, otherwise default values are used.
+Optional ID-COUNTER is the maximum permanent id previously given
+out in this outline.  Optional LABEL-TYPE, LEVEL-INDENT,
+LABEL-SEPARATOR, LABEL-MIN-WIDTH, BLANK-LINES, LEVELS-TO-SHOW,
+and LINES-TO-SHOW may also be given, otherwise default values are
+used.
```
rswgnu, line 646:
```suggestion
Optional ID-COUNTER is the maximum permanent id previously utilized in this outline.
Optional LABEL-TYPE, LEVEL-INDENT, LABEL-SEPARATOR, LABEL-MIN-WIDTH,
BLANK-LINES, LEVELS-TO-SHOW, and LINES-TO-SHOW may also be given;
otherwise, default values are used.
```

kotl/kview.el
rswgnu, line 731:
```suggestion
  "Return the first point that is followed by an invisible character.
Start from point or optional POS.  If not found, return the
endpoint of the cell contents.
```

kotl/kview.el
```diff
@@ -732,8 +742,8 @@ Value may be the character immediately after point."
     (or pos end)))
 
 (defun kview:first-visible-point (&optional pos)
-  "Return the first point following optional POS that is followed by a visible character, else (point-max).
-Value may be the character immediately after point."
+  "Return the first point that is followed by a visible character.
+Start from point or optional POS.  If none are found return (point-max)."
```
rswgnu, line 746:
```suggestion
  "Return the first point that is followed by a visible character.
Start from point or optional POS.  If not found, return (point-max)."
```

kotl/kview.el
```diff
@@ -764,7 +776,8 @@ the lines displayed, since it has hidden branches."
    kview t start end))
 
 (defun kview:goto-cell-id (idstamp-or-string)
-  "Move point to start of cell with permanent IDSTAMP-OR-STRING and return t, else nil."
+  "Move point to start of cell with permanent IDSTAMP-OR-STRING.
+On success return t, else nil."
```
rswgnu, line 780:
```suggestion
  "Move point to start of cell with permanent IDSTAMP-OR-STRING.
On success, return t, else nil."
```

kotl/kview.el
```diff
@@ -1157,7 +1176,8 @@ FILL-P is non-nil.  Leave point at TO-START."
     (set-marker new-start nil)))
 
 (defun kview:previous-visible-point (&optional pos)
-  "Return the first point preceding optional POS that is followed by a visible character, else (point-min).
+  "Return the first preceding point that is followed by a visible character.
+Start from point or optional POS.  If none are found return (point-min).
```
rswgnu, line 1180:
```suggestion
  "Return the first preceding point that is followed by a visible character.
Start from point or optional POS.  If not found, return (point-min).
```

rswgnu (changes requested):

rswgnu (changes requested):

rswgnu (approved):

PR #236, opened 2022-07-24
Cleanup - Remove unused function and do not use removed function

matsl:
## What

Do not use obsolete functions.

## Why

We can use the new function directly since they are supported from 27.2.

## Note

Also fixes a duplication related to an obsolete function. Maybe I missed this or made a bad rebase!?

### DIFF
--- a/hact.el
+++ b/hact.el
@@ -3,9 +3,9 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    18-Sep-91 at 02:57:09
-;; Last-Mod:     29-Aug-22 at 00:21:24 by Bob Weiner
+;; Last-Mod:     18-Sep-22 at 22:23:21 by Mats Lidell
 ;;
-;; Copyright (C) 1991-2021  Free Software Foundation, Inc.
+;; Copyright (C) 1991-2022  Free Software Foundation, Inc.
 ;; See the "HY-COPY" file for license information.
 ;;
 ;; This file is part of GNU Hyperbole.
@@ -383,7 +383,7 @@ performing ACTION."
       (let ((hist-elt (hhist:element)))
 	(run-hooks 'action-act-hook)
 	(prog1 (or (if (or (symbolp action) (listp action)
-			   (hypb:emacs-byte-code-p action)
+			   (byte-code-function-p action)
 			   (subrp action)
 			   (and (stringp action) (not (integerp action))
 				(setq action (key-binding action))))
--- a/hibtypes.el
+++ b/hibtypes.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    19-Sep-91 at 20:45:31
-;; Last-Mod:     18-Sep-22 at 22:24:41 by Mats Lidell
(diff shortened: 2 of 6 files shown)


hibtypes.el
```diff
@@ -1104,7 +1104,7 @@ This works when activated anywhere within file line references."
 	 (if (hpath:is-p label)
 	     (hact 'link-to-file label)
            ;; Remove prefix generated by actype and ibtype definitions.
-           (setq label (hypb:replace-match-string "[^:]+::" label "" t))
+           (setq label (replace-regexp-in-string "[^:]+::" "" label nil t))
```
rswgnu, line 1107: Similarly, if you have removed all of these calls, remove hypb:...

rswgnu (changes requested):

hibtypes.el
matsl: Almost same in this case: marked as obsolete here: https://github.com/rswgnu/hyperbole/blob/75c9214c359af9fd33960f43477ceea5e170686f/hypb.el#L573 but since argument order is different there is no replacement so if someone is using this they will need to figure out the difference. Marking it as obsolete and keeping the function for a a while remediates that. *BUT* if we view this as a purely internal function, no likely external usage, I guess we could just remove it?

WDYT?

hact.el
```diff
@@ -334,7 +334,7 @@ Autoloads action function if need be to get the parameter list."
 	       ((eq (car action) 'autoload)
 		(error "(action:params): Autoload not supported: %s" action))
 	       (t (car (cdr action)))))
-	((hypb:emacs-byte-code-p action)
+	((byte-code-function-p action)
```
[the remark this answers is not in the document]
rswgnu: Yes, remove it.

hypb.el
```diff
@@ -570,36 +566,6 @@ WINDOW pixelwise."
 	((symbolp object)
 	 (get object 'hyperbole))))
 
-(make-obsolete 'hypb:replace-match-string 'replace-regexp-in-string "8.0.1")
```
matsl, line 573: Removal of hypb:replace-match-string.

hypb.el
```diff
@@ -310,10 +310,6 @@ will this install the Emacs devdocs package when needed."
 		       dname))))
       (concat "@" dname))))
 
-;;;###autoload
```
matsl, line 313: Removal of obsolete definition of hypb:emacs-byte-code-p

hibtypes.el
matsl: OK. Removed it as well.

matsl:
@rswgnu I learned today how easy it is to remove a file from a PR and then it was easy to repurpose this PR to just a small cleanup. PTAL.

rswgnu:
Let's work on the merge of this when we are together as well.

matsl:
@rswgnu With the fix of the CI this PR now builds so can be merged if approved. Was that the reason for merging this together?

It has 9 commits but all boils down to two small changes. Replace of one function and removal of another. So just look at the combined result and not each commit. I'll squash so the commits becomes one.

rswgnu (approved):

PR #240, opened 2022-07-25
Doc strings within 80 char limit for hmail.el

matsl:
## What

Doc strings within 80 char limit for hmail.el.

## Why

Long doc strings produce warnings.


hmail.el
```diff
@@ -79,7 +79,7 @@ MSG-END."
     (if (search-backward hmail:hbdata-sep msg-start t) (1- (point)) msg-end)))
 
 (defun hmail:hbdata-to-p ()
-  "When in a buffer with embedded Hyperbole button data, move point to the start of the button data.
+  "Move point to the start of embedded button data.
```
rswgnu, line 82:
```suggestion
  "Move point to the start of embedded Hyperbole button data.
```

hmail.el
```diff
@@ -125,7 +125,8 @@ Optional SUBJECT and HELP message may also be given."
 	     "Replace subject, compose message, and then mail.")))
 
 (defun hmail:composing-dir (key-src)
-  "If button KEY-SRC is a mail/news composure buffer, return composure directory, else nil."
+  "If button KEY-SRC is a mail/news composure buffer, return composure directory.
+If not return nil."
```
rswgnu, line 129:
```suggestion
  "If button KEY-SRC is a mail/news composure buffer, return composure directory.
Otherwise, return nil."
```

rswgnu (changes requested):

rswgnu (approved):

PR #257, opened 2022-08-03
hui-mouse.el: Shorten docs strings to be within 80 char limit.

matsl:
## What

hui-mouse.el: Shorten docs strings to be within 80 char limit.

## Why
 

matsl:
Thought it was complete but a syntax error made flycheck report no more errors so I though I was done. :laughing:

PR #259, opened 2022-08-04
hui-select.el: Shorten docs strings to be within 80 char limit.

matsl:
## What

hui-select.el: Shorten docs strings to be within 80 char limit.

## Why
 
### DIFF
--- a/hui-select.el
+++ b/hui-select.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    19-Oct-96 at 02:25:27
-;; Last-Mod:     29-Aug-22 at 21:27:41 by Bob Weiner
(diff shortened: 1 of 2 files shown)


hui-select.el
```diff
@@ -817,7 +830,9 @@ The non-nil value returned is the function to call to select that syntactic unit
       (funcall func))))
```
matsl, line 830: _I change most doc strings just based on the old doc text. Try to change the language, brake up long lines etc. Some doc strings looks strange though when I stumble on them. Link missing mentioning of parameters that sounds like they could be of interest. This is the case here._

The doc strings mentions returning av function but it ends in a `funcall` calling the supplied `func`. Is this function properly documented?

rswgnu: I think so.  The function called probably returns a function symbol to call elsewhere.  My best guess.

rswgnu (approved):

PR #262, opened 2022-08-06
hversion.el: Shorten docs strings to be within 80 char limit.

matsl:
## What 

hversion.el: Shorten docs strings to be within 80 char limit.

## Why
 

hversion.el
```diff
@@ -51,9 +52,10 @@ Override this if the system-computed default is incorrect for your specific mous
 
 ;; Called in hyperbole.el.
 (defun hyperb:stack-frame (function-list &optional debug-flag)
-  "Return the nearest Emacs Lisp stack frame which called any function symbol from FUNCTION-LIST or nil if no match.
-If FUNCTION-LIST contains 'load, 'autoload or 'require, detect
-autoloads not visible within the Lisp level stack frames.
+  "Return the nearest elisp stack frame that called a function from FUNCTION-LIST.
```
rswgnu, line 55:
```suggestion
  "Return the nearest Elisp stack frame that called a function from FUNCTION-LIST.
```

rswgnu (changes requested):

rswgnu (changes requested):

rswgnu (approved):

PR #265, opened 2022-08-06
hyperbole.el: Shorten docs strings to be within 80 char limit.

matsl:
## What

hyperbole.el: Shorten docs strings to be within 80 char limit.

## Why
 

hyperbole.el
```diff
@@ -133,13 +133,15 @@
 See `hkey-initialize'.")
 
 (defcustom hyperbole-mode-lighter " Hypb"
-  "Text to display in the minor-mode area of the modeline when the Hyperbole global minor mode is active."
+  "String to display in mode line when the Hyperbole mode is enabled, nil for none."
```
rswgnu, line 136:
```suggestion
  "String to display in mode line when the Hyperbole global minor mode is enabled.
  Use nil for no Hyperbole mode indicator."
```

rswgnu (changes requested):

rswgnu (approved):

PR #269, opened 2022-08-06
hywconfig.el: Shorten docs strings to be within 80 char limit.

matsl:
## What

hywconfig.el: Shorten docs strings to be within 80 char limit.

## Why
 
### DIFF
--- a/hywconfig.el
+++ b/hywconfig.el
@@ -3,9 +3,9 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    15-Mar-89
-;; Last-Mod:     24-Jan-22 at 00:23:35 by Bob Weiner
+;; Last-Mod:      6-Aug-22 at 23:12:26 by Mats Lidell
 ;;
-;; Copyright (C) 1989-2021  Free Software Foundation, Inc.
+;; Copyright (C) 1989-2022  Free Software Foundation, Inc.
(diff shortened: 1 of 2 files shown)


hywconfig.el
```diff
@@ -123,7 +123,7 @@ NAME, confirms whether or not to replace it."
 ;;; Window configuration ring management (like text kill ring).
 ;;;###autoload
 (defun hywconfig-delete-pop ()
-  "Replace the current frame's window configuration with the one most recently saved to the ring.
+  "Replace the current frame's window configuration with the most recently saved.
```
rswgnu, line 126:
```suggestion
  "Replace current window config with the one most recently saved to the ring.
```

hywconfig.el
```diff
@@ -152,7 +152,7 @@ Use {\\[hywconfig-yank-pop]} to restore it at a later time."
 
 ;;;###autoload
 (defun hywconfig-yank-pop (n)
-  "Replace the current frame's window configuration with the prefix arg Nth prior one in save ring.
+  "Replace current frame's window config with prefix arg Nth prior one in ring.
```
rswgnu, line 155:
```suggestion
  "Replace current window config with the prefix arg Nth prior one in save ring.
```

rswgnu (changes requested):

rswgnu (approved):

PR #272, opened 2022-08-20
First stab at register support

matsl:
## What

First stab at register support

## Why

Sharing the first stab at implementing register support of buttons for ideas and comments.


matsl:
@rswgnu Added support for link-to-ibut

matsl:
@rswgnu Added test cases to the PR. The test `hypb-register-test--register-val-insert` fails due to what it seems a problem with `link-to-ibut`. The setup is in the test. I will try to debug it further but maybe you can have a look?

ChangeLog
```diff
@@ -1,3 +1,12 @@
+2022-09-11  Mats Lidell  <matsl@gnu.org>
+
+* test/hypb-register-tests.el (hypb-register-test--create-register-content)
```
rswgnu, line 3: Everywhere you use hypb-register, let's change that to hui-register since this is part of the UI.

hypb-register.el
```diff
@@ -0,0 +1,73 @@
...
+  "Button register struct."
+  label file mpos link)
+
+;;;###autoload
+(defun hypb-register-struct-at-point ()
+  "Make a Hyperbole link to button register struct for button at point."
+  (let* ((ebut-label (ebut:label-p))
+         (ibut-label (ibut:label-p))
+         (label (or ebut-label ibut-label)))
+    (when (null label)
```
rswgnu, line 39:
```suggestion
    (unless label
```

hypb-register.el
```diff
@@ -0,0 +1,73 @@
...
+      (hypb:error "Point must be at a Hyperbole button"))
+    (make-hypb-register-but
+     :label label
+     :file (buffer-file-name)
```
rswgnu, line 43: Note that there are explicit and implicit buttons where the button source buffer does not have an attached file, so you'll need to handle those appropriately, i.e. links to them work only within the current Emacs session, which may be okay for use with registers which typically are not stored across sessions either.

hypb-register.el
```diff
@@ -0,0 +1,73 @@
...
+     :mpos (point-marker)
+     :link (if ebut-label 'link-to-ebut 'link-to-ibut))))
+
+;;; ************************************************************************
+;;; Private functions
+;;; ************************************************************************
+
+(cl-defmethod register-val-jump-to ((val hypb-register-but) _arg)
+  "Move point to location for Hyperbole button stored in VAL."
+  (let ((buf (marker-buffer (hypb-register-but-mpos val)))
```
rswgnu, line 53: What if the source buffer of the button no longer exists?  You need to read it back in if it has a file attached.

test/hypb-register-tests.el
```diff
@@ -0,0 +1,74 @@
...
+
+(ert-deftest hypb-register-test--create-register-content ()
+  "Verify the struct contains its parts."
+  (let ((file (make-temp-file "hypb")))
+    (unwind-protect
+        (find-file file)
+        (insert "<[label]> $HOME")
+        (goto-char 5)
+        (let ((content (hypb-register-struct-at-point)))
+          (should (equal (hypb-register-but-label content) "label"))
```
rswgnu, line 30: Use a label with spaces in it here to be sure it is converted into a lbl-key without spaces as necessary (internally).

test/hypb-register-tests.el
```diff
@@ -0,0 +1,74 @@
...
+  (let ((file (make-temp-file "hypb")))
+    (unwind-protect
+        (find-file file)
+        (insert "<[label]> $HOME")
+        (goto-char 5)
+        (let ((content (hypb-register-struct-at-point))
+              (pos (point)))
+          (set-buffer "*scratch*")
+          (should (equal (buffer-name) "*scratch*"))
+          (register-val-jump-to content nil)
```
rswgnu, line 48: I think you should be testing the interactive register functions.  So 'insert-register' to test inserting, 'copy-to-register' to save a button to a register, 'jump-to-register' to go to it, and so forth.

rswgnu (changes requested):

ChangeLog
matsl: Changed name to hui-register

test/hypb-register-tests.el
matsl: Changed test to use label "label label"

test/hypb-register-tests.el
matsl: I can add that as well. The high level functions use the methods we test here so more proper unit tests IMHO.

hypb-register.el
matsl: Added an error message if buffer does not exists. Same behavior as for other types register jump to.

hypb-register.el
matsl: Does not `link-to-ebut` require a file name? 

How should we support the case for origin buffer with no filename? Shall we jump to button source and execute the action key instead of using `link-to-ebut`?

test/hypb-register-tests.el
matsl: @rswgnu `copy-to-register` copies the region to a register so I don't know how I could use that in combination with this!? 
On the other hand I noticed we have the function `hui-copy-to-register` Maybe that could embrace this functionality and if on an ibut or ebut copy that instead of thing at point?  wdut?

rswgnu: I think we'll need to do some work on this together.  We just need to decide on the format of the buttons to store, adjust the hui-copy-to-register code and then work out the insertion code, relativizing paths as needed.

matsl: So you think `hui-copy-to-register` is the right function to wrap this behavior?

matsl: @rswgnu Added a proof of concept letting the `hui-copy-to-register` prefer to copy button to thing at point if on a button.

matsl:
Added a test case for inserting an `ebut` and commented out the similar test case for `ibut.` The later test case looks like it breaks due to problem with `link-to-ibut` and having a red build seems contra productive at this stage.

rswgnu (approved):

hypb-register.el
rswgnu: Yes, jump to the button source buffer and move point to the button text (label) and activate the button there.  If the source buffer is no longer live, then raise an error.

matsl: Since `insert-register` inserts an `ebut` that the user acts on would we not need a new action type then for that behavior or let `link-to-ebut` and `link-to-ibut` also handle the buffer only case?

PR #274, opened 2022-09-18
Use maintenance release 28.2

matsl:
## What

Use maintenance release 28.2.

## Why

Bugfix release 28.2 is here so we should change the PR build to use that image instead of 28.1. Not sure how to go about that the build still says it requires 28.1 but after approval I hope it will me possible to merge this and change this requirement to 28.2.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


rswgnu (approved):

matsl:
@rswgnu It did not work. It still claims that 28.1 is required. Are you as owner allowed to merge this?

PR #275, opened 2022-09-25
Update README

matsl:
Adding a line to trigger a build to see if the 28.1 requirement is still enforced although the workflow was changed.


matsl:
Hmm... 28.1 requirement is sticky which is odd since it is removed from the workflow. Will try to add it back so the CI works and figure out after that what need to be done to drop 28.1.

PR #279, opened 2022-10-07
Show docstring warnings again

matsl:
## What

Show docstring warnings again.

## Why

When docstring warnings are gone we need to keep them under control and echoing them when byte compiling is our best option. This PR makes the build show docstring warnings for the `bin` and `eln` targets.

There is a flag that turns all warnings to errors but we have many more warnings still remaining so we are not ready for that yet I think. It can not be set to turn specific warnings to errors.

We could consider to implement our own checker that would parse (or grep) compilation output for specific warnings we want to fail the build on. For now that feels like over engineering. wdyt?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Author:       Bob Weiner
 #
 # Orig-Date:    15-Jun-94 at 03:42:38
-# Last-Mod:     14-Sep-22 at 22:16:31 by Mats Lidell
+# Last-Mod:      7-Oct-22 at 22:17:00 by Mats Lidell
 #
 # Copyright (C) 1994-2022  Free Software Foundation, Inc.
 # See the file HY-COPY for license information.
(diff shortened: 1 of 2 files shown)


rswgnu:
> We could consider to implement our own checker that would parse (or grep) compilation output for specific warnings we want to fail the build on. For now that feels like over engineering. wdyt?

I agree.  No need now.  Let's see how it goes with the warnings re-enabled.

Makefile
```diff
@@ -306,8 +306,7 @@ src: autoloads tags
 # which do not yet exist, plus build TAGS file.
 bin: src
 	$(RM) *.elc kotl/*.elc
-	$(EMACS_BATCH) --eval="(setq-default byte-compile-warnings '(not docstrings))" \
-		-f batch-byte-compile $(EL_KOTL) $(EL_COMPILE)
+	$(EMACS_BATCH) -f batch-byte-compile $(EL_KOTL) $(EL_COMPILE)
```
rswgnu, line 309: Would like you to save the prior version that disables docstring warnings in a comment, so we can easily re-enable it in the future if we want.  But otherwise, this is fine, so you should make that change and then just merge.

rswgnu (changes requested):

Makefile
matsl: The `bin-warn` target's default is to ignore `docstrings` warnings. That allows you to byte compile and get all the rest of the warnings if you want. Or do you want to have it available, i.e. commented our, so that the regular `bin` and `eln` targets can ignore docstrings again if we so choose?

rswgnu: Yes, for the regular targets.  -- Bob

PR #284, opened 2022-10-13
Fix lexical-dynamic warnings

matsl:
## What

Fix lexical-dynamic warnings.

## Why

```
kotl/kview.el:807:54: Warning: Lexical argument shadows the dynamic variable fill-prefix
kotl/kview.el:1399:10: Warning: Variable ‘pos’ left uninitialized
hmouse-sh.el:276:29: Warning: Lexical argument shadows the dynamic variable hmouse-middle-flag
hmouse-sh.el:472:30: Warning: Lexical argument shadows the dynamic variable hmouse-middle-flag
hpath.el:1656:27: Warning: Variable ‘modifier’ left uninitialized
```

### DIFF
--- a/hmouse-sh.el
+++ b/hmouse-sh.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     3-Sep-91 at 21:40:58
-;; Last-Mod:     27-Jul-22 at 00:20:49 by Mats Lidell
+;; Last-Mod:     13-Oct-22 at 22:02:30 by Mats Lidell
 ;;
 ;; Copyright (C) 1991-2022  Free Software Foundation, Inc.
 ;; See the "HY-COPY" file for license information.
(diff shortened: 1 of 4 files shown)


hmouse-sh.el
```diff
@@ -273,15 +273,15 @@ This includes depresses and drags.  Shifted Mouse Key 1 is
 	   [mode-line S-drag-mouse-5]
 	   [mode-line S-mouse-5])))))
 	   
-(defun hmouse-get-bindings (hmouse-middle-flag)
+(defun hmouse-get-bindings (middle-flag)
```
matsl, line 276: Parameter has same name as global dynamic variable. All instances calls the method with the arg `fill-prefix`

kotl/kview.el
```diff
@@ -804,10 +804,10 @@ On success, return t, else nil."
     (when (kotl-mode:goto-cell permanent-id)
       (kcell-view:label))))
 
-(defun kview:insert-contents (kcell contents no-fill fill-prefix)
+(defun kview:insert-contents (kcell contents no-fill cell-fill-prefix)
```
matsl, line 807: Global dynamic variable name used as parameter.

hpath.el
```diff
@@ -1653,7 +1653,7 @@ form is what is returned for PATH."
 		(file-readable-p path))
       (setq path (hpath:call
 		  (lambda (path non-exist)
-		    (let (modifier
+		    (let ((modifier nil)
```
matsl, line 1656: @rswgnu `modifier` is strange. Never set and used in call to two later `concat` calls. Code might be possible to clean up more so leaving this for you to look into if `modifier` can be removed and the `concat` calls simplified?

kotl/kview.el
```diff
@@ -1395,8 +1395,7 @@ new outlines is also set to this new value."
 				  (- sep-len-increase) ?\ )))
 		   (while (re-search-forward indent nil t)
 		     (delete-region
-		      (+ (match-beginning 0) 2) (match-end 0)))))))
-	 pos)
```
matsl, line 1399: pos does not look like it is used so removing it.

rswgnu: You are right that it is not really used but it is sent in a call about 2 lines down but then the value is ignored in the called function as well.  Let's push your updates and then I'll finish the changes here after.

rswgnu (approved):

PR #287, opened 2022-10-23
 Testing exclude, splitting build command in three steps, cleaning the workflow from comments

matsl:
## What

Did an experiment with excluding 28.1 from the CI build. 
Split the build command in three steps so it is easier to view warnings and se if the test step failed. 
Cleaned the workflow description from comments.

## Why

Found a way to remove the 28.1 build if we would like to do that. So I wanted to try if it worked. It might be that 28.1 is still in use by many so we might would like to keep it? 

While I was at it I split the build step in three and cleaned up the workflow files.


matsl:
Hmm... well it does still require 28.1 even though it is excluded so it did not work. Will close it.

PR #288, opened 2022-12-10
Update main.yml

matsl:
## What

Introduce section in the build (Clean, Compile, Test)

## Why

Makes it easier to see which build step that failed and to review warnings in the build step.


matsl:
OK. So 28.1 is still required. Hmm...

Removing the exclude for now since that did not work and just keep the sectioning with separate sections for clean, compile and test.

rswgnu (approved):

PR #291, opened 2022-12-17
Use pattern substitution to derive the elc files

matsl:
## What

Use pattern substitution to derive the elc files

## Why

Makes the makefile cleaner and avoids the risk of an el-file not being compiled.

## Note

There was one el-file that was not compiled.  `hui-em-but.el`. That might have been because of historical reasons. I can't see there would be any reason not to compile it now. The package manager will not care about the Makefile anyways to compiling it will be more close to the users build. Or am I missing something here?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Author:       Bob Weiner
 #
 # Orig-Date:    15-Jun-94 at 03:42:38
-# Last-Mod:     11-Oct-22 at 22:22:16 by Mats Lidell
+# Last-Mod:     18-Dec-22 at 00:29:07 by Mats Lidell
 #
 # Copyright (C) 1994-2022  Free Software Foundation, Inc.
 # See the file HY-COPY for license information.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -191,20 +191,9 @@ EL_KOTL = kotl/kexport.el kotl/kfile.el kotl/kfill.el kotl/kimport.el kotl/klabe
...
-	     hui-treemacs.elc hui-window.elc hui.elc hvar.elc hversion.elc hvm.elc hypb.elc hyperbole.elc \
-	     hyrolo-demo.elc hyrolo-logic.elc hyrolo-menu.elc hyrolo.elc hywconfig.elc \
-	     set.elc hypb-ert.elc hui-dired-sidebar.elc hypb-maintenance.elc hui-register.elc
-
-ELC_KOTL = kotl/kexport.elc kotl/kfile.elc kotl/kfill.elc kotl/kimport.elc kotl/klabel.elc \
-	   kotl/klink.elc kotl/kmenu.elc kotl/kotl-mode.elc kotl/kotl-orgtbl.elc \
-           kotl/kcell.elc kotl/kproperty.elc kotl/kview.elc kotl/kvspec.elc
+ELC_COMPILE = $(EL_COMPILE:.el=.elc)
+
+ELC_KOTL = $(EL_KOTL:.el=.elc)
```
rswgnu, line 196: You are missing the kotl/ prefix here.

rswgnu (changes requested):

Makefile
matsl: No, I don't think so. This should be a pattern substitution for all words in the string `ET_KOTL` a matching end is replaced. 
See https://github.com/rswgnu/hyperbole/actions/runs/3722317980/jobs/6313030557#step:4:29 
and https://www.gnu.org/software/make/manual/html_node/Substitution-Refs.html

rswgnu:
Got it.  Thanks, I misread it.-- BobOn Dec 18, 2022, at 2:26 AM, Mats Lidell ***@***.***> wrote:﻿
@matsl requested your review on: #291 Use pattern substitution to derive the elc files.

rswgnu (approved):

rswgnu (approved):

PR #293, opened 2023-01-01
Use kotl-mode:copy-region-as-kill in kotl-mode

matsl:
## What

Use `kotl-mode:copy-region-as-kill` in `kotl-mode.`

## Why

Looked at why`M-w` i `kotl-mode` copies the left margin and found that it is bound to `hui-kill-ring-save` which calls `kill-region-as-kill` directly. We have `kotl-mode:copy-region-as-kill` that seems to do the proper copying without picking up the left margin whitespace. This PR uses that function for copying the region when in `kotl-mode.`


matsl:
Trying to write a test case reveals some problems with this change so working on it.

PR #298, opened 2023-01-20
Silence byte compilation warnings for interaction-log

matsl:
## What

Silence byte compilation warnings for interaction-log

## Why

`hypb:activate-interaction-log-mode` uses many variables and functions from the interaction-log package. This produces many byte compilation warnings. This uses the technique to use `defvar` and `declare-function` to tell the byte compiler that these things exists.

This is a bit verbose way to accomplish this but is there any other way? One idea is to provide these utility functions as a file that is not included in the files that are compiled. Rather a separate file to be loaded by the user and as such could require the interaction-log package? wdyt?

See notes below for other issues.

### DIFF
--- a/hypb.el
+++ b/hypb.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     6-Oct-91 at 03:42:38
-;; Last-Mod:     10-Dec-22 at 00:52:04 by Mats Lidell
+;; Last-Mod:     21-Jan-23 at 00:38:07 by Mats Lidell
 ;;
 ;; Copyright (C) 1991-2022  Free Software Foundation, Inc.
 ;; See the "HY-COPY" file for license information.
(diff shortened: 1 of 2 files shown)


hypb.el
```diff
@@ -99,8 +121,7 @@ This displays a clean log of Emacs keys used and commands executed."
 	  (setq ilog-truncation-timer (run-at-time 30 30 #'ilog-truncate-log-buffer))
 	  (setq ilog-insertion-timer (run-with-timer ilog-idle-time ilog-idle-time
 						     #'ilog-timer-function))
-	  (message "Interaction Log: started logging in %s" ilog-buffer-name)
-	  (easy-menu-add ilog-minor-mode-menu))
```
matsl, line 103: `easy-menu-add` is marked as obsolete in 28.1 and has, according to the docs, always been a no-op in Emacs so can safely be removed.

hypb.el
```diff
@@ -51,6 +51,28 @@ It must end with a space."
...
+(declare-function ilog-note-buffer-change "ext:interaction-log")
+(declare-function ilog-post-command "ext:interaction-log")
+(declare-function ilog-record-this-command "ext:interaction-log")
+(declare-function ilog-show-in-other-frame "ext:interaction-log")
+(declare-function ilog-timer-function "ext:interaction-log")
+(declare-function ilog-toggle-view "ext:interaction-log")
+(declare-function ilog-truncate-log-buffer "ext:interaction-log")
+(declare-function interaction-log-mode "ext:interaction-log")
+
+(declare-function pushnew "ext:???") ;; FIXME - Unknown function
```
matsl, line 74: I don't think the function `pushnew` is defined in any Emacs so ought to be replace with something else.

rswgnu (approved):

PR #299, opened 2023-01-27
Adding installation instructions using git or git+borg

PlasmaShift:
I was looking at the info directory and I would need more time
to figure out the syntax so that part still needs done

I mixed Manual git and borg instructions in this change, They might want to be parsed out.

Notes I found while doing this
If you have Staight all you need is
(straight-use-package hyperbole)
and nothing else, I don't know where it gets the recipe from or if you want to override it.
If you use Elpaca it dosn't work without some additional work
https://github.com/progfolio/elpaca

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -178,6 +178,53 @@ lines to your personal Emacs initialization file,
 Restart Emacs and you should see the "Hypb" hyperbole minor mode
 indicator in your modeline after startup.
 
+** Building manually from git (using git submodules and/or borg)
+
+If you prefer work directly with the git and the raw files with a
+minimal package manager or none at all
+
+1. Clone the Hyperbole repo in your emacs config directory
+run borg-assimalte on the Hyperbole package to install it
+{M-x borg-assimilate RET hyperbole RET}
+
+or from the cli in your git directory
+git submodule add --name compat \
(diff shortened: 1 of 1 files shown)


rswgnu (changes requested):
Hi Matthew:  As you noted, this needs some work, e.g. typo fixing, looking at paths.  I suggest we look at it on the Sunday meeting, make updates and then get it merged in.  Good news on the simplicity of loading Hyperbole with straight.  If that is so simple, what does all of this setup with borg get you instead?

matsl:
Emacs initialization can be a bit tricky. My basic initialization code looks like this. 
```
;; Use own version for development if it exists
(when (file-exists-p "~/src/hyperbole")
  (add-to-list 'load-path "~/src/hyperbole"))

(require 'hyperbole)
(hyperbole-mode 1)
(hyperbole-toggle-messaging 1)
(setq action-key-modeline-function #'hmouse-context-ibuffer-menu)

(put 'hmouse-toggle-bindings 'disabled nil)
(hmouse-add-unshifted-smart-keys)
```

When inspecting the load path the kotl folder is there!? So seems with this I don't need to do anything more specific for Hyperbole to be set up properly.

PR #300, opened 2023-01-29
Remove more warnings

matsl:
## What

Remove more warnings

## Note

This PR removes more warnings from different places. It is split in many small commits all standalone. The changes are small though so going through the PR as normal should also work.

### DIFF
--- a/hib-kbd.el
+++ b/hib-kbd.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    22-Nov-91 at 01:37:57
-;; Last-Mod:     29-Jan-23 at 10:11:10 by Bob Weiner
+;; Last-Mod:     29-Jan-23 at 22:45:45 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -129,7 +129,7 @@ Any key sequence within the series must be a string of one of the following:
 		   ;; match to things like ${variable}.
 		   (when (memq (char-before start) '(nil ?\ ?\t ?\n ?\r ?\f ?\"))
 		     (when (and (stringp key-series)
-				(not (eq key-series "")))
+				(not (equal key-series "")))
 		       ;; Replace any ${} internal or env vars; leave
 		       ;; $VAR untouched for the shell to evaluate.
 		       (let ((hpath:variable-regexp "\\${\\([^}]+\\)}"))
@@ -270,7 +270,7 @@ Any key sequence within the series must be a string of one of the following:
 			 (kbd-key:remove-delimiters str "‘" "’")))
 	 binding)
     (when (and (stringp key-series)
-	       (not (eq key-series "")))
+	       (not (equal key-series "")))
       (setq key-series (kbd-key:normalize key-series)
 	    binding (kbd-key:binding key-series)))
     (and (stringp key-series)
--- a/hib-kbd.el
+++ b/hib-kbd.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    22-Nov-91 at 01:37:57
-;; Last-Mod:     29-Jan-23 at 22:45:45 by Mats Lidell
+;; Last-Mod:     29-Jan-23 at 22:46:13 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -129,7 +129,7 @@ Any key sequence within the series must be a string of one of the following:
 		   ;; match to things like ${variable}.
 		   (when (memq (char-before start) '(nil ?\ ?\t ?\n ?\r ?\f ?\"))
 		     (when (and (stringp key-series)
-				(not (equal key-series "")))
+				(not (string-equal key-series "")))
 		       ;; Replace any ${} internal or env vars; leave
 		       ;; $VAR untouched for the shell to evaluate.
 		       (let ((hpath:variable-regexp "\\${\\([^}]+\\)}"))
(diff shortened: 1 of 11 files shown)


hib-kbd.el
```diff
@@ -127,7 +127,7 @@ Any key sequence within the series must be a string of one of the following:
 		   ;; match to things like ${variable}.
 		   (when (memq (char-before start) '(nil ?\ ?\t ?\n ?\r ?\f ?\"))
 		     (when (and (stringp key-series)
-				(not (eq key-series "")))
+				(not (equal key-series "")))
```
rswgnu, line 130: Two empty strings always compare 'eq' to me.  Can you create a scenario where this test fails?

hib-kbd.el
```diff
@@ -268,7 +268,7 @@ Any key sequence within the series must be a string of one of the following:
 			 (kbd-key:remove-delimiters str "‘" "’")))
 	 binding)
     (when (and (stringp key-series)
-	       (not (eq key-series "")))
+	       (not (equal key-series "")))
```
rswgnu, line 271: Same comment as above.

hibtypes.el
```diff
@@ -1021,7 +1021,7 @@ xdb.  Such lines are recognized in any buffer."
         ;; For Meteor app errors, remove the "app/" prefix which
         ;; is part of the build subdirectory and not part of the
         ;; source tree.
-        (when (and (not (eq col-num "")) (string-match "^app/" file))
+        (when (and (not (equal col-num "")) (string-match "^app/" file))
```
rswgnu, line 1024: Same comment as above.

hui-mini.el
```diff
@@ -256,7 +256,7 @@ If on the menu name prefix or the first item, move to the last item."
 The documentation is formatted.  With optional HELP-STRING-FLAG,
 instead returns the one line help string for the key sequence."
   (when (and (stringp key-sequence)
-	     (not (eq key-sequence ""))
+	     (not (equal key-sequence ""))
```
rswgnu, line 259: Same comment as above.

hyrolo.el
```diff
@@ -1392,7 +1392,7 @@ Return number of matching entries found."
 		 (when (file-exists-p hyrolo-file-or-buf)
 		   (setq actual-buf (hyrolo-find-file-noselect hyrolo-file-or-buf)
 			 new-buf-p t))))
-	(let ((hdr-pos) (num-found 0) (curr-entry-level-len)
+	(let ((hdr-pos) (num-found 0)
```
rswgnu, line 1395: curr-entry-level-len was added for some work I am doing in hynote.el (not yet published) that utilizes the hyrolo.el code.  Let's leave it for now even though it is not yet used.

kotl/kexport.el
```diff
@@ -384,11 +384,7 @@ used.  Also converts Urls and Klinks into Html hyperlinks.
 		 (buffer-name (find-file-noselect export-from)))
 		(t (error
 		    "(kexport:html): `%s' is an invalid `export-from' argument" export-from))))
-	 (font-lock-auto-fontify) ;; Prevent syntax highlighting
```
rswgnu, line 387: Need some explanation of these removals.  Let's discuss tomorrow/today.

kotl/kview.el
```diff
@@ -485,7 +485,7 @@ If between kcells, move to the previous one.  The current cell may be hidden."
   (if (save-excursion
 	(goto-char (line-beginning-position))
 	(looking-at kview:outline-regexp))
-      (setq found (goto-char (- (match-end 0) 2)))
```
rswgnu, line 488: Need to discuss all the changes in this file too.

rswgnu (changes requested):

hib-kbd.el
matsl: I think this warning could originate from this page in the manual.
https://www.gnu.org/software/emacs/manual/html_node/elisp/Equality-Predicates.html

hyrolo.el
matsl: OK. If possible try to avoid adding code that is not used. Just adding a comment could be enough it is work in progress so I know the purpose.

kotl/kexport.el
matsl: It is in the commit message. From my investigation these are from XEmacs font lock and is not available in Emacs.

matsl: @rswgnu Will try to find if there are any font lock functionality in current Emacs that we need to address here. Can we add that as a todo to not block this PR since holding off removing these warnings will not make current Emacs work better.

rswgnu:
Yes, will do, adding to todo list for you to research and then not
holding up the PR for this reason.

rswgnu:
Asked for 3 more changes and then we should be good on this one.  See inline.

hyrolo.el
matsl: Removing this commit.

matsl:
@rswgnu I think I addressed all things we talk about . PTAL

rswgnu (approved):
Just finish with the defvar comment at the bottom and I think this is ready for merging; you can just do so.

kotl/kview.el
```diff
@@ -1013,7 +1013,6 @@ See also `kview:map-branch' and `kview:map-tree'."
   (with-current-buffer (kview:buffer kview)
     (save-excursion
       (let ((lbl-sep-len (kview:label-separator-length kview))
-	    cell-indent
```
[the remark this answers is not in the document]
matsl: @rswgnu Came back to this first today and, since time has passed since we spoke, I'm not sure what we were thinking of here. Could it be that we make cell-indent into a defvar variable? Maybe name it with a prefix like `kview--cell-indent`?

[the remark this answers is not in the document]
matsl: @rswgnu As I understood today's talk I removed all things related to `cell-indent`.

rswgnu (approved):

PR #302, opened 2023-02-05
Add tests for Hyperbole set library

matsl:
## What

Add tests for Hyperbole set library.

## Why

Tests are good and is a start for refactoring the set library using Emacs primitives. Also raises some questions about the expected functionality in current set.el. See comments `FIXME` in the PR for discussion.

### DIFF
--- a/test/set-tests.el
+++ b/test/set-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell
 ;;
 ;; Orig-Date:     5-Feb-23 at 09:12:52
-;; Last-Mod:      6-Feb-23 at 21:28:29 by Mats Lidell
+;; Last-Mod:      6-Feb-23 at 21:33:27 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -18,7 +18,7 @@
 
 (require 'set)
 
-(ert-deftest set-tests--function-tests ()
+(ert-deftest set-tests--function-tests-equal ()
   "Test Hyperbole set library functions."
   (should (set:equal (set:create) nil))
   (should (set:empty (set:create)))
@@ -50,6 +50,10 @@
   (should (set:equal (set:combinations (set:create 'a 'b)) '(nil a b (a b))))
   (should (set:equal (set:combinations (set:create 'a 'b 'c))
                      '(nil a b c (a b) (a c) (b c) (a b c))))
+
+  (should (set:equal (set:combinations (set:create 'a 'b 'c) 1) (set:create 'a 'b 'c)))
+  (should (set:equal (set:combinations (set:create 'a 'b 'c) 2) (set:create '(a b) '(b c) '(a c))))
+  (should (set:equal (set:combinations (set:create 'a 'b 'c) 3) (set:create '(a b c))))
 
   (should (set:empty (set:intersection (set:create) (set:create))))
   (should (set:empty (set:intersection (set:create) (set:create 'a))))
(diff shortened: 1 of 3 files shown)


test/set-tests.el
```diff
@@ -0,0 +1,125 @@
...
+  
+  ;; set:members works on lists!?
+  ;; FIXME: Use verification that list contains elements rather than checking
+  ;; a specific order
+  (should (equal (set:members '(1)) '(1)))
+  (should (equal (set:members '(1 2)) '(2 1)))
+  (should (equal (set:members '(1 1)) '(1)))
+  (should (equal (set:members '(1 1 2 2)) '(2 1))))
+
+(provide 'set-tests.el)
```
rswgnu, line 124: No .el

test/set-tests.el
```diff
@@ -0,0 +1,125 @@
...
+  (should-not (set:subset (set:create 'a) (set:create 'b)))
+
+  ;; set:get - requires elements to be of type (key . value)
+  (should (equal (set:get 'a (set:create (cons 'a 'value-a))) 'value-a))
+  (should (equal (set:get 'b (set:create (cons 'a 'value-a))) nil))
+
+  ;; set:replace - requires elements to be of type (key . value)
+  ;; FIXME: Is this the expected functionality? Should not new-value-a
+  ;; really replace value-a?
+  (should (set:equal (set:replace 'a 'new-value-a (set:create (cons 'a 'value-a)))
```
rswgnu, line 108: Yes, replace should replace elements.

rswgnu (changes requested):

test/set-tests.el
```diff
@@ -0,0 +1,125 @@
+;;; set-tests.el --- Hyperbole set library tests                -*- lexical-binding: t; -*-
```
rswgnu, line 1: Some additional tests to add:
Test with set:equal-op set to different values.
Test that a set-like object with duplicate values is not a set.

Since you test the same sets repeatedly with non-modifying tests,
I would just store the set in a variable where applicable.

test/set-tests.el
```diff
@@ -0,0 +1,145 @@
...
+  (should (set:equal (set:add 'a (set:create 'a)) (set:create 'a)))
+  ;; FIXME: set:add can't take an s-expr!? Wrong type argument
+  ;; (should (set:equal (set:add 'b (set:create 'a)) (set:create 'a 'b)))
+
+  (should (set:empty (set:remove 'a (set:create 'a))))
+  (should (set:equal (set:remove 'a (set:create 'b)) (set:create 'b)))
+  ;; FIXME: This fails
+  ;; (should (set:equal (set:remove 'a (set:create 'a 'b)) (set:create 'b)))
+  ;; This works but is wrong!?
+  (should (set:equal (set:remove 'a (set:create 'a 'b)) (set:create 'a 'b)))
```
matsl, line 112: @rswgnu I noticed this problem as well. The test case that should work is commented out and the one that works is left for illustrating what it is wrong. Remove should remove 'a from the set, right?

On add and remove. Why are they defined as macros? Can they not just be ordinary functions?

rswgnu:
Hi Mats: I did most of the updates needed in set.el including 2 new functions, then I pulled your set-tests.el and updated the comments as a number of the FIXMEs were resolved.  You just need to rebase against master and then add a few more tests in the set-tests.el file (look for remaining FIXMEs) and that should be it.

test/set-tests.el
matsl: I suggest we can look over refactoring in a separate PR. Makes it easier to work in increments.

matsl:
@rswgnu I have added some tests but I also had to do some updates to set.el. PTAL.
I tried to generalize the tests so they could be reused with different equal operations but that turned into a mess. I would also prefer to add these tests or any other refactoring in additional PRs. 

Designing test cases for a set using `eq` is tricky. You need to think carefully when designing the test cases.

rswgnu (approved):

PR #305, opened 2023-02-11
Add tests for key-series in help-mode

matsl:
## What

Add tests for key-series in help-mode

## Why

These key series breaks in help-mode. When that is fixed these tests should work.

## Note

The changed tests only runs in interactive mode so can not be triggered here. If we can find the root cause of this we might be able to design test cases that can run in non-interactive batch mode.


rswgnu (approved):

matsl:
See #310 for new version.

PR #306, opened 2023-02-12
Remove some warnings

matsl:
## What

Remove some warnings.

## Why

When looking at producing the set of warnings we have now I spotted a few easy fixes.  Each is in its own commit for easy review.

### DIFF
--- a/hyrolo.el
+++ b/hyrolo.el
@@ -675,7 +675,7 @@ Return t if entry is killed, nil otherwise."
 	  (setq file buffer-file-name)
 	  (if (file-writable-p file)
 	      (let ((kill-op
-		     (lambda (start level-len)
+		     (lambda (start _level-len)
 		       (kill-region
 			start (hyrolo-to-entry-end t))
 		       (setq killed t)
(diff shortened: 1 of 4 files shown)


kotl/kexport.el
```diff
@@ -462,7 +462,7 @@ used.  Also converts Urls and Klinks into Html hyperlinks.
 		 ">" "&gt;"
 		 (replace-regexp-in-string
 		      "<" "&lt;" (kview:label-separator kview))))
-	       i is-parent is-collapsible is-last-sibling no-sibling-stack level label contents)
```
matsl, line 465: Seems I added this while you removed using it so now I remove it.

hyrolo.el
```diff
@@ -1424,7 +1424,6 @@ Return number of matching entries found."
 		  (re-search-backward hyrolo-entry-regexp nil t)
 		  (setq start (point))
 		  (re-search-forward hyrolo-entry-regexp nil t)
-		  (setq curr-entry-level-len (length (buffer-substring-no-properties start (point))))
```
matsl, line 1427: This is only set here but not used!?

Can this be code you are working on and that these var can be used later?

rswgnu: You are correct.  This is code I inserted for compatibility with the in-development hynote.el and it is not yet finished.  Please ignore it for now and remove this change.

rswgnu (changes requested):
All changes are fine except the one I commented on.  Remove that and we can apply this.

PR #307, opened 2023-02-15
Make kotl-autoloads without building hyperbole-autoloads.el

matsl:
## What

Make `kotl/kotl-autoloads` without building `hyperbole-autoloads.el`

## Why

The target `kotl/kotl-autoloads.el` has as a side effect to also build the `hyperbole-autoloads.el.` That is not correct and can lead to unexpected problems and even errors. Patch by Stefan Monnier.  This PR fixed this so that only `kotl-autoloads` are produced as a result of that make target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Author:       Bob Weiner
 #
 # Orig-Date:    15-Jun-94 at 03:42:38
-# Last-Mod:     16-Feb-23 at 00:51:09 by Mats Lidell
+# Last-Mod:     17-Feb-23 at 00:32:14 by Mats Lidell
 #
 # Copyright (C) 1994-2023  Free Software Foundation, Inc.
 # See the file HY-COPY for license information.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -426,7 +427,7 @@ hyperbole-autoloads.el: $(EL_COMPILE)
 	$(EMACS_BATCH) --debug --eval "(progn (setq generated-autoload-file (expand-file-name \"hyperbole-autoloads.el\") backup-inhibited t) (let (find-file-hooks) (hload-path--make-directory-autoloads \".\" generated-autoload-file)))"
 
 kotl/kotl-autoloads.el: $(EL_KOTL)
-	$(EMACS_BATCH) --debug --eval "(progn (setq generated-autoload-file (expand-file-name \"kotl/kotl-autoloads.el\") backup-inhibited t) (let (find-file-hooks) (hload-path--make-directory-autoloads \"kotl/\" generated-autoload-file)))"
+	$(EMACS_PLAIN_BATCH) --debug --eval "(progn (setq generated-autoload-file (expand-file-name \"kotl/kotl-autoloads.el\") backup-inhibited t) (let (find-file-hooks) (make-directory-autoloads \"kotl/\" generated-autoload-file)))"
```
rswgnu, line 430: This breaks versions of Emacs that I have tested in the past that lack make-directory autoloads.  I will have to see if it is true of 27.1.  If not, then we can use it.

rswgnu (changes requested):

Makefile
matsl: Oh, you are right. That can be true. The interesting part is that the scraping for autoloads works for all our versions we use here but I don't think it uses that target. We don't build for 21.1 though. Only 27.2. I'll add a test to check.

matsl: The test proves it breaks on 27.2. I'll see if I can fix a workaround.

matsl: @rswgnu The autoload generation for kotl now fallbacks to Emacs 27 functionality.

rswgnu (approved):

PR #308, opened 2023-02-18
Add startup check hyperbole is found

matsl:
## What

Add startup check to ensure hyperbole is found.

## Why

Complete the install test after install; start Emacs and check that Hyperbole is found.  This has been tested with the latest elpa-devel which verifies that elpa-devel now works after the #307 PR.

## Note

One of the tests that compare file paths had to be adjusted to ignore following symbolic links in order to work with the straight package manager. 

The straight package manager requires this change to be merged to the git repo to check that it works properly.  It has been tested manually.

### DIFF
--- a/test/hmouse-drv-tests.el
+++ b/test/hmouse-drv-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    28-Feb-21 at 22:52:00
-;; Last-Mod:     22-May-22 at 11:13:48 by Mats Lidell
+;; Last-Mod:     19-Feb-23 at 15:49:31 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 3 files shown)


test/hmouse-drv-tests.el
```diff
@@ -106,15 +106,16 @@
...
-      (kill-buffer "DEMO")
-      (ibtype:delete 'ibtypes::defil-path-it))))
+  (let ((vc-follow-symlinks nil))
+    (unwind-protect
+	(with-temp-buffer
+          (insert "<<<DEMO>>>")
+          (goto-char 4)
+          (action-key)
+          (should (string= (expand-file-name "DEMO" hyperb:dir) buffer-file-name)))
+      (progn
```
rswgnu, line 116: You don't need the progn here; multiple forms are allowed, just like the else clause of an if.

rswgnu (approved):
See one inline comment to change before merging.

test/hmouse-drv-tests.el
matsl: Yes I know. I think it can be updated in multiple places since I was not aware of that when I started to write the test cases. So this is a general pattern I have been using.  Will fix it and apply and fix this in other places as well in a separate PR.

PR #309, opened 2023-02-19
Use view-mode for fast demo

matsl:
## What

Use view-mode for fast demo

## Why

There are some problems with Help-mode that make the keyseries for shell commands to not work properly. View-mode does not give the same problem so this is a quick fix and view-mode might work as well for the user.

## Note

I noticed that we do not suggest any mode when visiting file like `DEMO` or `FAST-DEMO`. We could add a Local variables section at the end suggesting view-mode?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2023-02-19  Mats Lidell  <matsl@gnu.org>
+
+* hypb.el (hypb:display-file-with-logo): Use view-mode for fast demo.
+
 2023-02-18  Mats Lidell  <matsl@gnu.org>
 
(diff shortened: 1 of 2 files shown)


rswgnu:
View mode will be invoked when a demo file is accessed from a menu.  I don’t think we should set it if they just do a find-file, then they should get the default editing mode which will be a text mode typically and should handle all the buttons properly.-- BobOn Feb 19, 2023, at 5:14 PM, Mats Lidell ***@***.***> wrote:﻿
@matsl requested your review on: #309 Use view-mode for fast demo.

rswgnu (approved):

PR #310, opened 2023-02-19
Test for shell commands in view mode

matsl:
## What

Add tests for key-series in view-mode

## Why

Fast demo has been changed to use view mode. The test defined in #305 is here changed to use view-mode. See #309 for the actual change of opening fast demo.

## Note

The changed tests only runs in interactive mode so can not be triggered here. So need to be tested in an interactive session such as using `make test-all`


matsl:
I will rebase,  to resolve the ChangeLog conflicts, when #309 has been approved and merged.

rswgnu (approved):

PR #313, opened 2023-03-04
Fix html checker warnings for kotl export

matsl:
## What

Fix html checker warnings for kotl export.

## Why

Generated HTML should be up to standards. 

## Note

This is a beginning of a rewrite of the `kexport:html` function. Some things can probably be better. It is still a long function that might benefit from breaking up in smaller functions. 

The resulting look might not be exactly as the previous version but that should be possible to fix by further css tweaks.

The issues that has been fixed fall in four categories:
1. Attributes in html is deprecated and have been moved to css
2. Font element is deprecated and have been moved to css
3. Elements not allowed as children of another element have been replaced with a clean html layout
4. Misc minor errors have been fixed 

There are a few FIXME's in the code that remains. PTAL at those.

The generated HTML verifies cleanly:  (Running a checker on the export from the `EXAMPLE.kotl` file)

image

~_Checked in example to be removed when we are happy with the rendering_~
~https://htmlpreview.github.io/?https://github.com/rswgnu/hyperbole/blob/fix_html_checker_warnings_for_kotl_export/EXAMPLE.html~ Removed before merge.


rswgnu:
Hi Mats:  The main concern is any change in visual appearance.  Can you do a before and after set of screen shots that you think represents any change decently.  If the look is reasonable, I am fine with applying this, though I would prefer all the FIXMEs be gone prior to a merge if possible.

matsl:
@rswgnu Added the generated html to the branch so we can review how the output looks. You can compare the difference by running the export in the master branch locally. I hope that this works. 

When we are ready to merge this we shall remove the file. That is the best way I could come up with now for easy review and update. (Open for other suggestions.)

Regarding the FIXMEs they ask review questions but I'm happy to resolve them in the simplest way if you have no big concerns on them.

kotl/kexport.el
```diff
@@ -43,15 +43,20 @@
 (defvar kexport:output-filename nil
   "This is automatically set to the full pathname of the exported file.")
 
+;; FIXME - set in css
 (defcustom kexport:html-body-attributes
```
rswgnu, line 47: I think you should be able to customize this from Lisp.  What is the problem here if we leave it as is?

kotl/kexport.el
```diff
@@ -43,15 +43,20 @@
...
   "BGCOLOR=\"#FFFFFF\"" ;; white background
   "*String of HTML attributes attached to the <BODY> tag.
 Part of an HTML exported koutline file."
   :type 'string
   :group 'hyperbole-koutliner)
 
+;; FIXME - We can't have quotes in the description field if we have
+;; the intent that users should be able to use the button from the
+;; generated html. In the generated html they must be inserted like
+;; this '&quot'
```
rswgnu, line 57: This change is fine and the FIXME can be removed.  I am adding changes so &quot; will be recognized as path and Info-node delimiters when editing HTML in Emacs.

Another level of improvement would be if we turned these into hrefs that pointed to the GNU HTML versions of their Texinfo manuals.  See https://www.gnu.org/manual/manual.en.html.

kotl/kexport.el
```diff
@@ -300,30 +332,40 @@ menuitem > menu > menuitem.hover > menu > menuitem{
...
+  allSpan[x].onclick=function()
+  {
+    if(this.parentNode)
+    {
+      var icon = childElt(this, 'span');
+      var childList = this.parentNode.getElementsByTagName('li');
+      for(var y = 0; y < childList.length; y++)
+      {
+        var currentState = childList[y].style.display;
+        if(currentState=='none')
```
rswgnu, line 354: Why not this whitespacing style:  "if (currentState == 'none')"?

Is it wrong?

kotl/kexport.el
```diff
@@ -422,110 +464,81 @@ used.  Also converts Urls and Klinks into Html hyperlinks.
 	  (when (string-match "\n" title)
 	    (setq title (substring title 0 (match-beginning 0)))))
 
-	(princ "<html><head>\n\n")
-	(princ "<a id=\"top\"></a><a id=\"k0\"></a>\n")
-	(princ (format "<title>%s</title>\n" title))
+	(princ "<!DOCTYPE html>\n")
+        ;;; FIXME make lang configurable!?
+	(princ "<html lang=\"en\">\n")
```
rswgnu, line 469: You are welcome to add a variable but we support English only thoughout Hyperbole, so...

kotl/kexport.el
```diff
@@ -422,110 +464,81 @@ used.  Also converts Urls and Klinks into Html hyperlinks.
...
-	;; (princ "         		</menu>
-	;; 		</menuitem>
-	;; 	</menu>
-   	;;     </nav>\n")
+
+        ;; BODY
+	(princ "<body>\n\n")
+
+        ;; FIXME: Is this needed?? Does k0 mean anything?
+	;; (princ "<div id=\"top\"></div><div id=\"k0\"></div>\n")
```
rswgnu, line 496: k0 represents the invisible top node of the outline that all level1 cells are children of.  It can have properties attached to it, which likely should be exported as meta data attached in the HTML header, though there may not presently be any code for that.

test/kexport-tests.el
```diff
@@ -96,16 +96,39 @@
...
+          (should (= (count-matches "<ul\\b" (point-min) (point-max))
+                     (count-matches "</ul\\b" (point-min) (point-max))))
+          (should (= (count-matches "<li\\b" (point-min) (point-max))
+                     (count-matches "</li\\b" (point-min) (point-max)))))
+      (delete-file kotl-file)
+      (delete-file html-file))))
+
 (ert-deftest kexport:display-creates-html-file-and-displays-it ()
   "kexport:display creates html file and displays it in external browser."
   (let* ((kotl-file (make-temp-file "hypb" nil ".kotl"))
```
rswgnu, line 134: Nice improvements here.

rswgnu (changes requested):

rswgnu:
Hi @Lidell Mats ***@***.***>:

I see 3 issues with the new HTML output when compared with the old:

1. The autogenerated cell number font-size is larger than the text size and
therefore too big.

2. There were some very tricky vertical spacing issues between cells that
we had solved.  They are now back in your new version and the cells are
separated by extra whitespace that does not look good.

3. Mouse over highlighting on collapsible tree parent cells no longer
works, so it is harder to tell there is dynamic behavior there.

kotl/kexport.el
matsl: 1. [body](https://www.w3schools.com/tags/tag_body.asp) does only support the [global attributes.](https://www.w3schools.com/tags/ref_standardattributes.asp) 
2. It is therefore not used in the code at the moment (so should really be removed in this form.)

So should ideally go into the css definition. I have not looked at what would be the best way to do that. One option would be to make the whole css customizable, another to just customize one attribute i.e. `html-body-bgcolor` etc, a third have some smarter template where multiple attributes can be set or overridden. Like a defauilt css and then add another css that will override the defaults.

kotl/kexport.el
matsl: Code is copied from html tutorial and only modified with what I needed. Did not bother about the code style. Will check if it matters and update if not.

kotl/kexport.el
matsl: Hardcoding is fine now so I will stick with that.

kotl/kexport.el
matsl: If I get it right then it does not led itself to have any tag like this. It should rather go into <head> than <body>. So it can be removed for now?

kotl/kexport.el
matsl: Changed to more readable code.

matsl:
@rswgnu Thanks for the review.

Sorry for maybe not setting the right expectations. The goal of this iteration was to create legal HTML that would have the right functionality and then focus on the css styling.

> The autogenerated cell number font-size is larger than the text size and
therefore too big.

That was a deliberate choice in this increment so I could see that I could manipulate the style with css. Will of course set this to what ever value we see fit. (Plus I rather like a bigger font on the cell numbers. ;-) 

> There were some very tricky vertical spacing issues between cells that
we had solved.  They are now back in your new version and the cells are
separated by extra whitespace that does not look good.

My plan is to solve these issues with changes to css. It was to hard to do that while working on the list structure together with getting the collapsible to work so I paused it for later.

> Mouse over highlighting on collapsible tree parent cells no longer
works, so it is harder to tell there is dynamic behavior there.

That has not been implemented yet either. I will add that soon when I work on the styling.

rswgnu:
Got it, thanks.  Let me know when you are through these and I'll have
another look.  Thanks.

PR #316, opened 2023-04-08
Do not insert ebut data section when looking for it

matsl:
## What

Do not insert ebut data section when looking for it.

## Why

A data section is inserted when executing buttons in non file buffers such as *scratch*, news summary or temp buffers. Test case for temp buffer with non existing ebut added. (Not an ideal test case since verifying that something does not happen is not the best test. The test can be applied locally before trying the fix to demonstrate that it happens.)

The fix is more of a hack. It locally check if the create flag is set, using the fact it should not be set when only looking for if button data exists!?, and if it is not set avoids the insertion. 

But maybe we should refactor this part of the code? It feels odd to me that `ebut:at-p` would even come close to code that does insertion. Also the explicit checks in there on the email and news interface looks like a code smell to me.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2023-04-22  Mats Lidell  <matsl@gnu.org>
+
+* test/hbut-tests.el
+    (hypb--ebut-at-p-should-not-insert-hbdata-section-in-non-file-buffers):
+    Add test to verify that no but data section is inserted.
+
 2023-04-16  Bob Weiner  <rsw@gnu.org>
 
 * hui.el (hui:link-possible-types): Fix link to outline heading where
--- a/hbdata.el
+++ b/hbdata.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     2-Apr-91
-;; Last-Mod:     29-Mar-23 at 21:04:06 by Bob Weiner
+;; Last-Mod:      9-Apr-23 at 01:22:24 by Mats Lidell
 ;;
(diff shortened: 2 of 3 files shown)


rswgnu (approved):

rswgnu:
I believe I fixed this with changes that I pushed last week.  So if this works on master, just remove this branch without deploying.

rswgnu (commented):
Drop this pull request; fixed elsewhere.

matsl:
@rswgnu The test case `hypb--ebut-at-p-should-not-insert-hbdata-section-in-non-file-buffers` still fails on the master branch. So I brought the PR back and rebased. I have also verified manually that the separator is still inserted on the master branch.

rswgnu:
The ebut separator is supposed to be added to buffers without files such as email and lisp interaction buffers, so it seems to me the test case is not correct.  

Adding the 'create' flag usage looks correct.

matsl:
Why should anything be added when checking if there is a button or trying to use what looks like a button? Should not that be a read only operation? That the ebut separator is missing should not that be taken as an indication that there is no valid button rather than inserting it just to realize that there actually is no button data?

rswgnu (approved):

matsl:
@rswgnu I have some vague memory that you wanted something more to be added to this PR but coming back to this now for the weekend I can't remember what it was.

rswgnu (approved):

rswgnu:
Just approved.  Thanks.On Apr 21, 2023, at 7:14 PM, Mats Lidell ***@***.***> wrote:﻿
@rswgnu I have some vague memory that you wanted something more to be added to this PR but coming back to this now for the weekend I can't remember what it was.

PR #317, opened 2023-04-10
Add tests for kimport

matsl:
## What

Add tests for kimport

## Why

More tests.

## Note

Does not currently work due to problems with import. Should work after master gets fixed.


rswgnu:
Please update to base branch and then merge.

PR #318, opened 2023-04-10
Update and add kotl-mode tests

matsl:
## What

Update and add `kotl-mode` tests

##  Why 

More tests...

## Note

This is work in progress, more tests will be added but feel free to suggest what tests to add.


rswgnu:
I would use the EXAMPLE.kotl file and pick features described there that seem useful and testable.  Maybe you just make a few levels of priorities, label the features and then do 1 priority level at a time, across time.

rswgnu (approved):
I have fixed the error and pushed my changes to this branch, so you can merge this now.

PR #320, opened 2023-04-23
Silence various byte compiler warnings

matsl:
## What

Fix a mix of different types of byte compiler warnings.

## Why

Taking yet another stab at getting the warnings down to a low number.

### DIFF
--- a/hbdata.el
+++ b/hbdata.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     2-Apr-91
-;; Last-Mod:      9-Apr-23 at 01:22:24 by Mats Lidell
+;; Last-Mod:     26-Apr-23 at 00:11:40 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -55,6 +55,15 @@
 (require 'hversion) ;; For hyperb:microsoft-os-p
 (require 'hbmap)
 (require 'hgnus)
+
+;;; ************************************************************************
+;;; Public declarations
+;;; ************************************************************************
+
+;; Functions from abstract mail and news interface. See "hmail.el"
+(declare-function lmail:to nil)
+(declare-function rmail:to nil)
+(declare-function rmail:summ-msg-to nil)
 
 ;;; ************************************************************************
 ;;; Public functions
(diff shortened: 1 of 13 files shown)


hbdata.el
```diff
@@ -314,7 +323,7 @@ If the hbdata buffer is blank/empty, kill it and remove the associated file."
 (defun hbdata:delete-entry-at-point ()
   (delete-region (point) (progn (forward-line 1) (point))))
 
-(defun hbdata:ibut-instance-last (lbl-key key-src &optional directory)
+(defun hbdata:ibut-instance-last (lbl-key _key-src &optional _directory)
   "Return highest instance number for repeated implicit button label.
 1 if not repeated, nil if no instance.
 Utilize arguments LBL-KEY, KEY-SRC and optional DIRECTORY."
```
rswgnu, line 329: Need to rename args here too with underscore.

hui-mouse.el
```diff
@@ -1067,7 +1083,8 @@ a helm section header."
 
 (defun smart-helm-get-current-action (&optional action)
   "Return the helm default action.
-Get it from optional ACTION, the helm saved action or from the selected helm item."
+Get it from optional ACTION, the helm saved action or from the
+selected helm item."
```
rswgnu, line 1087: First doc line has to be complete by itself, so that apropos works properly.

hycontrol.el
```diff
@@ -1732,13 +1738,13 @@ columns (rightmost) of the grid."
 ;;;###autoload
 (defun hycontrol-windows-grid-by-file-pattern (arg pattern &optional full-flag)
   "Display up to an abs(prefix ARG)-sized window grid of files matching PATTERN.
-Use absolute file paths if called interactively or optional FULL-FLAG is non-nil.
-PATTERN is a shell glob pattern.
+Use absolute file paths if called interactively or optional
+FULL-FLAG is non-nil.  PATTERN is a shell glob pattern.
```
rswgnu, line 1742: Make first line standalone.

rswgnu (changes requested):

rswgnu:
Looks good, otherwise.

hbdata.el
matsl: Good catch. Thanks. Maybe remove them from the doc since they are not used?

hui-mouse.el
matsl: I understand. For this PR though - The firs line is not changed.

hycontrol.el
matsl: Same here. First line is not altered.

rswgnu: But the last part of the first line shows up in the changed lines as part of the second line, so the line is getting wrapped on you somehow.

hbdata.el
rswgnu: I am not sure why those args are there then; maybe copied from the ebut function of the same kind and then called from an hbut genetic fun tion that requires the same signature.  Fine to remove them from the doc.

matsl: Yes, We can look at that separately. I will remove them from the docs for now.

hycontrol.el
matsl: Is not line 1740 the first line? "Display up to an abs(prefix ARG)-sized window grid of files matching PATTERN."?

rswgnu: Oh, the first line was elided, so I was commenting on the second line by mistake.  Appoved now.

rswgnu:
It won't let me.  Just merge.

matsl:
Hi Bob,

Ok. Thanks for trying. Will merge.

Yours

Robert Weiner ***@***.***> skrev: (27 april 2023 02:26:31 CEST)
>It won't let me.  Just merge.
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/rswgnu/hyperbole/pull/320#issuecomment-1524283893
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>
-- 
## Mats

PR #321, opened 2023-04-30
Add ebut rename tests

matsl:
## What

Add ebut rename tests.

## Why

More test cases are good and we have a bug with ebut rename with ebuts that has been made unique due to autonumbering of labels. This includes a test of the bug that is disable and should be able to use when the bug has been fixed.

### DIFF
--- a/test/hui-tests.el
+++ b/test/hui-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    30-Jan-21 at 12:00:00
-;; Last-Mod:      7-Apr-23 at 12:19:58 by Mats Lidell
+;; Last-Mod:     30-Apr-23 at 11:15:21 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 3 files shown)


test/hui-tests.el
```diff
@@ -639,6 +639,54 @@ With point on label suggest that ibut for rename."
...
+    (goto-char (point-min))
+    (should (equal (hattr:get (hbut:at-p) 'lbl-key) "label"))
+    (hui:ebut-rename "label" "new")
+    (should (equal (hattr:get (hbut:at-p) 'lbl-key) "new"))
+    (goto-char (- (point-max) 1))
+    (should (equal (hattr:get (hbut:at-p) 'lbl-key) "label2"))))
+
+(ert-deftest hui--ebut-rename-nonumbered-label ()
+  "Rename an ebut shall rename the label with no number."
+  (skip-unless false) ;; Skip until label rename bug is fixed
```
matsl, line 665: Test case disabled until the bug is fixed.

test/hui-tests.el
```diff
@@ -639,6 +639,54 @@ With point on label suggest that ibut for rename."
...
+    (should (equal (hattr:get (hbut:at-p) 'lbl-key) "label"))
+    (hui:ebut-rename "label" "new")
+    (should (equal (hattr:get (hbut:at-p) 'lbl-key) "new"))))
+
+(ert-deftest hui--ebut-rename-only-button-with-that-label ()
+  "Rename an ebut shall change the name of only button with that label."
+  (with-temp-buffer
+    (ebut:program "label" 'link-to-directory "/tmp")
+    (goto-char (point-max))
+    (ebut:program "label2" 'link-to-directory "/tmp")
```
rswgnu, line 655: In addition to testing as you do here when a label is a substring of another label, you need the additional test utilizing hypb:program-create-ebut-in-buffer-with-same-label, where you test that only the original button and not its versions changed.  Also need to test if multiple occurrences of the same label name exist (no version or exact same version) that all of them are renamed (within the same buffer).

rswgnu (changes requested):

test/hui-tests.el
matsl: > In addition to testing as you do here when a label is a substring of another label, you need the additional test utilizing hypb:program-create-ebut-in-buffer-with-same-label, where you test that only the original button and not its versions changed.

The tests `hui--ebut-rename-nonumbered-label` and `hui--ebut-rename-numbered-label` tests the cases with numbered labels. Either rename the one with a number or the one no number. It might look odd but ebut:program with same label creates buttons with numbered labels.

matsl: >  Also need to test if multiple occurrences of the same label name exist (no version or exact same version) that all of them are renamed (within the same buffer).

Will add that. What would be proper way to create two buttons with the same label.  Copy the region?

test/hui-tests.el
```diff
@@ -639,6 +639,66 @@ With point on label suggest that ibut for rename."
...
+    (goto-char (point-max))
+    (ebut:program "label" 'link-to-directory "/tmp")
+    (goto-char (- (point-max) 1))
+    (should (equal (hattr:get (hbut:at-p) 'lbl-key) "label:2"))
+    (hui:ebut-rename "label:2" "new")
+    (should (equal (hattr:get (hbut:at-p) 'lbl-key) "new"))
+    (goto-char (point-min))
+    (should (equal (hattr:get (hbut:at-p) 'lbl-key) "label"))))
+
+(ert-deftest hui--ebut-rename-all-copies ()
```
matsl, line 689: @rswgnu Adds the rename all test case.

test/hui-tests.el
```diff
@@ -639,6 +639,66 @@ With point on label suggest that ibut for rename."
...
+  "Rename an ebut shall rename all copies."
+  (with-temp-buffer
+    (ebut:program "label" 'link-to-directory "/tmp")
+    (end-of-line)
+    (hui-kill-ring-save (point-min) (point))
+    (yank)
+    (goto-char (point-min))
+    (should (looking-at-p "<(label)><(label)>"))
+    (hui:ebut-rename "label" "new")
+    (goto-char (point-min))
```
matsl, line 699: Rename moves point which might be acceptable so have to move it back.

test/hui-tests.el
rswgnu: Correct, just copy the button text and delimiters.  You can have as many copies as you want in the same buffer.

rswgnu (approved):

PR #322, opened 2023-05-01
Stefan scratch hyperbole updates

matsl:
## What

Stefan updates from scratch hyperbole branch on elpa rebased onto master.

## Why

Before splitting in smaller pieces I supply the full patch as one PR since I think going over each commit might speed things up. I think each commit stands out quite well on its own with only small dependencies between the commits. If you can review this PR and suggest what commit to pick from it I can create new PRs for those. 

Some of the changes comes in the form of  FIXMEs so is something we might want to address rather than commit?

### DIFF
--- a/hbut.el
+++ b/hbut.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    18-Sep-91 at 02:57:09
-;; Last-Mod:     22-May-23 at 23:06:54 by Bob Weiner
+;; Last-Mod:     28-May-23 at 15:56:07 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -373,19 +373,16 @@ button is found in the current buffer."
 		    (insert new-label)
 		    (setq end (point)))
 		   ((and (hmouse-use-region-p)
-			 (if (hyperb:stack-frame
-			      '(hui:ebut-create hui:ebut-edit hui:ebut-edit-region
-						hui:ebut-link-create hui:gbut-create
-                       				hui:gbut-edit hui:link-create ebut:program
-						hui:ibut-create hui:ibut-edit
-						hui:ibut-link-create ibut:program))
+			 (if hui--ignore-action-key-depress-prev-point
 			     ;; Ignore action-key-depress-prev-point
 			     (progn (setq mark (marker-position (mark-marker))
 					  start (region-beginning)
 					  end (region-end)
 					  buf-lbl (buffer-substring-no-properties start end))
 				    (equal buf-lbl curr-label))
-			   ;; Utilize any action-key-depress-prev-point
+			   ;; Utilize any action-key-depress-prev-point.
+			   ;; FIXME: Can't `action-key-depress-prev-point'
+			   ;; be nil at this point?
 			   (setq mark (marker-position (mark-marker)))
 			   (setq prev-point (and action-key-depress-prev-point
 						 (marker-position action-key-depress-prev-point)))
@@ -443,8 +440,9 @@ the button LABEL which is automatically provided as the first argument.
 
 For interactive creation, use `hui:ebut-create' instead."
   (save-excursion
-     (let ((but-buf (current-buffer))
-	   (actype-sym (actype:action actype)))
+    (let ((but-buf (current-buffer))
+	  (hui--ignore-action-key-depress-prev-point t)
+	  (actype-sym (actype:action actype)))
       (hui:buf-writable-err but-buf "ebut-create")
       (condition-case err
 	  (progn
@@ -1797,6 +1795,7 @@ Return symbol for button deleted or nil."
   "Delimit implicit button label spanning region START to END in current buffer.
 If button is already delimited or delimit fails, return nil, else t.
 Insert INSTANCE-FLAG after END, before ending delimiter."
+  ;; FIXME: Merge with `ebut:delimit'!
   (goto-char start)
   (when (looking-at (regexp-quote ibut:label-start))
     (forward-char (length ibut:label-start)))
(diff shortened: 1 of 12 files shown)


hbut.el
```diff
@@ -377,17 +377,16 @@ button is found in the current buffer."
 		    (insert new-label)
 		    (setq end (point)))
 		   ((and (hmouse-use-region-p)
-			 (if (hyperb:stack-frame
-			      '(hui:ebut-create hui:ebut-edit hui:ebut-edit-region hui:gbut-create
-                       				hui:gbut-edit hui:link-create ebut:program
-						hui:ibut-create hui:ibut-edit ibut:program))
+			 (if hui--ignore-action-key-depress-prev-point
```
rswgnu, line 380: Unsure about this change, have to look at it more.  If I view a file and don't add a comment, that means it is good for a PR (once I submit my review).  -- rsw

hibtypes.el
```diff
@@ -46,6 +46,7 @@
 ;;; Public variables
 ;;; ************************************************************************
 
+;; FIXME: The `mail-' prefix does not belong to Hyperbole!
```
rswgnu, line 49: I'm fine with adding an hypb- prefix here.

hui-mouse.el
```diff
@@ -128,6 +128,9 @@ Its default value is `smart-scroll-down'.  To disable it, set it to
 ;;; Public declarations
 ;;; ************************************************************************
 
+;; FIXME: What makes us think these functions will be available when we
+;; call them?
+
 (declare-function todotxt-archive "ext:todotxt")
```
rswgnu, line 134: We can't merge any FIXMEs but have to decide on handling each one.

hui-mouse.el
```diff
@@ -1426,7 +1433,10 @@ NO-RECURSE-FLAG non-nil prevents infinite recursions."
 			       ;; Does nothing unless the dash Emacs Lisp
 			       ;; library is available for the -flatten function.
 			       (and (require 'dash nil t)
-				    (assoc index-key (-flatten alist)))))))
+			            ;; FIXME: Use Emacs-27's `flatten-tree'?
```
rswgnu, line 1436: Would be fine with flatten-tree.

hui.el
```diff
@@ -514,7 +520,8 @@ See `hui:gibut-create' for details."
   (if ibut-flag
       (call-interactively #'hui:gibut-create)
     (hypb:assert-same-start-and-end-buffer
-      (let (actype
+      (let ((hui--ignore-action-key-depress-prev-point t)
```
rswgnu, line 523: These flags seem to pop up many places; seems like additional complexity to me on first glance.

hui.el
```diff
@@ -919,11 +929,13 @@ Signal an error when no such button is found in the current buffer."
 					  (ibut:label-p t) 'ibut)))))
   (unless (stringp lbl-key)
     (if (called-interactively-p 'interactive)
+	;; FIXME: Move this error to the `interactive' spec?
```
rswgnu, line 932: Sounds reasonable.

hversion.el
```diff
@@ -52,74 +52,6 @@ your specific mouse.")
...
-	      file (substitute-in-file-name file)
-	      file (locate-file file load-path
-				(if (null nosuffix) '(".elc" ".el" ".el.gz" ".el.Z"))
-				;; accept any existing file
-				nil)
-	      file (if (and (stringp file)
-			    (string-match hyperb:automount-prefixes file))
-		       (substring file (1- (match-end 0)))
-		     file)))))
-
```
rswgnu, line 122: I always thought this was pretty good and useful code.  Of course, if we don't need to call it, it would be good but we'll really have to look at the logic to be sure flags can achieve the same affect without having to annotate the code in many new places.

hypb.el
```diff
@@ -172,8 +172,8 @@ Trigger an error with traceback if the buffer is not live or its
 name differs at the start and end of BODY."
   (declare (indent 0) (debug t))
   `(let ((debug-on-error t)
-	 (start-buffer (current-buffer)))
-     (unless (buffer-live-p start-buffer)
+	 (start-buffer (current-buffer))) ;FIXME: Name capture.
+     (unless (buffer-live-p start-buffer) ;FIXME: Impossible?
        (error "Start buffer, '%s', is not live" (current-buffer)))
```
rswgnu, line 177: I doubt I added this without seeing the need for it first.  Maybe look at whether it can be better written.

kotl/kotl-mode.el
```diff
@@ -3296,10 +3296,14 @@ newlines at end of tree."
...
   (when (and (memq this-command '(self-insert-command orgtbl-self-insert-command))
-	     (eq major-mode 'kotl-mode)
 	     (not (kview:valid-position-p))
 	     ;; Prevent repeatedly moving point to valid position when moving trees
-	     (not (hyperb:stack-frame '(kcell-view:to-label-end))))
+	     ;; FIXME: This may have been useful in some older version of
+	     ;; the code but now that it runs from a `pre-command-hook',
+	     ;; `kcell-view:to-label-end' can never be in the call
+	     ;; stack (except of course during debugging and such).
+	     ;;(not (hyperb:stack-frame '(kcell-view:to-label-end)))
```
rswgnu, line 3305: Just need to look at this to ensure does not break anything.

kotl/kview.el
```diff
@@ -630,21 +630,13 @@ level."
 ;;;###autoload
 (defun kview:char-invisible-p (&optional pos)
   "Return t if the character after point is invisible/hidden, else nil."
+  ;; FIXME: A.k.a (invisible-p (or pos (point)))
```
rswgnu, line 633: If invisible-p covers the same cases, we can get rid of this, just inspect its implementation.

kotl/kview.el
```diff
@@ -708,6 +700,8 @@ are used.
 
 (defun kview:end-of-actual-line ()
   "Go to the end of the current line whether collapsed or not."
+  ;; FIXME: This "[\n\r]" is a leftover from when kotl was using
+  ;; `selective-display'.  We should use `end-of-line' nowadays.
   (when (re-search-forward "[\n\r]" nil 'move)
```
rswgnu, line 705: Makes sense; just have to be clear when to use visible vs non-visible lines, so do not break things.

rswgnu (changes requested):

hibtypes.el
matsl: Extracted into a separate [PR](https://github.com/rswgnu/hyperbole/pull/336)

kotl/kview.el
matsl: For discussion see [PR](https://github.com/rswgnu/hyperbole/pull/337)

rswgnu (approved):

matsl:
Closing this as it is superseded by #343

PR #334, opened 2023-05-24
First ibut tests

matsl:
## What

Example of ibut tests.

## Why

~Show that the tests are failing but this time with a changed approach. Tests are made to work but with asserting in some cases an erroneous response. Comments also highlights what is not working as expected.~

Switched back to checking for the right response. `hbut-tests-ibut-insert-text-temp-buffer` and `hbut-tests-ibut-insert-text-temp-file` does still contain some adoption to unexpected changes but is kept as is for now.

### DIFF
--- a/hbut.el
+++ b/hbut.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    18-Sep-91 at 02:57:09
-;; Last-Mod:     22-May-23 at 23:06:54 by Bob Weiner
+;; Last-Mod:     26-May-23 at 23:38:08 by Mats Lidell
 ;;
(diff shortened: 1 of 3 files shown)


hbut.el
```diff
@@ -2160,7 +2160,7 @@ button is found in the current buffer."
 				      (when arg2 (insert ", " arg2))
 				      (insert ">")))
       ('actypes::link-to-org-id (insert (format "\"id:%s\"" arg1)))
-      ('actypes::link-to-rfc (insert (format "rfc%d" arg1)))
+      ('actypes::link-to-rfc (insert (format "rfc%s" arg1)))
```
matsl, line 2163: The arg seems to come in a string and format does not like that for a %d.

PR #337, opened 2023-05-28
Use invisible-p

matsl:
## What

Use invisible-p

## Why

Suggested by Stefan. See the scratch [PR](https://github.com/rswgnu/hyperbole/pull/322#discussion_r1186848934)

## Note

Not sure about how overlays comes in here and if this change is too aggressive. Does the check for overlay invisible need to be kept? `inivisible-p` seems to be using text properties only. Here is the Emacs [src for invisible-p](https://git.savannah.gnu.org/cgit/emacs.git/tree/src/xdisp.c#n28873) But I'm not sure how we use overlays in this context if it matters. If that is the case then maybe `get-char-property` could be an alternative since it checks both overlays and char properties!? 

Don't think we have much unit tests to stand on here I'm afraid.

If you can advice on a good use case that uses overlays and not character properties or a mix of these I could create test cases.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2023-05-28  Mats Lidell  <matsl@gnu.org>
+
+* kotl/kview.el (kview:char-invisible-p, kview:char-visible-p): Use
+    invisible-p.
+
 2023-05-27  Bob Weiner  <rsw@gnu.org>
 
 * kotl/kotl-mode.el (kotl-mode:setup-keymap): Create this function instead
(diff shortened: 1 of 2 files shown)


rswgnu:
My guess is that overlays were used for invisibility in older versions of Emacs but probably pre-27.1, in which case we could change it.

Best way would be to search the built-in Emacs Lisp code for 'in isible' and if no overlay hits, then I van review this and move it forward.

Thanks,

Bob

PR #341, opened 2023-06-20
Add ibut:operate tests

matsl:
## What

First set of tests for ibut:operate

## Why 

This is new, this is tricky with many states, we need to make it stable, we need to test it! ;-)

## Note

Added some comments where I see unexpected results.

When executed interactively/manually there are additional problems with interaction what it looks like `action-key-depress-prev-point` that seems to have some strange value. By using the action key in the tests the problem went away. When running it with a fresh Emacs like using make it does not seems to matter so I have removed those hacks. (But those could be important for practical use.)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,15 @@
+2023-06-27  Mats Lidell  <matsl@gnu.org>
+
+* test/hbut-tests.el (hbut-tests--ibut-operate--none)
+    (hbut-tests--ibut-operate--aname)
+    (hbut-tests--ibut-operate--aname-region-skip-region)
+    (hbut-tests--ibut-operate--region)
+    (hbut-tests--ibut-operate--modify-named)
+    (hbut-tests--ibut-operate--modify-named-skip-region)
+    (hbut-tests--ibut-operate--add-new-name)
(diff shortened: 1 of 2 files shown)


rswgnu (approved):

matsl:
Hi @rswgnu Thanks for approving. Just want to check if you really want to merge this since many of the tests are checking what I believe is the wrong behavior. Those discrepancies are documented in comments in the test cases. What do you think of going through each discrepancy and see if they can be fixed. When that is done on master I'll removed the comments here and update the tests to reflect the proper behavior.

rswgnu:
Yes, we can wait and I'll just work with these when I update ibut:operate
and then I can adjust to the proper behavior with the new code.  -- Bob

matsl:
@rswgnu Updated the tests with more checks. Try to verify that the ibut is still identified as an ibut after modification. Seems it can be down to inserting proper whitespace or standard delimiter after a name is inserted.  All tests succeeds but see the comments and verification on the buffer contents that documents the misbehavior.

matsl:
@rswgnu Marked as `WIP` while waiting for master branch to catch up.

rswgnu:
I have updated and merged needed ibut:operate tests into master, so this PR should be abandoned.

PR #343, opened 2023-06-21
Stefan scratch hyperbole 20230621

matsl:
## What

Stefans latest updates merged with Makefile from #342

## Why

Put up Stefans changes again for review. 

Makefile changes to make LOAD_EL from Stefan seems to be possible to solve better as in #342.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Author:       Bob Weiner
 #
 # Orig-Date:    15-Jun-94 at 03:42:38
-# Last-Mod:     24-Jun-23 at 00:30:37 by Mats Lidell
+# Last-Mod:     24-Jun-23 at 00:31:35 by Mats Lidell
 #
 # Copyright (C) 1994-2023  Free Software Foundation, Inc.
 # See the file HY-COPY for license information.
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Author:       Bob Weiner
 #
 # Orig-Date:    15-Jun-94 at 03:42:38
-# Last-Mod:     22-Jun-23 at 19:23:14 by Mats Lidell
+# Last-Mod:     24-Jun-23 at 00:30:37 by Mats Lidell
 #
 # Copyright (C) 1994-2023  Free Software Foundation, Inc.
 # See the file HY-COPY for license information.
--- a/hactypes.el
+++ b/hactypes.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    23-Sep-91 at 20:34:36
-;; Last-Mod:     10-Jun-23 at 21:13:02 by Bob Weiner
+;; Last-Mod:     24-Jun-23 at 09:44:22 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -32,7 +32,7 @@
 ;;; Standard Hyperbole action types
 ;;; ************************************************************************
 
-(defact annot-bib (key) ;; FIXME: Clean up namespace use!
+(defact annot-bib (key)
   "Follow internal ref KEY within an annotated bibliography, delimiters=[]."
   (interactive "sReference key (no []): ")
   (let* ((key-regexp (concat "^[*]*[ \t]*\\[" (ebut:key-to-label key) "\\]"))
--- a/hactypes.el
+++ b/hactypes.el
@@ -32,14 +32,14 @@
 ;;; Standard Hyperbole action types
 ;;; ************************************************************************
 
-(defact annot-bib (key)
+(defact annot-bib (key) ;; FIXME: Clean up namespace use!
   "Follow internal ref KEY within an annotated bibliography, delimiters=[]."
   (interactive "sReference key (no []): ")
-  (let ((key-regexp (concat "^[*]*[ \t]*\\[" (ebut:key-to-label key) "\\]"))
-	citation)
-    (if (save-excursion
-	  (goto-char (point-max))
-	  (setq citation (re-search-backward key-regexp nil t)))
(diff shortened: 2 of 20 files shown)


hversion.el
```diff
@@ -65,20 +60,25 @@ Where a part in the term-type is delimited by a `-' or  an `_'."
...
 		      "emacs")
+		     ;; FIXME: Do we still care about NeXT?
 		     ((or (featurep 'eterm-fns)
 			  (equal (getenv "TERM") "NeXT")
 			  (equal (getenv "TERM") "eterm"))
 		      ;; NEXTSTEP add-on support to Emacs
 		      "next"))))
-    (set-frame-parameter frame 'hyperb:window-system
-			 (and term (setq term (substring term 0 (string-match "[-_]" term)))))
+    (set-frame-parameter frame 'hyperb:window-system term)
```
rswgnu, line 69: My recollection is that we wanted just the first part of this term value for simpler comparisons but I don't have strong opinions on this.

hversion.el
```diff
@@ -65,20 +60,25 @@ Where a part in the term-type is delimited by a `-' or  an `_'."
...
+  ;; FIXME: This apparently can return only "emacs", "next", or nil.
+  ;; What do these things mean?  What does "window system available" mean?
+  ;; What does "mouse available mean"?
   "Return name of window system or term type where the selected FRAME is running.
 If nil after system initialization, no window system or mouse
 support is available."
   (unless frame (setq frame (selected-frame)))
+  ;; FIXME: Why not compute it on the fly rather than precomputing it
+  ;; via a hook and then saving it as a frame property?
   (frame-parameter frame 'hyperb:window-system))
```
rswgnu, line 82: Probably, just seemed simpler than going through a bunch of conditionals each time it was used.

hversion.el
```diff
@@ -122,7 +122,7 @@ support is available."
 	     ;; Force execution of Info-mode-hook which adds the
 	     ;; Hyperbole man directory to Info-directory-list.
 	     (info)
-	     (if (string-match "^(\\([^\)]+\\))\\(.*\\)" index-item)
+	     (if (string-match "^(\\([^)]+\\))\\(.*\\)" index-item)
 		 (let ((file (match-string-no-properties 1 index-item))
 		       (item-name (match-string-no-properties 2 index-item)))
 		   (if (and file (setq file (hpath:substitute-value file)))
```
rswgnu, line 128: Rest is fine.

Makefile
```diff
@@ -306,7 +306,7 @@ bin: src
 
 # Create -l "file.el" load-file command-line args for each Hyperbole .el file for use in
 # eln native compile target below.
-LOAD_EL = $(shell echo "$(EL_KOTL) $(EL_COMPILE)" | sed - -e 's+ +" -l "./+g' -e 's+^+-l "+')"
+LOAD_EL = $(EL_KOTL:%.el=-l ./%.el) $(EL_COMPILE:%.el=-l ./%.el)
```
rswgnu, line 309: Will this work with non-GNU make?  We want that support.

rswgnu (changes requested):

hactypes.el
```diff
@@ -716,7 +717,7 @@ Optional SECTIONS-START limits toc entries to those after that point."
       (insert "Sections of " rfc-buf-name ":\n")
       (set-buffer-modified-p nil))
     (when opoint
-      (select-buffer buf-name)
+      (set-buffer buf-name)
```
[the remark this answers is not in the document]
matsl: @rswgnu This is odd. `select-buffer` does not exist in any version of Emacs. So the old code is broken. What do you think of keeping this change and work on fixing the rfc-toc function in separate PRs?

hversion.el
monnier: It's definitely no simpler since recomputing it on-the-fly means extracting the computation into its own function and throwing away everything else :-)

hyperbole.el
```diff
@@ -562,6 +563,19 @@ frame, those functions by default still return the prior frame."
...
+(let ((us (if (fboundp 'macroexp-file-name)
+              (macroexp-file-name) load-file-name)))
+  (when us
+    ;; FIXME: Contrary to the usual ELPA autoloads files, `kotl-autoloads'
+    ;; does not add its directory to `load-path', so let's do it here by
+    ;; hand.
+    (add-to-list 'load-path
+                 (expand-file-name "kotl" (file-name-directory us)))
+    (require 'kotl-autoloads nil t)))
+
```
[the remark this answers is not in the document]
monnier: The main purpose of this code is not to be executed when `hyperbole.el` is loaded but when `hyperbole-autoloads.el` is loaded, i.e. at Emacs startup (thanks to the autoload cookie).

hbut.el
```diff
@@ -2690,8 +2700,6 @@ type for ibtype is presently undefined."
 	      (htype:create ,type ibtypes ,doc nil ,at-func
 			    '(to-p ,to-func style ,style))))))
 
-(put      'defib 'lisp-indent-function 'defun)
-
```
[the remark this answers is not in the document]
monnier: The `(indent defun)` in the `declare` statement in the definition of `defib`.

hyrolo.el
```diff
@@ -71,11 +71,10 @@
   :group 'hyperbole)
 
 (defcustom hyrolo-date-format "%m/%d/%Y"
-  "*Format of date string used in Rolo automatic date stamps.
+  "Format of date string used in Rolo automatic date stamps.
 Default is American style.  See documentation of the function
 `format-time-string' for format options."
-  :type 'string
-  :group 'hyperbole-rolo)
+  :type 'string)
```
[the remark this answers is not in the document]
monnier: They're redundant because they specify the group defined earlier in the file, which is used by default in the absence of a `:group`.

PR #353, opened 2023-06-29
Remove redundant :group args and obsolete * in defcustom docstrings

matsl:
## What

`:group` args are redundant and `*` is obsolete in `defcustom` docstrings.

Contribution from Stefan Monnier. Thanks Stefan.

See: https://github.com/rswgnu/hyperbole/pull/343/files#r1244628372

cc: @monnier

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,6 +17,9 @@
 
 2023-06-29  Mats Lidell  <matsl@gnu.org>
 
+* hyrolo.e: Remove redundant `:group` args.  Remove obsolete `*` in
+    `defcustom` docstrings. Contribution from Stefan Monnier. Thanks Stefan.
+
 * hbut.el (defib): Remove redundant indent property. Change by Stefan
     Monnier.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,7 +17,7 @@
 
 2023-06-29  Mats Lidell  <matsl@gnu.org>
 
-* hyrolo.e: Remove redundant `:group` args.  Remove obsolete `*` in
+* hyrolo.el: Remove redundant `:group` args.  Remove obsolete `*` in
     `defcustom` docstrings. Contribution from Stefan Monnier. Thanks Stefan.
 
 * hbut.el (defib): Remove redundant indent property. Change by Stefan
--- a/hyrolo.el
+++ b/hyrolo.el
@@ -71,11 +71,10 @@
   :group 'hyperbole)
 
 (defcustom hyrolo-date-format "%m/%d/%Y"
-  "*Format of date string used in Rolo automatic date stamps.
+  "Format of date string used in Rolo automatic date stamps.
 Default is American style.  See documentation of the function
 `format-time-string' for format options."
-  :type 'string
-  :group 'hyperbole-rolo)
+  :type 'string)
 
 (defvar hyrolo-display-format-function
   (lambda (entry)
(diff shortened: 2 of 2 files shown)


hyrolo.el
```diff
@@ -71,11 +71,10 @@
   :group 'hyperbole)
 
 (defcustom hyrolo-date-format "%m/%d/%Y"
-  "*Format of date string used in Rolo automatic date stamps.
+  "Format of date string used in Rolo automatic date stamps.
 Default is American style.  See documentation of the function
 `format-time-string' for format options."
-  :type 'string
-  :group 'hyperbole-rolo)
```
rswgnu, line 78: I want to group all hyperbole customizations together under 'hyperbole-' groups.  It seems to me if you remove this, the group for customizations here would be 'hyrolo', so it would not match this standard.  You could test it out but I think Stefan is just thinking that the grouping should match the filename rather than some broader linkage of customizations.

rswgnu (changes requested):

hyrolo.el
matsl: I think the defgroup `hyperbole-rolo` creates the grouping that you want. hyperbole-rolo is a group that is a subgroup of hyperbole. hyperbole-rolo is then default for all the defcustom.

This is how the custom settings look after the change. (Byte compiled the branch and restarted Emacs. That would be enough to see the new settings I think.)

bild

bild

rswgnu:
Ok, great.-- BobOn Jul 1, 2023, at 5:36 PM, Mats Lidell ***@***.***> wrote:﻿



In hyrolo.el:
>  Default is American style.  See documentation of the function
 `format-time-string' for format options."
-  :type 'string
-  :group 'hyperbole-rolo)

I think the defgroup hyperbole-rolo creates the grouping that you want. hyperbole-rolo is a group that is a subgroup of hyperbole. hyperbole-rolo is then default for all the defcustom.
This is how the custom settings look after the change. (Byte compiled the branch and restarted Emacs. That would be enough to see the new settings I think.)

ChangeLog
```diff
@@ -5,6 +5,9 @@
 
 2023-06-29  Mats Lidell  <matsl@gnu.org>
 
+* hyrolo.e: Remove redundant `:group` args.  Remove obsolete `*` in
```
rswgnu, line 8: Missing 'l' on .el.

hyrolo.el
```diff
@@ -71,11 +71,10 @@
   :group 'hyperbole)
 
 (defcustom hyrolo-date-format "%m/%d/%Y"
-  "*Format of date string used in Rolo automatic date stamps.
+  "Format of date string used in Rolo automatic date stamps.
 Default is American style.  See documentation of the function
 `format-time-string' for format options."
-  :type 'string
```
rswgnu, line 77: Missing the add back of the :type line here?

rswgnu (approved):
Approved with changes requested.

hyrolo.el
matsl: I don't understand the comment. The `:type` line kept!?

rswgnu:
All the other places, the :type line was left in (or added back in the
diff).  Here it just seems to be deleted.

matsl:
OK. I have merged the PR now. I'll make a note of that and take a look later.

matsl:
Later is now. I went to the master branch and it looks like it is still there. https://github.com/rswgnu/hyperbole/blob/ac6e50e3ecf98017999d43808ac8cab50d84278f/hyrolo.el#L77

PR #354, opened 2023-06-30
Remove some byte compilation warnings

matsl:
## What

Remove some byte compilation warnings

## Details

* hui.el:
  hpath.el:
  hbut.el:
  hbdata.el: Fix single quotes in docstrings.

* kotl/kview.el (kcell-view:to-label-end): Add missing point to error
    message.

* hactypes.el (require): Load hsys-org for org-fold-show-context.
    (link-to-org-id-marker): Use org-fold-show-context.
    (link-to-org-id): Silence byte compile warnings when org-roam not loaded.

### DIFF
--- a/hactypes.el
+++ b/hactypes.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    23-Sep-91 at 20:34:36
-;; Last-Mod:     25-Jun-23 at 13:48:01 by Bob Weiner
+;; Last-Mod:     30-Jun-23 at 22:44:36 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -18,7 +18,7 @@
 ;;; Other required Elisp libraries
 ;;; ************************************************************************
 
-(eval-and-compile (mapc #'require '(bookmark hvar hsettings comint hbut hpath hargs hmail man)))
+(eval-and-compile (mapc #'require '(bookmark hvar hsettings comint hbut hpath hargs hmail man hsys-org)))
 
 ;;; ************************************************************************
 ;;; Public declarations
@@ -503,7 +503,7 @@ available.  Filename may be given without the .info suffix."
     (hypb:error "(link-to-Info-node): Invalid Info node: `%s'" string)))
 
 (defact link-to-ibut (name-key &optional but-src point)
-  "Perform implicit button action specified by NAME-KEY, optional BUT-SRC and POINT.
+  "Do implicit button action specified by NAME-KEY, optional BUT-SRC and POINT.
 NAME-KEY must be a normalized key for an ibut <[name]>.
 BUT-SRC defaults to the current buffer's file or if there is no
 attached file, then to its buffer name.  POINT defaults to the
@@ -600,7 +600,9 @@ information on how to specify a mail reader to use."
   (when (stringp id)
     (let (m
 	  (inhibit-message t)) ;; Inhibit org-id-find status msgs
-      (when (setq m (or (and (featurep 'org-roam) (org-roam-id-find id 'marker))
+      (when (setq m (or (and (featurep 'org-roam)
+                             (with-no-warnings
+                               (org-roam-id-find id 'marker)))
 			(org-id-find id 'marker)))
 	(hact #'link-to-org-id-marker m)))))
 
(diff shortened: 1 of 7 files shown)


kotl/kview.el
```diff
@@ -507,7 +507,7 @@ If between kcells, move to the previous one.  The current cell may be hidden."
 	   ;; only in cases where no 'kcell property is found in cells,
 	   ;; e.g. before in-memory representation is created.
 	   (goto-char (- (match-end 0) 2)))
-	  (t (error "(kcell-view:to-label-end): In cell at pos %d, can't find end of cell's label")))))
+	  (t (error "(kcell-view:to-label-end): In cell at pos %d, can't find end of cell's label" (or pos (point)))))))
```
matsl, line 510: Copied `(or pos (point))` construct from line 497 but should maybe just be `(point)` here?

hactypes.el
```diff
@@ -503,7 +503,7 @@ available.  Filename may be given without the .info suffix."
     (hypb:error "(link-to-Info-node): Invalid Info node: `%s'" string)))
 
 (defact link-to-ibut (name-key &optional but-src point)
-  "Perform implicit button action specified by NAME-KEY, optional BUT-SRC and POINT.
+  "Do implicit button action specified by NAME-KEY, optional BUT-SRC and POINT.
```
rswgnu, line 506: 'Do' does not read as well.  We could use 'Execute' if you prefer that to 'Perform'.

hactypes.el
```diff
@@ -600,7 +600,9 @@ information on how to specify a mail reader to use."
   (when (stringp id)
     (let (m
 	  (inhibit-message t)) ;; Inhibit org-id-find status msgs
```
rswgnu, line 602: Shouldn't this inhibit-message binding inhibit warnings from the org-roam-id-find call?

hactypes.el
```diff
@@ -612,7 +614,7 @@ See doc of `ibtypes::org-id' for usage."
     (org-mark-ring-push)
     (hact #'link-to-buffer-tmp (marker-buffer marker) marker)
     (move-marker marker nil)
-    (org-show-context))
+    (org-fold-show-context))
```
rswgnu, line 617: This is not built-in to Emacs 27.1 so we should not update it.  For built-in packages, we always need to check this rather than listening to the package updates about what is obsolete; otherwise, we will break things for people who have not yet updated.

kotl/kview.el
rswgnu: (or pos (point)) - looks right.

rswgnu (changes requested):

hactypes.el
matsl: The change is for removing a byte compile warning. The line is too long and causes a 80-char warning. So Execute does unfortunately not resolve that. So we need to cut it down some how. Using Do was the best I could think of. Feel free to suggest any other way to make the line shorter.

hactypes.el
matsl: If org-roam is not loaded there is warning about org-roam-id-find not being available. I choose to use the no-warning here but I guess we could also declare org-roam-id-find as we have done for other external packages not necessary available.  Do you prefer that?

hactypes.el
rswgnu: Activate implicit button given by NAME-KEY, optional BUT-SRC and POINT.

hactypes.el
rswgnu: Yes, use the declaration instead.  Thanks.

rswgnu (changes requested):

hactypes.el
matsl: We have already taken this step by provide compatibility code in [hsys-org](https://github.com/rswgnu/hyperbole/blob/a19a1e1d278a654aa7a7d27fdcbdd4b5ff01adba/hsys-org.el#L86) hence the reason to require hsys-org to get this definition. Is it not best if we move to the new definition in one part of the code to do it everywhere?

Using the type of compatibility code that we have there looks to me as a good way to resolve the issues with obsolete functions.

rswgnu: Ok, sounds fine but I am running the latest Hyperbole and did not see a def for that function.  Let me check what is happening.

rswgnu: Ok, the hsys-org.el aliasing looks fine.

rswgnu (approved):

PR #355, opened 2023-06-30
Preload `kotl-autoloads`

matsl:
## What

* hyperbole.el: Preload `kotl-autoloads`

cc: @monnier

### DIFF
--- a/hyperbole.el
+++ b/hyperbole.el
@@ -7,7 +7,7 @@
 ;; Author:           Bob Weiner
 ;; Maintainer:       Bob Weiner <rsw@gnu.org>, Mats Lidell <matsl@gnu.org>
 ;; Created:          06-Oct-92 at 11:52:51
-;; Last-mod:     25-Jun-23 at 12:01:04 by Bob Weiner
+;; Last-Mod:      1-Jul-23 at 10:33:45 by Mats Lidell
 ;; Released:         03-Dec-22
 ;; Version:          8.0.1pre
 ;; Keywords:         comm, convenience, files, frames, hypermedia, languages, mail, matching, mouse, multimedia, outlines, tools, wp
@@ -566,9 +566,9 @@ frame, those functions by default still return the prior frame."
 (let ((us (if (fboundp 'macroexp-file-name)
               (macroexp-file-name) load-file-name)))
   (when us
-    ;; FIXME: Contrary to the usual ELPA autoloads files, `kotl-autoloads'
-    ;; does not add its directory to `load-path', so let's do it here by
-    ;; hand.
+    ;; Contrary to the usual ELPA autoloads files, `kotl-autoloads'
+    ;; does not add its directory to `load-path', so let's do it here
+    ;; by hand.
     (add-to-list 'load-path
                  (expand-file-name "kotl" (file-name-directory us)))
     (require 'kotl-autoloads nil t)))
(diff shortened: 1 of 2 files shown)


hyperbole.el
```diff
@@ -560,6 +560,19 @@ frame, those functions by default still return the prior frame."
 
 (makunbound 'hyperbole-loading)
 
+;; Autoload this form so that when `package.el' activates Hyperbole's autoloads
+;; it also sets up Kotl's autoloads.
+;;;###autoload
+(let ((us (if (fboundp 'macroexp-file-name)
+              (macroexp-file-name) load-file-name)))
+  (when us
+    ;; FIXME: Contrary to the usual ELPA autoloads files, `kotl-autoloads'
```
rswgnu, line 569: Remove the FIXME word, this is just a comment.  If there is some way to make this happen in the autoload file, we should but I don't know what is different there compared to the Hyperbole one.

rswgnu (changes requested):

hyperbole.el
```diff
@@ -7,7 +7,7 @@
 ;; Author:           Bob Weiner
 ;; Maintainer:       Bob Weiner <rsw@gnu.org>, Mats Lidell <matsl@gnu.org>
 ;; Created:          06-Oct-92 at 11:52:51
-;; Last-mod:     25-Jun-23 at 12:01:04 by Bob Weiner
+;; Last-mod:      1-Jul-23 at 10:12:06 by Mats Lidell
```
matsl, line 10: @rswgnu Noticed now that out header support does not handle the alignment properly and that one number dates does not get prefixed with a leading 0 as seem to be the standard from the past. Something that we should fix?

hyperbole.el
```diff
@@ -7,7 +7,7 @@
 ;; Author:           Bob Weiner
 ;; Maintainer:       Bob Weiner <rsw@gnu.org>, Mats Lidell <matsl@gnu.org>
 ;; Created:          06-Oct-92 at 11:52:51
-;; Last-mod:     25-Jun-23 at 12:01:04 by Bob Weiner
+;; Last-Mod:      1-Jul-23 at 10:33:45 by Mats Lidell
```
matsl, line 10: Updated field tag to `Last-Mod` with capital M, which is what we have in all other files, which led to that my comment about the date format got lost from the PR. So repeating it here.

@rswgnu Noticed that we for dates with one digit do not prefix with a leading zero that we have a for the Created date. So looks a bit odd. This is of course a minor thing. Not sure i we want to fix it and what would be the appropriate fix - add leading zero or remove leading zero from the Created field?

There is also an issue with the alignment that we might want to fix in the header update code. Also that a minor of course.

hyperbole.el
rswgnu: Padding with a space like you show here so the dates still align is fine.  Are the dates not aligning properly now?

hyperbole.el
rswgnu: I think for single digit dates, using a space to pad is best.  If Mod needs to be capitalized to work properly, then please fix it.  I am open to any code changes that fix bugs.

rswgnu (approved):

hyperbole.el
matsl: We have Last-Mod with capital M in all files except here in hyperbole.el.

The alignment I'm referring to is the alignment with the other fields. Seems we have the "args" part moved further to the right here in hyperbole.el and that is causing the issue.

PR #357, opened 2023-07-05
Do not include hyperbole-pgk.el

matsl:
## What

Do not include hyperbole-pgk.el

## Why

We do not have it in the workspace (so `make pkg` currently fails due
to this). It is generated by elpa and it is not needed for a plain tar
archive of the package. So we do not have to provide it, we can just
skip it.


rswgnu (commented):
- Let's discuss.  I would like to know when this was added as I think it was pretty long ago bt can't recall.  I remember getting Stefan to let us include it in what we sent him, maybe for other package managers that use it. Just don't want to break something else.

matsl:
We removed `hyperbole-pkg.el` from git in commit https://github.com/rswgnu/hyperbole/commit/df1d9401177169ea521961f62b69b084455c71c8. Since then `make pkg` has been broken but probably been under the radar since that will not be noticed until the folder is git cleaned or a fresh clone is made (and used for making a package.)

rswgnu (approved):

PR #358, opened 2023-07-07
Test anchor with white space and trailing text

matsl:
## What

Test anchor with white space and trailing text and various minor updates to the text cases (cleanups)

## Why

Pathnames with an anchor that contains whitespace and there are traling text did not work properly. This adds a test for this. (So is expected to fail until this is fixed on master.)


matsl:
@rswgnu This PR adds a test for the anchor link where the anchor contains whitespace and there is additional text after that. When master branch is fixed it should be possible to rebase/merge this PR and it should work (Hopefully! :smile:)

rswgnu (approved):

matsl:
@rswgnu Thanks for approving. Marked as `WIP` while waiting for master branch to catch up.

test/hmouse-drv-tests.el
```diff
@@ -271,43 +268,62 @@
...
+      (hy-delete-file-and-buffer file))))
+
+(ert-deftest hbut-pathname-anchor-trailing-text ()
+  "Pathname with anchor and trailing text."
+  (let ((file (make-temp-file "hypb" nil nil
+                              "Text\n* Anchor Plus More\n")))
   (unwind-protect
       (with-temp-buffer
-        (insert "\"${hyperb:dir}/DEMO#Smart Keys\"")
+        (insert (concat "\"" file "#Anchor Plus\""))
```
matsl, line 293: @rswgnu 

This is still failing. If the text, none whitespace, after the `anchor` is started with a parentheses. Like this:
```
* Anchor Plus (text after anchor)
```
Then it works as expected and the anchor `Anchor Plus` is found.

Is that what we want to support or should the test case as it is written work? i.e. any text can follow the anchor like this:
```
* Anchor Plus text after anchor
```

rswgnu:
The former for now.  We’re trying to do an exact match sans any parenthesized-type text.Otherwise, you don’t know which heading is meant when the anchor is #MatchAnd headings are:* Match this* Match that-- BobOn Jul 12, 2023, at 3:11 AM, Mats Lidell ***@***.***> wrote:﻿
@matsl requested your review on: #358 Test anchor with white space and trailing text.

matsl:
OK. I'll adjust the test cases to reflect the desired behavior. Thanks for clarifying.

PR #359, opened 2023-07-07
Add koutline tests

matsl:
## What 

Add koutline tests.

### DIFF
--- a/test/kotl-mode-tests.el
+++ b/test/kotl-mode-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    18-May-21 at 22:14:10
-;; Last-Mod:      9-Oct-23 at 00:50:49 by Mats Lidell
+;; Last-Mod:      9-Oct-23 at 00:51:28 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


test/kotl-mode-tests.el
```diff
@@ -714,15 +739,29 @@ In non interactive mode there shall be no beep (nor message)"
...
+            (funcall-interactively #'kotl-mode:previous-line 1))
           ;; Non interactive does not call neither message nor beep so the mock fails.
           (should-error
            (with-mock
              (mock (beep) => t)
              (kotl-mode:previous-line 1))))
       (hy-delete-file-and-buffer kotl-file))))
 
-;; TODO: Add test for moving up to first line with no error.
+(ert-deftest kotl-mode-move-up-to-first-line ()
```
matsl, line 750: @rswgnu This test fails due to error is signaled when moving up to first line. I think this is a bug. wdyt?
The test is run in interactive mode so does not affect the CI build. So I added it to not forget it.

rswgnu: I just tested trying to move to the previous line from the first line in kotl-mode and in another mode.  Both basically behaved the same.  With debug-on-error t, both beep and display a message but do not show a backtrace because they are not actually triggering an error (just look at the code for kotl-mode:previous-line).  Try it with a test yourself.  So if there is an issue it seems it would be with the test.

rswgnu (changes requested):

test/kotl-mode-tests.el
matsl: @rswgnu The test is moving up to the first line from the `second` line. There is another test case for trying to move to the previous line form the first line.

rswgnu: Ok, I understand now.  Will have a look.

rswgnu: You changed hte name to 'kotl-mode-move-up-from-first-line' and you don't insert a newline in this test so point ends on the first line and should indeed beep when you try to move up.  Maybe you have another test that moves from the second line.

rswgnu:
Let's figure out how to close this PR out.  Thanks.

rswgnu:
> Let's figure out how to close this PR out. Thanks.

Hi Mats:  could you resolve the final with-mock issue here or test again?

matsl:
Hi  @rswgnu, Yes. Thanks for the reminder. I will take a look.

matsl:
Hi @rswgnu, I reworked the PR a bit to make the tests clearer. Not sure what was the issue before. Maybe the should-error construct  for mocks that ate **not** called is a bit convoluted but it is one way to verify that functions are not being called. Maybe `flet` is a better machinery for that but not sure since that would require more code.

rswgnu (approved):

PR #365, opened 2023-07-10
remove warnings 2023-07-10 default-dir usage

matsl:
## What

Use hypb--klink-default-dir (Based on earlier PR for convenience) See commit 5b3b674ad02c51d5fcacb77433d5c9beaa2cc962 for the actual code that is of interest.

## Why

I want to understand the `default-dir` usage and how we can implement
that in another way. Adding this PR for the discussion.

### DIFF
--- a/hargs.el
+++ b/hargs.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    31-Oct-91 at 23:17:35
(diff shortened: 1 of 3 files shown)


rswgnu:
Not really sure about the default-dir changes as you say, so just approved
the prior non-'b' PR and should discuss this one as you said.  -- Bob

rswgnu:
I think I solved this with my prior changes that now pass the default dir as an attribute of the current button.  Check it out but I think we can abandon this change.

PR #366, opened 2023-07-14
Add tests for pathname with line number specification

matsl:
## What

Add tests for pathname with line number specification.

## Why

There is a bug with duplicate ibuts to the same file so we need to
test that. This the test `hbut-pathname-line-test-duplicate` pinpoints
the problem. It is setup to succeed with a FIXME to highlight the test
that is wrong.


rswgnu (commented):
Approved.

PR #383, opened 2023-09-26
Let git see el-files

matsl:
## What

Do not ignore el-files.

## Why

 * Emacs lisp files are the core of the package so most likely what we want git to track. 
 * Tools that respect gitignore also fails to look at Emacs lips files. An example of that is `ripgrep` which makes the setting extra tedious to work with since all searches fail to find things in Emacs lisp files unless explicitly told to look there.   
 * (Possibly and oversight or error that this change was committed!?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,3 @@ TODO*
 
 # Video Demos
 videos
-*.el


matsl:
@rswgnu If you have emacs lisp files in the hyperbole folder that you want git to ignore you can have a [user local gitignore for those](https://stackoverflow.com/a/5724499). That way it will only affect your environment.

rswgnu:
You can use the hypb:rgrep command (which can be configured to use ripgrep)
to search over all the Hyperbole .el files.  If invoked within a .el
buffer, it will automatically filter to search just .el files.

Do you really need anything else.  Stop using grep within a shell for
Hyperbole and you will find your life will get a bit easier.

Bob

matsl:
> Do you really need anything else

:smile: 

@rswgnu I like how your are thinking... BUT It is not only about me. Everyone else that uses the repo will get the, IMHO, strange behavior that el-files are ignored in an Emacs lisp project and it will affect all other tools that respect the .gitignore configuration as a hint that these are uninteresting files. That will be both Emacs tools and external tools. 

Please reconsider to make this into your personal setting instead to not affect general users. That can either be done by using a local setting for the Hyperbole repo, as shown in the link, or in your general .gitiignore in your home folder.

rswgnu:
Ok, will work on it as well as temporarily reverting using org-mode in the demos, so the tests will pass and we can merge to master.-- BobOn Oct 2, 2023, at 8:54 AM, Mats Lidell ***@***.***> wrote:﻿

Do you really need anything else

😄
@rswgnu I like how your are thinking... BUT It is not only about me. Everyone else that uses the repo will get the, IMHO, strange behavior that el-files are ignored in an Emacs lisp project and it will affect all other tools that respect the .gitignore configuration as a hint that these are uninteresting files. That will be both Emacs tools and external tools.
Please reconsider to make this into your personal setting instead to not affect general users. That can either be done by using a local setting for the Hyperbole repo, as shown in the link, or in your general .gitiignore in your home folder.

rswgnu (approved):

rswgnu:
I have done all this and merged into master.  You can now merge your
.gitignore PR and then fork off master to work on cleaning up your PR that
removes all the byte-compile errors.
Thanks.

PR #384, opened 2023-09-29
First stab at compiling files separately - and remove warnings

matsl:
## What

First stab at compiling files separately - and remove warnings that are shown then

## Why

We want to remove warnings.

## Note

I flipped the coin and used the compiler to tell me what functions and variable that needs to be declared to silence the compiler. The added declarations are made to stand out by the header `;; FIXME: Circular dependencies`. (Not all declarations are due to circular dependencies but I started out that way. ) This is done with the hope that is will be easier to see what changes we should promote to master when we see in clear text what is missing.

One idea is to let these changes in, after removing the boiler plat comments, with the regular declarations and work over time to reduce the number of functions and vars that are declared in favor of using require. Not sure about the best practice here. I thought the proper way was to use `require` and work on the dependencies so that the amount of `declarations` are minimal. But if you can load the package and things work then maybe it does not matter and using declarations is a means to cut dependency circles!? 

The `kview` global variable had to be fixed as well. I renamed it `hypb-kview`. That change is committed in a separate commit for easier review.

It does not show here but if this is compiled with the new `new-bin` target, that compiles each file in isolation, there are nearly no warnings left.  For what I know this is equivalent with how the package manager compiles the files.

Worth to note is that I have problems with the mouse actions on the master branch so I can't rule out that this PR introduces problems. All test pass though which is encouraging.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,3 @@ TODO*
 
 # Video Demos
 videos
-*.el
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Author:       Bob Weiner
 #
 # Orig-Date:    15-Jun-94 at 03:42:38
-# Last-Mod:     28-Sep-23 at 23:57:18 by Mats Lidell
+# Last-Mod:     29-Sep-23 at 16:43:34 by Mats Lidell
 #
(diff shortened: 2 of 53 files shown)


rswgnu:
This PR should be exclusively for eliminating compilation errors.  Put any other changes into a separate branch.

PR #387, opened 2023-10-02
Compile elc files in separation

matsl:
## What

Compile elc files in separation

## Why

This is a preparation for the work with removing warnings. By doing the Makefile changes first we can 
 * check that this step works as expected 
 * we can see all the warnings that are produced and 
 * compare that with what we get using a package manager. 

## Note

I tried to silence all the "Loading" messages that are printed but gave up on it. Not sure it is a good thing either. So keeping it for now.


matsl:
Was merged as part of #388.

PR #388, opened 2023-10-03
Rework of compile files separately before kview update

matsl:
## What 

Rework of compile files separately before kview update. Most updates are adding `defvar` and `declare-functions` but there a few other fixes as well using `require`. No further fixes for warnings introduced in master later than the original work branch for this has been done. So there are a few warnings left. These will be fixed in a later separate PR.
 
## Why

Re-bases work on latest master and starts from PR #387 so that files are compiled separately and we can see the difference in number of warning.

## Note

I have also tried to tidy up the code, removing most comments about the circular dependencies etc. There are some stray comments left. I will deal with those later. Submitting this for review (and it is getting late so will stop work for today here.) 

The kview global variable changes are not included as we said that will also be done in a separate PR.


hyrolo.el
```diff
@@ -1110,7 +1168,7 @@ Return number of matching entries found."
       (insert "No result.")
     (print contacts (get-buffer-create "*contacts-data*"))
     (dolist (contact contacts)
-      (let* ((child)
+      (let* ((child nil)
```
matsl, line 1171: Although let for a variable should according to the docs bind it to `nil` we get the warning " Variable ‘child’ left uninitialized" if we don't explicitly set it to nil.

rswgnu: Some if the compilers get very pedantic with their warnings.  This is perfectly good Lisp style but I guess they are trying to keep newbies from hurting themselves now.  C’est la vie.

rswgnu (approved):
Hi Mats: This is so big, I’ll just assume you got everything working and we’ll have to test a lot after the merge to ensure things work as before.  Great work getting through all this.

matsl:
Thanks for approving. Is it nearly 100% just defvars and declarations that should not have any impact on functionality. I appreciate the pragmatic view and will merge this and then we can filter out any problems.

PR #389, opened 2023-10-04
Add kotl prefix to kview buffer local variable

matsl:
## What

Add kotl prefix to kview buffer local variable.

## Why

Removes warnings for kview lacking a prefix. Makes code easier to read since parameters exists with the same name. Warnings for that are also avoided by this change.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2023-10-05  Mats Lidell  <matsl@gnu.org>
+
+* test/kotl-mode-tests.el (kotl-mode-kview-buffer-local)
+    (kotl-mode-kvspec-saved-with-file)
+    (kotl-mode-kvspec-independent-between-files): Add test for buffer
+    local kotl-kview.
+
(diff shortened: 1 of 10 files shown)


ChangeLog
```diff
@@ -1,3 +1,11 @@
+2023-10-04  Mats Lidell  <matsl@gnu.org>
+
+* kotl/kview.el (kotl-kview): Add kotl prefix to kview and use
+    defvar-local.
+
+* General removal of warnings by using defvar and declare-function for
```
matsl, line 6: Added this to cover for the lack of ChangeLog comments in the previous PR.

rswgnu (approved):

rswgnu:
Approved.  Give it a good run through, changing the viewspec of two different koutlines side by side to ensure the view local changes work, then save, kill and reload and see if viewspec restores separately for both.Cheers,-- BobOn Oct 4, 2023, at 6:57 PM, Mats Lidell ***@***.***> wrote:﻿
@matsl requested your review on: #389 Add kotl prefix to kview buffer local variable.

PR #390, opened 2023-10-05
Remove compiler warnings introduced lately

matsl:
## What

Remove compiler warnings introduced lately. These are the last easy warning to remove.

## Why

We want hyperbole to compile with not warnings.

### DIFF
--- a/hui-mouse.el
+++ b/hui-mouse.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    04-Feb-89
-;; Last-Mod:      3-Oct-23 at 22:56:59 by Mats Lidell
+;; Last-Mod:      6-Oct-23 at 16:13:42 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -48,6 +48,8 @@
 (unless (fboundp 'smart-c-at-tag-p)
   (require 'hmouse-tag))
 (require 'imenu)
+(eval-when-compile
+  (require 'eieio))
 
 (eval-when-compile (require 'tar-mode))
 
@@ -1531,6 +1533,11 @@ If assist-key is pressed:
 	   (if non-text-area-p
 	       (magit-section-cycle-global)
 	     (smart-magit-display-file (key-binding (kbd "RET"))))))))
+
+;; Silence compiler about unknown slot. From eieio-tests.el
+(eval-when-compile
+  (dolist (slot '(content washer hidden start))
+    (cl-pushnew slot eieio--known-slot-names)))
 
 ;; Thanks to Jonas Bernoulli <tarsius>, magit author, for most of this
 ;; next function.
--- a/hui.el
+++ b/hui.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    19-Sep-91 at 21:42:03
-;; Last-Mod:      3-Oct-23 at 23:02:02 by Mats Lidell
+;; Last-Mod:      4-Oct-23 at 20:07:50 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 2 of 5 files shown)


hui.el
```diff
@@ -1291,7 +1291,7 @@ drag from a window to another window's modeline."
       ;; single C-u prefix argument.  In such a case, don't use the
       ;; prefix argument as a flag to prompt for the ibutton name as
       ;; we want to just insert the appropriate ibut without any prompting.
-      (when (and name-arg-flag (not (eq name-arg-flag '(4))) (not name-key))
+      (when (and name-arg-flag (not (equal name-arg-flag '(4))) (not name-key))
```
matsl, line 1294: This is the warning about the potential problem with eq not giving the expected behavior since atoms might not be the same although their value is the same.

Odd thing is:
```
(eq '(4) '(4)) => nil
```
Which makes me wonder how this code is supposed to work. We also have many test cases which uses eq with constants that should not be trusted but seems to work!? Maybe you know.

Anyway I think using equal is a small price for not getting the warning.

rswgnu: You are right, good catch.  Two instances of the same atom like nimbers ir symbols will compare eq but two lists of the same number will not, so this was just wrong and I imagine never triggering.

hui-mouse.el
```diff
@@ -48,6 +48,7 @@
 (unless (fboundp 'smart-c-at-tag-p)
   (require 'hmouse-tag))
 (require 'imenu)
+(require 'eieio)
```
rswgnu, line 51: We should not be using this library.  Let’s eliminate any dependency on it.  How could the code have been running if this were necessary?

hui-mouse.el
```diff
@@ -1532,6 +1533,11 @@ If assist-key is pressed:
 	       (magit-section-cycle-global)
 	     (smart-magit-display-file (key-binding (kbd "RET"))))))))
 
+;; Silence compiler about unknown slot. From eieio-tests.el
+(eval-when-compile
+  (dolist (slot '(content washer hidden start))
+    (cl-pushnew slot eieio--known-slot-names)))
+
```
rswgnu, line 1540: Sorry, don’t want this in.  Not a fan of the complexity therein.

rswgnu (changes requested):

hui-mouse.el
matsl: It comes from `magit` that uses `eieio` `slots` that we touch. So far I have not found any other way to remove the warnings for the slots. So it is either this or getting the warnings for the slots. [Check earlier build.](https://github.com/rswgnu/hyperbole/actions/runs/6423671655) to see how it look.

So we are not using anything additional in there I think. It is just being able to add the slots we use to the known slot names.

Now you know. I agree it looks a bit dirty but the warnings do stick out too. With few warnings the ones that are left get more space! ;-)

All in all I have no big issue with removing this. Just wanted to make it clear the reason for why it is needed.

rswgnu:
Yes, I think just wrap an eval-when-compile around the require and we should be good.-- BobOn Oct 5, 2023, at 10:57 PM, Mats Lidell ***@***.***> wrote:﻿



In hui-mouse.el:
> @@ -1532,6 +1533,11 @@ If assist-key is pressed:
 	       (magit-section-cycle-global)
 	     (smart-magit-display-file (key-binding (kbd "RET"))))))))
 
+;; Silence compiler about unknown slot. From eieio-tests.el
+(eval-when-compile
+  (dolist (slot '(content washer hidden start))
+    (cl-pushnew slot eieio--known-slot-names)))
+

It comes from magit that uses eieio slots that we touch. So far I have not found any other way to remove the warnings for the slots. So it is either this or getting the warnings for the slots. Check earlier build. to see how it look.
So we are not using anything additional in there I think. It is just being able to add the slots we use to the known slot names.
Now you know. I agree it looks a bit dirty but the warnings do stick out too. With few warnings the ones that are left get more space! ;-)
All in all I have no big issue with removing this. Just wanted to make it clear the reason for why it is needed.

rswgnu (approved):

PR #401, opened 2023-10-30
Add mac as possible window-system on Mac OS

matsl:
## What

Add mac as possible window-system on Mac OS

## Why

https://debbugs.gnu.org/cgi/bugreport.cgi?bug=66838

### DIFF
--- a/hversion.el
+++ b/hversion.el
@@ -4,7 +4,7 @@
 ;; Maintainer:   Bob Weiner, Mats Lidell
 ;;
 ;; Orig-Date:     1-Jan-94
-;; Last-Mod:      3-Oct-23 at 23:31:30 by Mats Lidell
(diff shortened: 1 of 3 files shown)


hversion.el
```diff
@@ -39,7 +39,7 @@
 
 (defvar hyperb:mouse-buttons
   (if (or (and hyperb:microsoft-os-p (not (memq window-system '(w32 w64 x))))
-	  (memq window-system '(ns dps)))
+	  (memq window-system '(mac ns dps)))
```
matsl, line 42: Do we need this as well for two button mice on Mac's? Is that the current default?

rswgnu: NextStep used 2-button mice, some people still have them.  Mac can be one or more buttons.

rswgnu (approved):

PR #408, opened 2023-11-24
Add hyrolo search test cases

matsl:
## What

Add hyrolo search test cases.

## Note

1. Pushing the first test case that verifies searching all file types with matches gives @loc> headers. Will add more.
2. Added test cases for matching a single record in respective file type and then using action-key to go to record.

### DIFF
--- a/test/hyrolo-tests.el
+++ b/test/hyrolo-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    19-Jun-21 at 22:42:00
-;; Last-Mod:     17-Nov-23 at 10:48:26 by Bob Weiner
+;; Last-Mod:     25-Nov-23 at 00:16:26 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


rswgnu:
While you have the different types of matched entries in a buffer, you should use the action key on each, e.g. by moving to each string match first and then ensure the proper record from each original file is displayed.Might be good to have more than one record with some hierarchy matched from each file type so you see if the tree-level matching works.  The Koutline needs the most testing in this area.Cheers,BobOn Nov 24, 2023, at 11:25 PM, Mats Lidell ***@***.***> wrote:﻿What
Add hyrolo search test cases.
Note
Pushing the first test case that verifies searching all file types with matches gives @loc> headers. Will add more.



You can view, comment on, or merge this pull request online at:
  https://github.com/rswgnu/hyperbole/pull/408

Commit Summary

  66d87cf  Verify searching all file types with matches gives @loc> headers


File Changes  (2 files)

  
    M
    ChangeLog
    (5)
  
  
    M
    test/hyrolo-tests.el
    (21)
  


Patch Links:

  https://github.com/rswgnu/hyperbole/pull/408.patch
  https://github.com/rswgnu/hyperbole/pull/408.diff

matsl:
> While you have the different types of matched entries in a buffer, you should use the action key on each, e.g. by moving to each string match first and then ensure the proper record from each original file is displayed.Might be good to have more than one record with some hierarchy matched from each file type so you see if the tree-level matching works.  The Koutline needs the most testing in this area.Cheers

Yes. I will add more test cases around this area. Stay tuned.

test/hyrolo-tests.el
```diff
@@ -270,5 +271,117 @@ and {b} the previous same level cell."
...
+            (action-key)
+            (should (equal (current-buffer) (find-buffer-visiting outl-file)))
+            (should (looking-at-p "* heading"))))
+      (hy-delete-file-and-buffer outl-file)
+      (kill-buffer "*HyRolo*")
+      (delete-directory temporary-file-directory))))
+
+(ert-deftest hyrolo-fgrep-and-goto-next-visible-md-heading ()
+  "Verify move to next heading, then action-key to go to record for markdown mode."
+  :expected-result :failed
```
matsl, line 365: @rswgnu Tagging this as expected failed due to => The `hyrolo-next-visible-heading` step does not bring point to the markdown heading. Is not that what is supposed to happen?

rswgnu (approved):

PR #409, opened 2023-11-25
Remove warnings

matsl:
## What

Remove warnings

## Why

Recent updates have introduced a set of new warnings. This tries to remove the warnings by:
1. First of all using declare-function for declaring unknown functions. 
2. ~Fix call to derived-mode-p that only takes one arg, a list, and not multiple args.~ Reverted. This change is introduced in Emacs 30.1 so will not affect older versions and it is backwards compatible so we can ignore this for now.
3. Move private vars before their first use.
4. Remove a warning for a variable not used. (Not in a separate commit. See comment below.)

### DIFF
--- a/hyrolo.el
+++ b/hyrolo.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     7-Jun-89 at 22:08:29
-;; Last-Mod:     19-Nov-23 at 18:11:33 by Bob Weiner
+;; Last-Mod:     25-Nov-23 at 17:09:20 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -96,6 +96,13 @@
 (declare-function outline-previous-visible-heading "outline")
 (declare-function outline-show-all "outline")
 (declare-function outline-up-heading "outline")
+
+(declare-function hmouse-pulse-line "hui-window")
+(declare-function hpath:find "hpath")
+(declare-function hpath:expand-list "hpath")
+(declare-function hbut:get-key-src "hbut")
+(declare-function markdown-outline-level "ext:markdown-mode")
+(declare-function org-outline-level "org")
 
 (defvar org-directory)                  ; "org.el"
 (defvar markdown-regex-header)          ; "markdown-mode.el"
--- a/hyrolo.el
+++ b/hyrolo.el
@@ -137,35 +137,6 @@ executable must be found as well (for Oauth security)."
   :type 'boolean
   :group 'hyperbole-hyrolo)
 
-(defun hyrolo-google-contacts-p ()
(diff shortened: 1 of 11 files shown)


hyrolo.el
```diff
@@ -1595,11 +1684,10 @@ Return number of matching entries found."
 			(forward-line)
 			(setq hdr-pos (cons (point-min) (point))))
 		      (let ((case-fold-search t)
-			    match-start
```
matsl, line 1598: match-start is not used and can thus be removed.

rswgnu (changes requested):
derived-mode-p as of Emacs 28.2 uses &rest to create a single list from all args given, so sending it a list as an arg may break it and is not necessary.  Did the calling signature change at some point?

Remove that change and the rest looks good.

rswgnu (approved):

matsl:
Thanks for approving.

> derived-mode-p as of Emacs 28.2 uses &rest to create a single list from all args given, so sending it a list as an arg may break it and is not necessary. Did the calling signature change at some point?

It is changed in Emacs 30.1 to only take either one symbol or a list of symbols. So that hit me as a warning since I develop using the latest and greatest! 

But the test cases in CI made me aware that this change is not backwards compatible so I ha to revert that part of the PR, see commit https://github.com/rswgnu/hyperbole/pull/409/commits/1f4381d8e9ff4c8aaf2ed3ce207f5e64ea2cca8e. Hence point 2 is over struck although I understand it is not clear that the following comment is an explanation on why we keep the old code with using multiple symbols. The reason is that the change in 30.1 is backwards compatible so calling it with multiple symbols still work. I guess it will be removed at some point in the future. So for now we will have to live with the warning in Emacs 30.1. Maybe later it is time to deal with this.

rswgnu:
Yes, will be interesting to see why they made this change.-- BobOn Nov 26, 2023, at 3:20 PM, Mats Lidell ***@***.***> wrote:﻿
Thanks for approving.

derived-mode-p as of Emacs 28.2 uses &rest to create a single list from all args given, so sending it a list as an arg may break it and is not necessary. Did the calling signature change at some point?

It is changed in Emacs 30.1 to only take either one symbol or a list of symbols. So that hit me as a warning since I develop using the latest and greatest!
But the test cases in CI made me aware that this change is not backwards compatible so I ha to revert that part of the PR, see commit 1f4381d. Hence point 2 is over struck although I understand it is not clear that the following comment is an explanation on why we keep the old code with using multiple symbols. The reason is that the change in 30.1 is backwards compatible so calling it with multiple symbols still work. I guess it will be removed at some point in the future. So for now we will have to live with the warning in Emacs 30.1. Maybe later it is time to deal with this.

matsl:
> Yes, will be interesting to see why they made this change.

In the bug report there is some rational. https://lists.gnu.org/archive/html/bug-gnu-emacs/2023-11/msg00497.html

PR #410, opened 2023-11-30
Add tests for ebut-link-directly and gbut-link-directly

matsl:
## What

Add tests for ebut-link-directly and gbut-link-directly.

## Note

Copied from similar ibut test cases.

### DIFF
--- a/test/hui-tests.el
+++ b/test/hui-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    30-Jan-21 at 12:00:00
-;; Last-Mod:     15-Nov-23 at 01:57:15 by Bob Weiner
+;; Last-Mod:     30-Nov-23 at 18:39:54 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


test/hui-tests.el
```diff
@@ -831,6 +831,57 @@ With point on label suggest that ibut for rename."
...
+	  (goto-char (line-end-position))
+	  (skip-chars-backward "^:")
+	  (when (/= (point) (point-min))
+	    (goto-char (1- (point))))
+          (split-window)
+          (find-file file)
+          (with-simulated-input "button RET"
+            (hui:ebut-link-directly (get-buffer-window) (get-buffer-window dir-buf))
+	    ;; Implicit link should be the `dir' dired directory,
+	    ;; possibly minus the final directory '/'.
```
matsl, line 877: This is copied from the ibut test case. @rswgnu Why the possibly? Apparently a directory can be without a trailing slash in the ebut args, but why? 

I transformed it in this test case to a file name so the test will pass but that feels a bit hacky. Should we not make sure all link-to-directory ebuts have a trailing slash? 

Can this be related to that environment variables can be inserted in the ebut args and that it could make it hard to be consistent? Do we need to relax the requirement for a trailing slash here so that the helper function, `hy-test-helpers-verify-hattr-at-p`, should ignore the trailing slash difference?

rswgnu: I think so.  There are too many code paths where a dir without a final slash could come in.

rswgnu (approved):

PR #412, opened 2023-12-01
Declare function to silence warning

matsl:
# What

Remove warning for unknown function.

# Why

Keeping the number of warning as low as possible.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,7 @@
 2023-12-01  Mats Lidell  <matsl@gnu.org>
 
+* hmouse-tag.el (ibtype:def-symbol): Declare function to silence warning.
+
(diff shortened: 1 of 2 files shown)


rswgnu (approved):
Generally ahould declar in alphabetical irder by function name for easy reference.

matsl:
@rswgnu I agree I just tossed this in here. I'm conflicted between keeping them grouped around what external package they come from and generally ordered. 

All these declarations feels like a code smell to me. At least for the internal dependencies should we not strive towards just using requires? For the external dependencies I guess it is the only practical way to remove these warnings. In practice it has been a good tool for reducing even the internal warnings. All attempts with using require has failed due to cross or circular dependencies between files. We should not settle for that in the long run.

PR #414, opened 2023-12-06
Add test for hui:link-possible-types

matsl:
# What

Add test for `hui:link-possible-types.`

# Note

Continuing working with link-types. Looks like there are some issues here. Se comments below. 

All tests are run in one test case. Totally flipping what I said in my talk but in this case it might make sense since all tests are around the list of responses from the function under test. But I might split test tests into separate ert-deftests later to make it easier to detect regressions.

### DIFF
--- a/test/hui-tests.el
+++ b/test/hui-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    30-Jan-21 at 12:00:00
-;; Last-Mod:     30-Nov-23 at 19:53:31 by Mats Lidell
+;; Last-Mod:      7-Dec-23 at 00:32:03 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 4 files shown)


test/hui-tests.el
```diff
@@ -932,6 +932,114 @@ With point on label suggest that ibut for rename."
...
+          (hui:gibut-create "global" "/tmp")
+          (find-file global-but-file)
+          (should (equal (caar (hui:link-possible-types)) 'link-to-gbut)))
+      (hy-delete-file-and-buffer global-but-file)))
+
+  ;; Explicit Button          link-to-ebut
+  (with-temp-buffer
+    (ebut:program "label" 'link-to-directory "/tmp")
+    (should (equal (hattr:get (hbut:at-p) 'categ) 'explicit))
+    ;; BUG!? SHOULD BE LINK TO EBUT HERE
```
matsl, line 962: Should be `link-to-ebut`

test/hui-tests.el
```diff
@@ -932,6 +932,114 @@ With point on label suggest that ibut for rename."
...
+      (hy-delete-file-and-buffer file)))
+
+  ;; Outline Heading          link-to-string-match
+  (let ((file (make-temp-file "hypb" nil ".outl" "* heading\nbody\n")))
+    (unwind-protect
+        (progn
+          (find-file file)
+          (outline-mode)
+          (goto-char 1)
+          ;; BUG!? SHOULD link-to-string-match
```
matsl, line 1020: Can't get this to give a `link-to-string-match`. Sharing for info. Need to debug this more to understand if the test case is wrong or if the outline should look differently!?

test/hui-tests.el
```diff
@@ -932,6 +932,114 @@ With point on label suggest that ibut for rename."
...
+
+  ;; Mail Reader Message      link-to-mail
+  (let ((hmail:reader 'gnus-article-mode))
+    (with-temp-buffer
+      (gnus-article-mode)
+      (mocklet ((rmail:msg-id-get => "msg-id"))
+        (should (equal (caar (hui:link-possible-types)) 'link-to-mail)))))
+
+  ;; Directory Name           link-to-directory
+  ;; TBD
```
matsl, line 999: Not sure when this should kick in. What is the situation when we should get this?

test/hui-tests.el
```diff
@@ -932,6 +932,114 @@ With point on label suggest that ibut for rename."
...
+
+  ;; File Name                link-to-file
+  ;; TBD
```
matsl, line 1002: Same here: Not sure when this should kick in. What is the situation when we should get this?

test/hui-tests.el
```diff
@@ -932,6 +932,147 @@ With point on label suggest that ibut for rename."
...
+    (should (equal (caar (hui:link-possible-types)) 'link-to-texinfo-node)))
+
+  ;; Mail Reader Message      link-to-mail
+  (let ((hmail:reader 'gnus-article-mode))
+    (with-temp-buffer
+      (gnus-article-mode)
+      (mocklet ((rmail:msg-id-get => "msg-id"))
+        (should (equal (caar (hui:link-possible-types)) 'link-to-mail)))))
+
+  ;; Directory Name           link-to-directory
```
matsl, line 996: @rswgnu 

I don't have more energy to continue on this tonight so just reporting in and if you have some ideas for my next round on it.

I added the TBD cases but I think there are issues there. All tests pass and the ones where there are unexpected results have that documented for now as comments.  

I tried to add more cases setting `(let ((hargs:reading-type 'directory))` but that results in `link-to-ibut` response. I have tried to verify the behavior manually but I'm not that familiar with creating links from when prompted in the minubuffer so I used dired and the let-thingy so I might have got that wrong.

(Same behavior for the File Name link-to-file case below.)

My gut feeling is that some earlier functionality in the `cond` lathers kicks in but I have not verified that yet.

Having all these tests here might not be the right place. We should maybe move the different tests down to the hargs-functions that does the work of identifying what case it is, but I realized that first when debugging.

test/hui-tests.el
```diff
@@ -932,6 +932,147 @@ With point on label suggest that ibut for rename."
...
+      (hy-delete-file-and-buffer file)))
+
+  ;; Outline Heading          link-to-string-match
+  (let ((file (make-temp-file "hypb" nil ".outl" "* heading\nbody\n")))
+    (unwind-protect
+        (progn
+          (find-file file)
+          (outline-mode)
+          (goto-char 1)
+          ;; BUG!? Should be link-to-string-match?
```
matsl, line 1052: @rswgnu 

We missed this one today during the Sunday talk. If I get it right we should get a `link-to-string-match` on a org or outline header. Is that the expected behavior? I get a `link-to-file`. :shrug:

rswgnu:
I was having some trouble with the link-to-string-match so I think i
shifted to a positional file link instead.  Still need to look at this area
and decide if we add the persistent link id from Org mode when you use
org-store-link or something like that from org.

Bob

matsl:
Closing this in favor of #422 which contains basically the same changes.

test/hui-tests.el
rswgnu: Just acknowledging that I read this.  Thanks.

PR #415, opened 2023-12-07
Fix hyperbole-group typo

matsl:
# What

Fix hyperbole-group spelling.


rswgnu (commented):
Let's drop this one.  I am making a bunch of hyrolo.el  changes, so will fix this myself and then roll out the updated file.

PR #416, opened 2023-12-07
Add tests for hywconfig

matsl:
# What

Add tests for hywconfig.

# Why

This has not been covered before and there is a bug with restoring named configs.

### DIFF
--- a/test/hywconfig-tests.el
+++ b/test/hywconfig-tests.el
@@ -1,0 +1,100 @@
+;;; hywconfig-tests.el --- unit tests for hywconfig         -*- lexical-binding: t; -*-
+
+;; Author:       Mats Lidell <matsl@gnu.org>
+;;
+;; Orig-Date:    30-Jan-21 at 12:00:00
+;; Last-Mod:      7-Dec-23 at 23:17:36 by Mats Lidell
+;;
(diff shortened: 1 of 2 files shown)


test/hywconfig-tests.el
```diff
@@ -0,0 +1,100 @@
...
+(ert-deftest hywconfig--get-not-existing-config-errors ()
+  "Verify retrieving a config that has not been saved gives an error message."
+  (hywconfig-tests--remove-names)
+  (let ((err (should-error (hywconfig-restore-by-name "config") :type 'error)))
+    (should (string-match-p "No window configuration for this frame named" (cadr err)))))
+
+(ert-deftest hywconfig--add-by-name ()
+  "Verify config is added by name."
+  :expected-result :failed
+  ;; See backtrace for the error -  (wrong-type-argument listp #<window-configuration>)
```
matsl, line 78: It looks like this is the error. set:member gets a list instead of a single window-config!? Sounds related to the error in the current  TODOs.

Both failing test cases are set with :expected-result :failed to not break the CI-build and make it easier to merge the PR for later updates related to the error. i.e. if you approve of the tests they can go in and then we can work on the fix in a separate PR.

rswgnu (approved):

PR #423, opened 2023-12-26
Set enable-local-variables to nil for tests with DEMO and FAST-DEMO

matsl:
# What

Set enable-local-variables to nil for tests with DEMO and FAST-DEMO.

# Why

This avoids the query for the local vars when the tests are run interactively.

### DIFF
--- a/test/demo-tests.el
+++ b/test/demo-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    30-Jan-21 at 12:00:00
-;; Last-Mod:     11-Dec-23 at 01:58:53 by Bob Weiner
+;; Last-Mod:     25-Dec-23 at 23:30:20 by Mats Lidell
 ;;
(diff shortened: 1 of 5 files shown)


test/demo-tests.el
```diff
@@ -39,7 +39,7 @@
 (ert-deftest demo-smart-mouse-keys-ref-test ()
   "Go to the header from a #ref."
   (unwind-protect
-      (progn
+      (let ((enable-local-variables nil))
```
rswgnu, line 42: Just FYI, standard Lisp protocol is just:
   (let ((enable-local-variables))

to set it to nil but yours is of course more explicit, so let's let it ride right now.

rswgnu (approved):

test/demo-tests.el
matsl: Thanks. Yes, I know. I put that in on purpose. I want to be explicit here. I thought that it would communicate the meaning with the setting better that way!? Not sure if it does. :shrug: YMMW.

PR #424, opened 2023-12-28
Matsl rsw fix failing tests

matsl:
# What

Matsl rsw fix failing tests.

# Why

Two tests were failing due to strange circumstances. This seems to fix it. One problem was with that `y-o-n-p` behaves differently when running `interactive` and `noniniteractive`. So we need to accommodate for that. An alternative would be to mock `y-o-n-p`  to always return t and not depend on any input given!?

For this version I have chosen to keep requiring the input but add a RET if run noninteractively. Consequently I have reinstalled the use of `with-simulated-input` to be consistent with our other tests. Using `kbd-key:key-series-to-events` looks better because it does not crash or complain when not all its input is used. That is a good feature of `with-simulated-input.`

To be fair, I have seen some other strange things in other, unrelated to this test case, with `with-simulated-input,` but I need to understand better what is going on there. So I'll come back to that.

# Note

Did also do some refactoring and changed visiting "*HyRolo*" buffer into actually checking that the search brings us there. Which is the expected behavior.

### DIFF
--- a/test/hyrolo-tests.el
+++ b/test/hyrolo-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    19-Jun-21 at 22:42:00
-;; Last-Mod:     26-Dec-23 at 01:57:38 by Bob Weiner
+;; Last-Mod:     27-Dec-23 at 16:52:20 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


rswgnu (approved):

rswgnu:
Merged your PR to my local and pushed my updates.  Last run of all tests I
got 2 beeps but all tests passed.

Bob

rswgnu:
Here was the failing test from the Emacs master version.  Special chars in
a regexp were not quoted.  Hopefully the fix in my latest push resolves it.

Test hyrolo-fgrep-and-goto-next-visible-md-heading backtrace:
signal(error ("(hyrolo-grep-file): Stuck looping in buffer \"hypbsvK
error("(hyrolo-grep-file): Stuck looping in buffer \"%s\" at positio
hyrolo-grep-file("/tmp/hypbpSWnWw/hypbsvKzIE.md" "string" nil nil ni
hyrolo-grep("string" nil nil nil nil nil)
hyrolo-fgrep("string")
(progn (hyrolo-fgrep "string") (save-current-buffer (set-buffer "*Hy
(unwind-protect (progn (hyrolo-fgrep "string") (save-current-buffer
(let* ((temporary-file-directory (make-temp-file "hypb" t)) (md-file
(closure (t) nil (let* ((temporary-file-directory (make-temp-file "h
ert--run-test-internal(#s(ert--test-execution-info :test #s(ert-test
ert-run-test(#s(ert-test :name hyrolo-fgrep-and-goto-next-visible-md
ert-run-or-rerun-test(#s(ert--stats :selector t :tests [#s(ert-test
ert-run-tests(t #f(compiled-function (event-type &rest event-args) #
ert-run-tests-batch(nil)
ert-run-tests-batch-and-exit()
(let ((auto-save-default)) (load-file "test/demo-tests.el") (load-fi
eval((let ((auto-save-default)) (load-file "test/demo-tests.el") (lo
command-line-1(("--eval" "(progn (setq debug-on-error t) (setq backt
command-line()
normal-top-level()
Test hyrolo-fgrep-and-goto-next-visible-md-heading condition:
(error
"(hyrolo-grep-file): Stuck looping in buffer \"hypbsvKzIE.md\" at position
11")

PR #430, opened 2024-01-03
Require hbut so defib is defined, needed for compiling

matsl:
# What

Require hbut so defib is defined, needed for compiling.

# Why

Without hbut the macro defib is not defined and the compile errors.

# Note

Fllymake can as well be required since it is a standard part of Emacs since 2003 and fixed a double spacing issue while at it.

### DIFF
--- a/hsys-flymake.el
+++ b/hsys-flymake.el
@@ -7,7 +7,7 @@
 ;; E-mail:       rsw@gnu.org
 ;;
 ;; Orig-Date:    31-Dec-23 at 13:54:08
-;; Last-Mod:      3-Jan-24 at 13:41:18 by Bob Weiner
+;; Last-Mod:      3-Jan-24 at 15:45:33 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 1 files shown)


hsys-flymake.el
```diff
@@ -23,13 +23,16 @@
...
 ;;    takes effect, but we don't require `flymake-mode' or
-;;    `repeat-mode'. Instead we leave it to each function to check
+;;    `repeat-mode'.  Instead we leave it to each function to check
 ;;    whether it is enabled and active in the current buffer.  This way,
 ;;    if the user never uses `flymake-mode' that library is never loaded
 ;;    and the implicit button type never triggers.
 ;;
 ;;; Code:
 
+(require 'flymake)
```
rswgnu, line 33: Read the above code comment for why we don’t want to require flymake.  If symbols are needed, just declare them.

rswgnu (changes requested):

hsys-flymake.el
matsl: :rofl: Sorry. Was a late night change to make the thing compile and when I was at it I throw in the require to remove warnings. This should work as expected now.

PS. Not particular happy about this type of design but might be the best or only way. Let's come back to this later. Don't want to derail and take time on this now. I'm happy with the solution for now and I need to think about possible alternatives. DS. :pray:

rswgnu (approved):

rswgnu:
Approved.On Jan 4, 2024, at 9:31 AM, Mats Lidell ***@***.***> wrote:﻿
@matsl requested your review on: #430 Require hbut so defib is defined, needed for compiling.

PR #431, opened 2024-01-04
Remove flymake/flycheck/byte-compile warnings

matsl:
# What

Remove flymake/flycheck/byte-compile warnings.

# Why

Warnings should be understood and eliminated.


hywconfig.el
```diff
@@ -204,7 +204,8 @@ oldest one comes the newest one."
 
 (defun hywconfig-named-get (name)
   "Return the selected frame's window configuration with NAME."
-  (lax-plist-get (hywconfig-named-get-entries) name))
+  (with-suppressed-warnings ((obsolete lax-plist-get)) ;; Obsolete since 29.1, use plist-get
+    (lax-plist-get (hywconfig-named-get-entries) name)))
```
rswgnu, line 208: We need to compare items with 'equal' here.  I don't think plist-get will do that, will it?  Maybe they have added an arg that adds the comparison operator but then that isn't backwards compatible, so we're better off with the 'lax' call for now.

rswgnu (approved):

hywconfig.el
matsl: I don't know more than what the byte compiler reports which is "deprecated since 29.1, use plist -get" I would expect such a statement to mean it is a drop in replacement but I guess it could also mean that you should use plist-get and change your code accordingly. :shrug: 
Either way we are good now since we are using lax-plist-get. It is when it is removed we will have to worry about this. 

I'm still confused about the proper way to deal with obsolete functions. Shall we have our own function that hides the call to the obsolete function dependent on emacs-version, use the backwards compatible library or is there some other way!? Other packages must face the same issue so it ought to be just to pick up a good pattern from there.

PR #432, opened 2024-01-05
xref is used by etags since 26.1 so use it unconditionally

matsl:
# What

`xref` is used by `etags` since 26.1 so use it unconditionally.

# Why

Using a separate PR for this since it is a more complex change to remove warnings. The conditional definitions cause warnings. It is likely fixed by forward declarations but maybe there is a simpler solution to not declare them conditionally. Se reasoning below.

As I understand xref is used by `etags` since 26.1. Do we still need to check if it is present? It should be available in all versions we support.

And is the check for `xref-definition` needed. I can only find it in Hyperbole. It is not part of Emacs 27.2 what I can see. 

Can we not just define them without any conditions? That avoids all warnings too related to this.

# Note
Do I understand right that the `xref-definition,` `xref-item-buffer` and `xref-item-position` are Hyperbole local functions. I can't find them in the `xref` package. What do you think of giving them a proper prefix such as `hypb-xref` or similar? They are used in only a few places so should be easy to replace.

The supplied PR assumes we can skip both tests mentioned above.

### DIFF
--- a/hmouse-tag.el
+++ b/hmouse-tag.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    24-Aug-91
-;; Last-Mod:      5-Jan-24 at 14:02:46 by Bob Weiner
+;; Last-Mod:      5-Jan-24 at 23:01:12 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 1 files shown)


rswgnu:
Yes, I agree in general with all of this.

rswgnu (approved):

matsl:
So you want me to go ahead with the suggested prefix change or we just merge this for now?

rswgnu:
Let me add the prefixes and figure it out.  Add this to my todo list too.  You finish up your test cases for HyRolo or Koutliner.-- BobOn Jan 5, 2024, at 10:48 PM, Mats Lidell ***@***.***> wrote:﻿
So you want me to go ahead with the suggested prefix change or we just merge this for now?

rswgnu:
You can merge it to rsw as is.-- BobOn Jan 5, 2024, at 11:08 PM, Robert Weiner ***@***.***> wrote:﻿Let me add the prefixes and figure it out.  Add this to my todo list too.  You finish up your test cases for HyRolo or Koutliner.-- BobOn Jan 5, 2024, at 10:48 PM, Mats Lidell ***@***.***> wrote:﻿
So you want me to go ahead with the suggested prefix change or we just merge this for now?

PR #434, opened 2024-01-06
Matsl rsw mock revisit file query

matsl:
Check that is works across all versions.


matsl:
Merged into rsw branch.

PR #436, opened 2024-01-07
Add more hyrolo show, hide and movement tests

matsl:
# What

Matsl rsw complete hyrolo hiding tests.

# Why

This completes the hyrolo tests for hide/show/movement  in respect to the TODO item for hyrolo tests. When more specific tests are needed lets create new TODO-items for those to avoid having long lasting TODO-items.

# Note

The PR contains skip guards in test cases depending on org folding functionality. i.e. org fold changes in org 9.6. When hyrolo is updated to handle the incompatible changes in org mode these guards shall be removed.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,15 @@
+2024-01-09  Mats Lidell  <matsl@gnu.org>
+
+* test/hyrolo-tests.el (hyrolo-tests--outline-next-visible-heading-md)
+    (hyrolo-tests--outline-next-visible-heading-all)
+    (hyrolo-tests--outline-next-visible-heading-kotl)
+    (hyrolo-tests--outline-next-visible-heading-all-file-types)
+    (hyrolo-tests--outline-show-when-moving-out-of-hidden-line): Add more
+    test cases for hyrolo outline show, hide and movement.
+
+    (hyrolo-tests--gen-kotl-outline, hyrolo-tests--verify-hidden-line)
+    (hyrolo-tests--verify-not-hidden-line): Add helper functions and use.
+
 2024-01-09  Bob Weiner  <rsw@gnu.org>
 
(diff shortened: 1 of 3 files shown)


rswgnu:
How about if we put tests that are not ready for deployment somewhere where they won’t be run but we can manually run them all from one file and they are all logged to the todo list for fixing?

Nothing should go into master until it is ready for release, i.e. no todos, etc., just plain working code and comments.  We can discuss on a call how best to achieve this.

Thanks,

-- Bob

matsl:
> How about if we put tests that are not ready for deployment somewhere where they won’t be run but we can manually run them all from one file and they are all logged to the todo list for fixing?

By guarding tests for not running, when certain conditions are not met for the code that is tested to work properly, is actually helpful for fixing bugs in the code that are already in. Example: The tests that checks the code that depends on the org fold code in this PR is what I believe correct. The code that is in the rsw branch that is tested is broken. So by having the test in the rsw branch does bot hurt at all, quite the contrary. When we work on the fix it should just be to remove the guard and run the normal test suite.

rswgnu:
Let’s try it your way and see if it helps.-- BobOn Jan 7, 2024, at 6:10 PM, Mats Lidell ***@***.***> wrote:﻿

How about if we put tests that are not ready for deployment somewhere where they won’t be run but we can manually run them all from one file and they are all logged to the todo list for fixing?

By guarding tests for not running, when certain conditions are not met for the code that is tested to work properly, is actually helpful for fixing bugs in the code that are already in. Example: The tests that checks the code that depends on the org fold code in this PR is what I believe correct. The code that is in the rsw branch that is tested is broken. So by having the test in the rsw branch does bot hurt at all, quite the contrary. When we work on the fix it should just be to remove the guard and run the normal test suite.

rswgnu (approved):

PR #439, opened 2024-01-08
Matsl rsw fix not great looking  imho  local variable header

matsl:
# What

Matsl rsw fix not great looking imho local variable file header.

# Why

The local variable and mode line first in the docs files does not look that great IMHO. Using a local variables section will accomplish the same settings but might also be surprising to the users. So I'm a bit conflicted here but not conflicted enough not to share this idea with you. I'm sure you have considered it already but would like to hear your view on this.

# Note

This is just a partial PR for the discussion. I believe there are other files that could go with this change and use a Local Variables section.

### DIFF
--- a/DEMO
+++ b/DEMO
@@ -1,5 +1,3 @@
--*- Mode: org; org-cycle-global-at-bob: t; hsys-org-enable-smart-keys: t -*-
-
 * GNU Hyperbole Full Demo/Tutorial by Bob Weiner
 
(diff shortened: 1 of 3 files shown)


rswgnu:
Load one of these files with {C-h h d d}.  Then press {TAB} repeatedly to see it cycle through the heading visibility states.
I don't think you can do that if the variables are at the end of the file.

rswgnu:
Closing, this is no longer needed since we solved the problem without local variables.

PR #441, opened 2024-01-13
Matsl rsw strange auto mode alist org issue

matsl:
# What

Demoing the strange auto mode alist org issue.

# Why

Seems we have a regression whith org mode being removed from the auto-mode-alist. There are also  other org related errors, like missing variable.

See commit for the actual change: https://github.com/rswgnu/hyperbole/pull/441/commits/f0837c214518f176a3bc70801d44134d6b4bf88f


rswgnu:
Closing this.  It is no longer needed as this has been fixed in the 'rsw' branch.

PR #442, opened 2024-01-14
Remove tar-ball dependency on elc files

matsl:
# What

Remove tar-ball dependency on elc files. ~The el-files are now compiled under pkg_hyperbole as part of the tar-ball creation.~

# Why

There was a dependency that caused the elc-file to be required before the pkg_hyperbole build step. ~That caused make to compile them to early.~ This PR removes that dependency ~and builds the elc-files as part of creating the tar-ball in the pkg_hyperbole folder.~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Author:       Bob Weiner
 #
 # Orig-Date:    15-Jun-94 at 03:42:38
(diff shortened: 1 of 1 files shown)


rswgnu (approved):

matsl:
Removed the byte compilation step so that elc will not be part of the tar-ball.

rswgnu (approved):

PR #443, opened 2024-01-14
Matsl rsw forward backward same level tests

matsl:
# What

Matsl rsw forward backward same level tests.

# Note

While cleaning up I noticed this branch with tests for moving forward and backwards at the same level. One of the tests fail due to point jumping over a few lines. Added it as "expected-result failed."

Here are the details. The \*HyRolo\* file.

```
===============================================================================
@loc> "/tmp/hypbVG9ksf.org"
===============================================================================
* heading-org 1
body-org 1
===============================================================================
@loc> "/tmp/hypbOQBeuv.md"
===============================================================================
# heading-md 1
body-md 1
===============================================================================
@loc> "/tmp/hypbrGtujm.otl"
===============================================================================
* heading-otl 1
body-otl 1
===============================================================================
@loc> "/tmp/hypbDIK8zl.kotl"
===============================================================================
   1. heading-kotl
      body-kotl
```
When pressing 'f' on "* heading-org 1" point moved immediately down to "* heading-otl 1". Should it not land on the next "===" line? (And next go to "# heading-md 1" etc...) 

On the way up with 'b' it also does not go to the top. It stays at line 2. "@loc> "/tmp/hypbVG9ksf.org"".

I might be confused here since I thought this was solved. But there has been multiple PRs and I think I just communicated this problem without sharing test case and with no PR.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2024-01-15  Mats Lidell  <matsl@gnu.org>
+
+* test/hyrolo-tests.el (hyrolo-tests--forward-same-level-all-file-types-level1)
+    (hyrolo-tests--forward-same-level-org-level2): Add tests for hyrolo
+    forward and backward movement on same level.
+
 2024-01-14  Bob Weiner  <rsw@gnu.org>
 
 * hypb.el (require 'org): Add to define 'org-show-all'.  Fix typo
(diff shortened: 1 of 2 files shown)


rswgnu:
This had been working.  We’ll have to see what happened.-- BobOn Jan 14, 2024, at 6:52 PM, Mats Lidell ***@***.***> wrote:﻿
@matsl requested your review on: #443 Matsl rsw forward backward same level tests.

rswgnu (approved):

PR #445, opened 2024-01-16
Print build environment info when using eln target

matsl:
# What

Matsl rsw echo emacs and env info

# Why

We need to show emacs version when compiling natively to make it easier to see that the proper version is used.

# Note

Reuses the echo target for printing the environment and added emacs version info to echo target as well.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2024-01-16  Mats Lidell  <matsl@gnu.org>
+
+* Makefile (eln): Use echo target for showing build info as part of the
+    eln build.
+    (echo): Show Emacs location and version in build environment info.
+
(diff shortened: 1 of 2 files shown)


rswgnu (approved):

rswgnu:
Thanks, Mats.  That works but can you tell me how to override the EMACS var value on the command line.  Adding ‘EMACS=path’ either befor or after ‘make eln’ did not work for me.  -- BobOn Jan 16, 2024, at 10:24 AM, Mats Lidell ***@***.***> wrote:﻿
Merged #445 into rsw.

matsl:
Ah. You want to be able to control which Emacs is used not only get a confirmation which is!? In that case we need to do is differently. New PR will come.

PR #447, opened 2024-01-17
Remove skip guard on test that works in batch mode

matsl:
# What

Remove skip guard on test that works in batch mode.

# Why

Try to run as many tests as possible in batch mode.

# Note 1

It seems most tests that has the guard for not running them in batch mode actually requires that. I only found one test.

I have not tested to run all with skip guards in batch mode. Instead I have tested a few and if they failed to run I have looked for calling patterns in related test cases. If they use the same setup I have concluded that they probably not will work in batch mode either.

# Note 2 

I have some good thing to mention now though. While working on the presentation I found a hack that we might be able to use to run the interactive tests in the CI/CD. It passes my first limited test anyway. It is on my todo to play more with that. I think it is post 9.0 release though. We can run the interactive tests locally using docker if we want to sanity check before the release.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2024-01-17  Mats Lidell  <matsl@gnu.org>
+
+* test/hui-select-tests.el (hui-select--thing): Remove not interactive
+    guard.
+
 2024-01-16  Mats Lidell  <matsl@gnu.org>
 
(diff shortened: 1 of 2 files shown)


rswgnu:
Sounds good.

rswgnu (approved):

PR #449, opened 2024-01-19
Cleanup of flycheck/flymake info and warning messages

matsl:
# What

Cleanup of flycheck/flymake info and warning messages.

# Why

We should keep the code base as clean as possible.

# Note

This is a more or less mechanical go through of all info and warnings produce by flycheck and trying to resolve them. (I hope and think that flycheck uses the same backbone functions to check for problems as flymake so this work would cover for a similar exercise using flymake.)

I know you are not a big fan of these huge PRs but it is mostly small changes affecting docstrings. There are a few code changes but should not affect execution.

Common things that are fixed are:
1. Error message shall not end in a period.
2. There shall be no extra whitespace at end of line.
3. Docstring is missing.
4. Docstring is missing arg in capital letters.
5. Key control sequence is not properly escaped and quoted.
6. Dired should be written as Dired, with a capital first letter, everywhere in docstrings.
7. ...

Not all of these are fixed either. Not all functions have been given docstrings. Not all parameters have been added in capital letters.I might have missed something. Anyway -- I have tried fix that but given up due to it either does not make sense to add it or most often, I could not come up with something decent enough so juts left it as is.

There is no Changelog for this work either. If it gets accepted I can add a Changelog item covering the whole initiative.

### DIFF
--- a/hypb.el
+++ b/hypb.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     6-Oct-91 at 03:42:38
-;; Last-Mod:     20-Jan-24 at 20:20:28 by Mats Lidell
+;; Last-Mod:     20-Jan-24 at 20:22:08 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -444,7 +444,8 @@ This will install the Emacs devdocs package if not yet installed."
 
 (defun hypb:error (&rest args)
   "Signal an error typically to be caught by `hyperbole'.
-The error message is formatted passing ARGS to the function `format'."
+The error message is formatted passing the rest of the ARGS to
+the `format' function."
   (let ((msg (if (< (length args) 2)
 		 (car args)
 	       (apply 'format (cons (car args)
--- a/hypb.el
+++ b/hypb.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     6-Oct-91 at 03:42:38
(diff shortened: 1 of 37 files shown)


rswgnu:
Approved the lint PR and working my way through this good, long one, as I get time today.  -- BobOn Jan 19, 2024, at 7:01 PM, Mats Lidell ***@***.***> wrote:﻿
@matsl requested your review on: #449 Cleanup of flycheck/flymake info and warning messages.

hypb.el
```diff
@@ -443,7 +443,8 @@ This will install the Emacs devdocs package if not yet installed."
       (concat "@" dname))))
 
 (defun hypb:error (&rest args)
-  "Signal an error typically to be caught by `hyperbole'."
+  "Signal an error typically to be caught by `hyperbole'.
+The error message is formatted passing ARGS to the function `format'."
```
rswgnu, line 447: "Signal an error typically to be caught by `hyperbole'.
The error message is formatted passing the rest of the ARGS to the
`format' function."

hyrolo.el
```diff
@@ -1262,6 +1265,7 @@ Return number of matching entries found."
     (hyrolo-grep-file hyrolo-file-or-buf regexp max-matches count-only)))
 
 (defun hyrolo-bbdb-entry-format (entry)
+  "Format for a BBDB ENTRY."
```
rswgnu, line 1268: "Format for a Bbdb ENTRY."

Or you could change the variable name to BBDB-ENTRY, if you prefer.

rswgnu (changes requested):
Looks very good.  Just a number of simple updates and should be good to go.

rswgnu (approved):

PR #459, opened 2024-01-31
Matsl rsw hyrolo reveal mode tests

matsl:
# What 

Add one hyrolo reveal mode test.

# Note

This is based on #458 so see last commit for the new test or look at the end.

This is partially to show the result to get a review if it is the expected behavior. 

It is also to show the strange issue with getting the display redrawn. I had TAB through the matches and using the helper `(hy-test-helpers:consume-input-events)`. Calling the function `hyrolo-next-match` did not work. In the first occurrence I even had to call it twice!? Se comment below. (So this is likely work in progress to understand things better,)

### DIFF
--- a/test/hyrolo-tests.el
+++ b/test/hyrolo-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    19-Jun-21 at 22:42:00
-;; Last-Mod:     30-Jan-24 at 14:24:16 by Mats Lidell
+;; Last-Mod:      1-Feb-24 at 00:10:57 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


test/hyrolo-tests.el
```diff
@@ -1067,5 +1067,261 @@ body
...
+
+(defun hyrolo-tests--hyrolo-section-header (filename)
+  "Return a hyrolo section header for FILENAME."
+  (concat
+   "===============================================================================\n@loc> \""
+   filename
+   "\"\n"
+   "===============================================================================\n"))
+
+(ert-deftest hyrolo-tests--hyrolo-reveal-mode ()
```
matsl, line 1266: New test added in this PR.

test/hyrolo-tests.el
```diff
@@ -1067,5 +1067,261 @@ body
...
+          (hyrolo-grep "body")
+          (hyrolo-top-level)
+
+          (should (string=
+                   (concat (hyrolo-tests--hyrolo-section-header org-file1)
+                           "* h-org 1...\n* h-org 2...\n")
+                   (hyrolo-tests--outline-as-string)))
+
+          (should (hact 'kbd-key "TAB"))
+	  (hy-test-helpers:consume-input-events)
```
matsl, line 1282: In order to make the test work on first run I have to call this function twice!? Without this the test fails when run with other tests like this (ert "hyrolo"). If just run after failure from *ert* buffer as a single test then after a few tries it starts to work and then it seems to never fail again!? Is there some caching involved here that can explain this odd behavior?

test/hyrolo-tests.el
```diff
@@ -1067,5 +1067,261 @@ body
...
+  "Return a hyrolo section header for FILENAME."
+  (concat
+   "===============================================================================\n@loc> \""
+   filename
+   "\"\n"
+   "===============================================================================\n"))
+
+(ert-deftest hyrolo-tests--hyrolo-reveal-mode ()
+  "Verify hidden sections are shown when using {TAB} to move through matches."
+  (skip-unless (not noninteractive))
```
matsl, line 1268: Requires interactive run.

rswgnu (approved):

PR #462, opened 2024-02-02
Matsl rsw fix ebut act

matsl:
# What

Rename i- ebut:act to act-label and create new act functions.

# Why

The old functions takes a label and to be more align with the hbut:act they should take a button. So the old functions have been moved away and new ones have been created. Two very simple tests have been added.

# Note

These function are not used a lot. The work horse is hbut:act that centralizes button action. So I'm not sure how well used these will be since you can call hbut:act in most cases just as well. In fact it its current version there is no check that hbut:current  actually is of the right type. Should that be added? Anyway, not sure the use case adds that much, but makes the API more complete.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,8 @@
 2024-02-08  Mats Lidell  <matsl@gnu.org>
 
+* hbut.el (ibut:act-label): Rename ibut:act to ibut:act-label, it takes a
+    label as arg.
+
 * hui-menu.el (hui-menu-explicit-buttons): Use ebut:act-label.
 
 * hbut.el (ebut:act-label): Rename ebut:act to ebut:act-label, it takes a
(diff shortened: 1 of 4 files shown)


rswgnu (approved):

rswgnu:
I agree that as hbut:act does a check that the button sent in is valid, so
too should ibut:act and ebut:act (the new versions you wrote).

rswgnu:
Have to discuss this on the phone.  Too involved for me to think through it
in a message right now.  Trying to do some work on HyRolo to eliminate any
issues so we can ship.  -- Bob

matsl:
> Have to discuss this on the phone.  Too involved for me to think through it
in a message right now.  Trying to do some work on HyRolo to eliminate any
issues so we can ship.  -- Bob

No problem. You should keep focused on fixing the known "old" things. 

We can keep this as is for now and fix it post the release. Or I can add a check that guards the functions from running a but of the other type and use that both for a supplied arg and for hbut:current. We can leave the more deep discussion to later to not waste time on that now.

matsl:
@rswgnu Added check on type of but and corresponding test cases. PTAL

matsl:
@rswgnu Did an update on the error message, renamed the parameters and added a test for when there is no hbut:current. PTAL

rswgnu (approved):

PR #463, opened 2024-02-02
Matsl rsw silence warnings again

matsl:
# What

Silences some compiler warnings.

# Why

Work with markdown has introduced a few vars and functions that needs to be declared to silence byte compiler. Plus some other small things.

### DIFF
--- a/hyrolo.el
+++ b/hyrolo.el
@@ -1560,7 +1560,7 @@ Return number of matching entries found."
       (insert "No result.")
     (print contacts (get-buffer-create "*contacts-data*"))
     (dolist (contact contacts)
-      (let* (
+      (let* ((child nil)
 	     (name-value (nth 0 (xml-get-children contact 'gd:name)))
              (fullname (xml-node-child-string (nth 0 (xml-get-children name-value 'gd:fullName))))
              (givenname (xml-node-child-string (nth 0 (xml-get-children name-value 'gd:givenName))))
--- a/hyrolo.el
+++ b/hyrolo.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     7-Jun-89 at 22:08:29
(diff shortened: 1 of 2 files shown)


hyrolo.el
```diff
@@ -1560,7 +1584,7 @@ Return number of matching entries found."
       (insert "No result.")
     (print contacts (get-buffer-create "*contacts-data*"))
     (dolist (contact contacts)
-      (let* (
+      (let* ((child nil)
```
matsl, line 1587: This let variable was oddly removed in the past to silence warning about unused let. Not sure what happened but it does look like it is needed again but no code seems changed!? Well, well.

hyrolo.el
```diff
@@ -113,10 +113,33 @@
...
+(declare-function markdown-gfm-checkbox-after-change-function "ext:markdown-mode")
+(declare-function markdown-imenu-create-flat-index "ext:markdown-mode")
+(declare-function markdown-imenu-create-nested-index "ext:markdown-mode")
+(declare-function markdown-line-is-reference-definition-p "ext:markdown-mode")
+(declare-function markdown-live-preview-if-markdown "ext:markdown-mode")
+(declare-function markdown-live-preview-remove-on-kill "ext:markdown-mode")
+(declare-function markdown-make-gfm-checkboxes-buttons "ext:markdown-mode")
+(declare-function markdown-pipe-at-bol-p "ext:markdown-mode")
+(declare-function markdown-remove-gfm-checkbox-overlays "ext:markdown-mode")
+
```
rswgnu, line 142: I think some updates I have made but haven't yet pushed eliminate the need for all of these forward declarations but we will have to see.

rswgnu (commented):
I think you'll have to wait for my merge and then see what of this one is still necessary.

hyrolo.el
matsl: Fine. I'll hold my horses on this one.

rswgnu (approved):
Looks like we do need these declarations!

PR #474, opened 2024-02-27
Sanity check kotl-mode user commands

matsl:
# What

Sanity check kotl-mode user commands.

# Why

Does a rough check that most user level commands in kotl-mode is not through an error on a simple task. It does not verify that the outcome of the command is correct so is more like a smoketest or sanity check.

# Note

Not all commands can be executed by just giving static arguments from a fixed point in the kotl-file. So these commands need either more advance setup or to be tested separate. They are kept for now in the test data with the ignore field set to t so that they can be filtered out from execution.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2024-02-27  Mats Lidell  <matsl@gnu.org>
+
+* test/kotl-mode-tests.el (kotl-mode-tests--gen-kotl-outline)
+    (kotl-mode-tests--func, kotl-mode-tests--sanity-check-function-list)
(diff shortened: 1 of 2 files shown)


matsl:
For some reason the "unrelated" test kotl-mode-kill-contents fails so must be some interesting interaction between the test cases... huh!?

matsl:
Had to set `transient-mark-mode` for tests to work. `kotl-mode:kill-region` depends on that and it seems the old test for some reason worked without that!? :shrug:

rswgnu (approved):

PR #481, opened 2024-03-06
Pick next window from selected frame

matsl:
# What

Only pick next window from the selected frame.

# Why

I think the intent is to from a spit window situation pick one window
as the link-but-window and the other as the referent-window. In the
call to next-window we need then to limit the choices to the selected
frame.

# Note

A unit test for hmouse-choose-link-and-referent-windows would have
been appropriate but at the moment that became to complicated. The
function mixes interactive events with checking the window setup and
I'm not sure we can mock that in a good way. So for now I have put
that effort aside.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,11 @@
-2024-03-06  Mats Lidell  <matsl@gnu.org>
+2024-03-08  Mats Lidell  <matsl@gnu.org>
+
+* test/hmouse-drv-tests.el (hmouse-drv--hmouse-choose-link-and-referent-windows--two-windows-same-frame)
+    (hmouse-drv--hmouse-choose-link-and-referent-windows--two-windows-different-frame):
+    Add tests for choosing the windows non interactively.
 
 * hmouse-drv.el (hmouse-choose-link-and-referent-windows): Only pick next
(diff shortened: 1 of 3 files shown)


rswgnu:
Hi Mats:  Why do you think the windows have to be in the same frame?

matsl:
Because of the check on line 806 if there are two windows in the frame!?

rswgnu:
But line 807 in the same cond clause checks across all visible frames.  Why don’t you make a separate clause for that and that should resolve any issue.-- BobOn Mar 7, 2024, at 1:33 AM, Mats Lidell ***@***.***> wrote:﻿
Because of the check on line 806 if there are two windows in the frame!?

matsl:
Hi, was on the phone yesterday so the short answer. Sorry about that. Great idea to split the cases so I did that and added two test cases. Seems it is failing for 27.2 so I need to take a look at that later.

matsl:
Ah. Found it. Used `length=` which is a function from 28.1. Will fix this later today when back at keyboard.

matsl:
@rswgnu Fixed the test for 27.2. PTAL

rswgnu (approved):

PR #486, opened 2024-03-13
Add tests for verifying action key on hyrolo matches

matsl:
# What

Add tests for verifying action key on hyrolo matches

# Why

I noticed a problem with kotl files if the first line contains a slash
'/'.  So added a few tests to explore that problem. The problematic
test is marked as expected failure.

# Note

The problem is that the kotl-files contents "display" is changed and the file is not a any longer a valid kotl-file. The file stored on disk is not changed and is still a valid kotl-file. So something happend when the action key is used to move there when there is a a slash in the line!?

### DIFF
--- a/test/hyrolo-tests.el
+++ b/test/hyrolo-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    19-Jun-21 at 22:42:00
-;; Last-Mod:     12-Mar-24 at 23:04:11 by Mats Lidell
+;; Last-Mod:     14-Mar-24 at 00:26:34 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


test/hyrolo-tests.el
```diff
@@ -1513,6 +1513,108 @@ body
       (kill-buffer hyrolo-display-buffer)
       (hy-delete-files-and-buffers hyrolo-file-list))))
 
+(ert-deftest hyrolo-tests--goto-org-match ()
+  "Verify that moving to a match reaches the target buffer."
```
rswgnu, line 1517: Need to change the doc string to match the test for each of the fun tions you added to this buffer.

test/hyrolo-tests.el
```diff
@@ -1513,6 +1513,108 @@ body
...
+      (hy-delete-files-and-buffers hyrolo-file-list))))
+
+(ert-deftest hyrolo-tests--goto-kotl-header-with-slash-match ()
+  "Verify that moving to a match reaches the target buffer."
+  :expected-result :failed
+  (let* ((kotl-file1 (hyrolo-tests--gen-kotl-outline "h/kotl" "body" 1))
+         (hyrolo-file-list (list kotl-file1)))
+    (unwind-protect
+        (progn
+          (hyrolo-grep "h/kotl")
```
rswgnu, line 1602: /  is the child separator in hyrolo searches, so this means find the ‘h’ node and then a child node starting with ‘kotl’.  Should be documented in the manual.  So the test needs rewriting.  Maybe we should not have the same behavior in hyrolo-fgrep and hyrolo-grep.

rswgnu (changes requested):

test/hyrolo-tests.el
matsl: I will fix. I wanted to you too see it so I hacked and pushed away! ;-)

test/hyrolo-tests.el
matsl: Ah I see. OK. To bad the point I was trying to make got lost because it got masked by that feature. So you are right that the test case needs to be rewritten but not because of the child separator. It is a problem with having a / in the first line in the cell if there is a match on that line. To late this night to rewrite it so I will redesign it tomorrow to better prove the point. Or I can show you a real life example on Sunday.

matsl:
@rswgnu Updated docstrings and reworked the tests slightly. Most important. Removed the slash in the search since that has special meaning i hyrolo but kept it in the contents of the cells. There are test cases both for slash in the first line in the cell and in the second line in the cell. There is different behavior.

rswgnu (approved):

PR #489, opened 2024-03-16
Add test for ibut:at-p on a remote path name

matsl:
# What

Add test for ibut:at-p on remote path names.

# Why

There is a problem when point is in the email address part of the
remote name. It is identified as an email address so instead of
opening the remote file the email editor is opened.

The test case narrows it down a bit to ibut:at-p that is tested on
three positions in the remote file name. Before and after the email
portion and then on the email portion (which fails).

The test case is consequently configured as expected failed to
document the problem.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2024-03-16  Mats Lidell  <matsl@gnu.org>
+
+* test/hbut-tests.el (hbut-tests--ibut-at-p-identifies-a-remote-pathname):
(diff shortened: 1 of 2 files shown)


matsl:
@rswgnu Bob, I found another issue and documented it with a test case. Maybe the fix is just to check for pathname before email-address or we need to have a more advanced check for email-address that noticed if it is part of a remove file name. Leaving the fix for later in any case.

rswgnu (approved):

PR #490, opened 2024-03-16
Add target to run make with a dockerized Emacs

matsl:
# What

Add target to run make with a dockerized Emacs.

# Why

It is good to easily be a be able to test with alternative versions of
Emacs.


matsl:
@rswgnu First attempt to support dockerized developer builds.

# Usage:
#   make dockerized version=28.1 targets='clean bin test'

So one target and then with parameters you specify version and which make file targets you want to run! More to type but flexible. What would support you best in your development?

rswgnu (approved):

Issue #491, opened 2024-03-18
Is it possible for hyperbole to create links to words in the org-mode?


Odysseus6000:
Is it possible for hyperbole to create links it can link to words with plain text content in the org-mode file?
The new explicit link button created with {C-h h e l} links to a location in the org-mode file. The way it creates a link between two windows is very convenient and can be used with ace-window. These are amazing.If the file is modified and this location is moved, it won't link to the original location of the text? The rules of org-mode require the addition of <<Tagret>> to link to the text content, and the need to keep <<Tagret>> unique.
Another way is to create an ID attribute for the headline. But for a paragraph, splitting the headline will destroy the original content.

Odysseus6000:
For example: 
* Test1
    ...
    There are many important cases for companies in this situation. 
     ...

Below that there is a headline labeled Test1, and below it is a paragraph of text... indicates that some of the plain text content has been omitted. I want to link to the important case these words . If I use the implicit button link the link is formed as follows
"~/file.org:100:25"
The position of "important case" is 100:25. i don't want to set the line where the important case is located as headline. is there any other way to link to important case and still link to it when the position of the paragraph where these words are located changes.?

rswgnu:
Manually, you can just type the implicit link like so:“~/file.org#important cases”We will have to think about how this case could be automated.— rswOn Mar 20, 2024, at 12:57 AM, Odysseus6000 ***@***.***> wrote:﻿
For example:

Test1
...
There are many important cases for companies in this situation.
...

Below that there is a headline labeled Test1, and below it is a paragraph of text... indicates that some of the plain text content has been omitted. I want to link to the important case these words . If I use the implicit button link the link is formed as follows
"~/file.org:100:25"
The position of "important case" is 100:25. i don't want to set the line where the important case is located as headline. is there any other way to link to important case and still link to it when the position of the paragraph where these words are located changes.?

matsl:
@rswgnu 
> We will have to think about how this case could be automated.

Would it make sense if we create a string match link if the "target" buffer region is active?

rswgnu:
Mats suggestion above has been added to the latest Hyperbole pre-release, 8.0.2pre, available now from elpa-devel.  Simply select the region of text you want to link to (must be a single line) in the referent target buffer, then switch to the source buffer where the link button should go and press {C-h h i l} and a link to the instance of the string you have selected will be inserted.

PR #492, opened 2024-03-19
Use numbers for line and column in links to org headers, add tests

matsl:
# What

Use numbers for implicit links to org headers.

# Why

~Links were created with pattern file#header:L{number}:C{number} which is not identified as a link? Is is a bug? I'm surprised it passed our testing in that case but it seems I did not add any tests for this functionality.  But now it is there.~ My bad. It is an optional feature but then we could use the more standard link without the L and C. Any reason not to do that?

I noticed this while looking at #491 to figure out if link to string match would be possible to use easily.

I also added tests to document the changed behavior. A link to a header now looks file:#header:{number}:{number}.

# Note

Is there any direct way now to create a link to string match? Was thinking if we could prefer that link type if the region is active in the file to link to? Or would that be to subtle?

### DIFF
--- a/hui.el
+++ b/hui.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    19-Sep-91 at 21:42:03
-;; Last-Mod:     21-Mar-24 at 15:30:24 by Bob Weiner
+;; Last-Mod:     23-Mar-24 at 23:52:49 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
--- a/hui.el
+++ b/hui.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    19-Sep-91 at 21:42:03
-;; Last-Mod:     23-Mar-24 at 23:52:49 by Mats Lidell
+;; Last-Mod:     23-Mar-24 at 23:53:35 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
--- a/hui.el
+++ b/hui.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    19-Sep-91 at 21:42:03
-;; Last-Mod:     23-Mar-24 at 23:52:49 by Mats Lidell
+;; Last-Mod:     21-Mar-24 at 15:30:24 by Bob Weiner
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
--- a/hui.el
+++ b/hui.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    19-Sep-91 at 21:42:03
-;; Last-Mod:     23-Mar-24 at 23:53:35 by Mats Lidell
+;; Last-Mod:     23-Mar-24 at 23:52:49 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2024-03-23  Mats Lidell  <matsl@gnu.org>
+
+* test/hui-tests.el (hui--ibut-link-directly-to-org-header-first-column)
+    (hui--ibut-link-directly-to-org-header-second-column)
+    (hui--ibut-link-directly-to-org-body): Add tests for ibut links to org
+    mode files.
+
 2024-03-21  Mats Lidell  <matsl@gnu.org>
 
 * .github/workflows/main.yml (jobs): Ensure grep for warnings does not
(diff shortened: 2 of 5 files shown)


rswgnu:
Yes, there must be a number following the L and C or the link is invalid.On Mar 19, 2024, at 4:14 PM, Mats Lidell ***@***.***> wrote:﻿
@matsl requested your review on: #492 Use numbers for line and column in links to org headers, add tests.

matsl:
> Yes, there must be a number following the L and C or the link is invalid.

Sorry if I was not clear enough. Seems the link that is currently created using the master branch is with an L and C preceding the number for the line and column. However that link is is not valid on my machine at least. But if the L and C is dropped and only the numbers are used the links are valid!? Hence the question if this might be a bug? 

This PR changes the code so that such a link, without L and C, is created. There are test cases that verifies that the new links looks right.

I see now that the tests could also be improved to also verify that the links works as they should. I will look into that.

rswgnu:
The L and C are optional and the link should work with them; otherwise, it is a bug we should fix.On Mar 20, 2024, at 6:45 AM, Mats Lidell ***@***.***> wrote:﻿

Yes, there must be a number following the L and C or the link is invalid.

Sorry if I was not clear enough. Seems the link that is currently created using the master branch is with an L and C preceding the number for the line and column. However that link is is not valid on my machine at least. But if the L and C is dropped and only the numbers are used the links are valid!? Hence the question if this might be a bug?
This PR changes the code so that such a link, without L and C, is created. There are test cases that verifies that the new links looks right.
I see now that the tests could also be improved to also verify that the links works as they should. I will look into that.

matsl:
> The L and C are optional and the link should work with them

Yes. On further testing today it seems to work. My bad. Sorry for the noise. Not sure why it broke yesterday. Probably bad manual testing on my part.

I can revert the removal of the optional letter (see comment below) and keep the tests since I believe they are an improvement.

But before that: Why do we add them here and not for the file paths links with line and column? Seems a bit add hoc to use it for the `string match` and not for the other path links? Should we not be consistent and use the same style everywhere for generated links?

matsl:
@rswgnu 

> I can revert the removal of the optional letter (see comment below) and keep the tests since I believe they are an improvement.

I have reverted the removal of L and C and updated the test consequently.

rswgnu:
> I have reverted the removal of L and C and updated the test consequently.

Further, we’ll want to unify the two link-directly types and have them both use the L and C.  I have added support for ‘I’ as an instance/occurrence number but have’t pushed that yet.  If you don’t get to this change today, I’ll probably make it tonight.

matsl:
> we’ll want to unify the two link-directly types and have them both use the L and C

The L and C was missing from `hpath:section-line-and-column-regexp`  which made `defib pathname-line-and-column` not identify pathnames with the L-C-format. 

This is in a sense a big change since, at least at a quick check, most docstrings related to the  pathname with line and column does not mention the optional L anc C characters. So before I go over that I just want to sync this change with you. If you think the change is good I'll complete the PR with the update to the docstrings and docs. (Alternatively we make that in to a separate PR?)

_There is something here that I have not fully got my head around this morning. I'm pretty sure the "string match" links using an L and C worked before although the link was not identified as a pathname. I tried to verify that running in a clean workspace but failed. Maybe I'm mixing up things so will pause this investigation in favor of your comments on the PR._

rswgnu:
As long as things work with both:  :L3, :3, :L3:C4, :3:4, then I am good with a fix.-- BobOn Mar 24, 2024, at 3:55 AM, Mats Lidell ***@***.***> wrote:﻿

we’ll want to unify the two link-directly types and have them both use the L and C

The L and C was missing from hpath:section-line-and-column-regexp  which made defib pathname-line-and-column not identify pathnames with the L-C-format.
This is in a sense a big change since, at least at a quick check, most docstrings related to the  pathname with line and column does not mention the optional L anc C characters. So before I go over that I just want to sync this change with you. If you think the change is good I'll complete the PR with the update to the docstrings and docs. (Alternatively we make that in to a separate PR?)
There is something here that I have not fully got my head around this morning. I'm pretty sure the "string match" links using an L and C worked before although the link was not identified as a pathname. I tried to verify that running in a clean workspace but failed. Maybe I'm mixing up things so will pause this investigation in favor of your comments on the PR.

matsl:
> As long as things work with both:  :L3, :3, :L3:C4, :3:4, then I am good with a fix.

Well. It passes all tests. :shrug:  So I assume it means you see no apparent problem with the implemented solution. :smile: 

My next step is then to go over the docstrings so that the optional L and C are mentioned appropriately. I'll do that as part of this PR.

hui.el
```diff
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    19-Sep-91 at 21:42:03
-;; Last-Mod:     21-Mar-24 at 15:30:24 by Bob Weiner
+;; Last-Mod:     23-Mar-24 at 23:53:35 by Mats Lidell
```
rswgnu, line 6: Maybe you could just eliminate this timestamp change.

rswgnu (approved):

hui.el
matsl: Puh. Good catch. It is eliminated. (It took a bunch of commits due to my mistake but if we squash these it should be OK.)

Does this mean you are good with this PR, and I can do the docstrings update in another, or just that you wanted this removed when you spotted it was the only changed in that file?

rswgnu:
Youcan do the doc updates in another PR.

PR #494, opened 2024-03-22
First attempt to support test coverage using the testcover package

matsl:
# What

Add new test file for producing test coverage data. Uses the testcover
package.

# Why

It is good to get some feedback how the tests are exercising the code
under test. This PR provides some convenience functions for running
code coverage interactively using the test suite or running it from
the make file.

# Note 

There seems to be some issues with stability of the tests for the
monitored file. Maybe because of the instrumentation. More experience
needs to be gathered here but it is a first attempt.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,6 +5,14 @@
 
 2024-03-31  Mats Lidell  <matsl@gnu.org>
 
+* Makefile (coverage): Make target to use the coverage function from the
+    command line.
+
+* test/hy-test-coverage.el: New test file for producing test coverage
+    data. Uses the testcover package.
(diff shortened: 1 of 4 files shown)


rswgnu (approved):

ChangeLog
rswgnu, line 1: Did we ever get any data out of this to see what our coverage looks like?

matsl: Yes and no. After implementing it, and looked at how the data looks, we have not used or tested it in practice on any file. It was so long ago, or my memory is so weak, I would have to look at it again to remember how it works.

PR #495, opened 2024-03-24
Add Emacs 29.3 to CI builds

matsl:
# What

Add Emacs 29.2 to versions used for the CI builds.

# Why

Noticed when 29.3 was released that we had not included the previous
minor release in the versions we use to CI builds. 

The docker image for 29.3 is not available yet so will add that later.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,8 +15,8 @@ jobs:
       fail-fast: false
       matrix:
         version: [27.2, 28.1, 28.2, 29.2, master]
-      exclude:
-        - version: 28.1
+        exclude:
+          - version: 28.1
     container: silex/emacs:${{ matrix.version }}-ci
 
     steps:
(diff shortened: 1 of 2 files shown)


rswgnu (approved):

matsl:
@rswgnu I think I understand now how to deal with the required build step `build (28.1)`. I think that is set in the branch protection rules and we need to remove that there.

matsl:
@rswgnu Since I wrote yesterday 29.3 silex docker image was added so I have bumped the version to 29.3. So if you can look into removing the `build (28.1)` branch protection rule so that the PR can be merged would be great.

rswgnu (approved):

matsl:
@rswgnu Did you save the removal of the required build-28.1?

rswgnu:
Missed that there was a save button at the bottom.   Just did it now.-- BobOn Mar 31, 2024, at 1:03 PM, Mats Lidell ***@***.***> wrote:﻿
@rswgnu Did you save the removal of the required build-28.1?

PR #497, opened 2024-03-29
Add install-melpa target for checking a melpa install

matsl:
# What

Add install-melpa target for checking a melpa install.

# Why

We want to be able to verify that the version of hyperbole packaged
by melpa can be installed with expected result. i.e. tests works and
Emacs can be started and Hyperbole is initialized properly.

# Note

When run on my machine there are a few test cases that fail when using `install-melpa` but none with `install-elpa` nor `install-elpa-devel`. I have not  checked why yet. Could be that we depend on some file for the tests that we did not include in the melpa recipe or something more serious. I believe this install test config is correct though.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2024-03-29  Mats Lidell  <matsl@gnu.org>
+
+* Makefile: Add install-melpa target. Used to validate hyperbole can be
+    installed from melpa.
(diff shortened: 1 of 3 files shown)


matsl:
@rswgnu I spotted the problem. Missing subfolders in recipe. Correction in this PR https://github.com/melpa/melpa/pull/8975

rswgnu (approved):

PR #500, opened 2024-03-30
Verify hyrolo match count

matsl:
# What

Verify hyrolo match count.

# Why

Hyrolo match regexps under different options. This is a start
verifying the count result using either match all or just match
headlines. _Both types of matches looks like they fail._ 

The first search for all matches returns 2 but in *HyRolo* all
three lines containing "match" are highlighted. The second
search for just searching in headlines does not find any match.

Sharing for information. I tried to debug this today but with no
success.

Marked as expected to fail for not breaking the CI.

# Note

Can I have I misunderstood what headline means in this 
context? I'm assuming that in org mode it is the first line 
/following a star.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2024-03-31  Mats Lidell  <matsl@gnu.org>
+
+* test/hyrolo-tests.el (hyrolo-test--grep-count): Verify hyrolo match count.
+
 2024-03-31  Bob Weiner  <rsw@gnu.org>
 
 * hyrolo.el (hyrolo-grep-file): Remove this line that was skipping past the
--- a/test/hyrolo-tests.el
+++ b/test/hyrolo-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    19-Jun-21 at 22:42:00
-;; Last-Mod:     30-Mar-24 at 23:35:15 by Mats Lidell
+;; Last-Mod:     31-Mar-24 at 11:36:33 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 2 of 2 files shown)


rswgnu:
The HyRolo count values are how many records matched the search pattern, not how many matches exist within the records, i.e. records can have more than one match to the pattern.  So the correct value in your test is 2.  That is what you should get.

matsl:
> The HyRolo count values are how many records matched the search pattern, not how many matches exist within the records, i.e. records can have more than one match to the pattern. So the correct value in your test is 2. That is what you should get.

Aha. I see. A third option on how it should be interpreted. :smile: That is good news. Just a misunderstanding on my part then. I'll update the test case and see what more tests I can construct.

What do you think of making the docs clearer on this part so even I can understand? :laughing:

matsl:
@rswgnu 

>  I'll update the test case and see what more tests I can construct.

The tests are now updated and I have added some more verification. Count entries with only headerline match, count entries with no headerline match and no match at all. I can't get the numbers to line up so the test case is still marked as an expected failure. 

Can it be further misunderstand on my part of what different values on the parameters MAX-MATCHES and HEADLINE-ONLY really means?

rswgnu:
Well, I did clarify the docs with my pushes last night.  We can have a look today.-- BobOn Mar 31, 2024, at 6:29 AM, Mats Lidell ***@***.***> wrote:﻿
@rswgnu

I'll update the test case and see what more tests I can construct.

The tests are now updated and I have added some more verification. Count entries with only headerline match, count entries with no headerline match and no match at all. I can't get the numbers to line up so the test case is still marked as an expected failure.
Can it be further misunderstand on my part of what different values on the parameters MAX-MATCHES and HEADLINE-ONLY really means?

matsl:
> Well, I did clarify the docs with my pushes last night. 

:smile: That is like hitting a running target. To hard for me!

rswgnu (approved):

PR #507, opened 2024-04-01
Use kotl-mode.el instead of version.texi which is not exported.

matsl:
# What

Use kotl-mode.el instead of version.texi which is not exported.

# Why

Not all files are exported to melpa package in the recipe and
version.texi is not. When fixing this same issue for the file COPYING
recently I missed there were more of the same problem.  

All files used in tests should now be picked from files available
through all different install packages.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,6 +25,8 @@
 * test/hpath-tests.el (hpath:prepend-shell-directory-test):
 * test/demo-tests.el (demo-manifest-test): Replace COPYING with DEMO since
     it is not exported in all installations, i.e. MELPA.
+    (hpath:prepend-shell-directory-test): Use kotl-mode.el instead of
+    version.texi which is not exported.
 
 2024-04-01  Bob Weiner  <rsw@gnu.org>
 
--- a/test/hpath-tests.el
+++ b/test/hpath-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    28-Feb-21 at 23:26:00
-;; Last-Mod:      7-Apr-24 at 10:40:03 by Bob Weiner
+;; Last-Mod:      8-Apr-24 at 16:45:22 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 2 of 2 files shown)


rswgnu:
We need to export version.texi in the melpa recipe.  That is an oversight.-- BobOn Apr 1, 2024, at 4:23 PM, Mats Lidell ***@***.***> wrote:﻿
@matsl requested your review on: #507 Use kotl-mode.el instead of version.texi, it is not exported..

matsl:
> We need to export version.texi in the melpa recipe

OK. I'll update the recipe and scrap this PR then.

matsl:
@rswgnu 

> > We need to export version.texi in the melpa recipe
> 
> OK. I'll update the recipe and scrap this PR then.

When trying to fix this I noticed that the recipe [does include the texi files](https://github.com/melpa/melpa/blob/a4473b759096dc42433ff0f0e69f3448cc04915b/recipes/hyperbole#L10) but there is [built in machinery in melpa](https://github.com/melpa/melpa/commit/c5157b46ac35aae55a89688f6b3e729366df5a94) to process the texi files in order to produce info pages. (And as a consequence of that the texi files are not included in the resulting package!? Easy checked by looking at the package tar file.) 

This is a service to packages that include a texi file and melpa will create the info and dir file so it will show up in the info pages in Emacs. As of this writing I'm not sure if here is an option to control this or if it is hardcoded. I suspect that melpa regenerated the hyperbole info file.

Is there any other reason for us to include version.texi in the package than to let users have the complete set of sources? (And, as a side effect, for running the tests that uses the version.texi file?)

rswgnu:
Hi @Lidell Mats ***@***.***>, have you updated the recipe?  Thanks.

matsl:
Hi Bob, @rswgnu, look at my previous reply (what I noticed after I said I would update the recipe) In short: The recipe already contains our texi-files. Melpa turnes these into info files that are included `instead` of the texi. I think Melpa is doing the right thing and we are trying to force it to be different. We should not. Let's discuss this in our meeting where I can elaborate.

rswgnu:
It is part of our source doc and someone might want to provide us a patch to include.  There must be a way in the recipe to force inclusion of specific files and if not, let’s hear from the melpa maintainers.  If .el files are included, so should .texi be.-- BobOn Apr 2, 2024, at 5:10 AM, Mats Lidell ***@***.***> wrote:﻿
@rswgnu


We need to export version.texi in the melpa recipe

OK. I'll update the recipe and scrap this PR then.

When trying to fix this I noticed that the recipe does include the texi files but there is built in machinery in melpa to process the texi files in order to produce info pages. (And as a consequence of that the texi files are not included in the resulting package!? Easy checked by looking at the package tar file.)
This is a service to packages that include a texi file and melpa will create the info and dir file so it will show up in the info pages in Emacs. As of this writing I'm not sure if here is an option to control this or if it is hardcoded. I suspect that melpa regenerated the hyperbole info file.
Is there any other reason for us to include version.texi in the package than to let users have the complete set of sources? (And, as a side effect, for running the tests that uses the version.texi file?)

rswgnu (approved):

PR #510, opened 2024-04-07
Use DEMO in test since COPYING is not exported in melpa

matsl:
# What

Bring back DEMO in test to replace COPYING.

# Why

COPYING is not exported in the Melpa recipe so should not be used in
tests. Not sure how this was reverted but here is a PR anyway to bring it back to using DEMO
in tests and not COPYING since that is not exported everywhere i.e. Melpa.


rswgnu (approved):

matsl:
Fixed in #507

PR #512, opened 2024-04-08
Tests for org-links outside of org-mode

matsl:
# What

Tests for org-links outside of org-mode.

# Why

There was a bug here where `hsys-org-enable-smart-keys`  would
interfere with the org-links which is should not. It is only
applicable within `org-mode`. A fix for that has already been pushed.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2024-04-08  Mats Lidell  <matsl@gnu.org>
+
+* test/hsys-org-tests.el (hsys-org--org-outside-org-mode-tmp-buffer)
+    (hsys-org--org-outside-org-mode-tmp-file): Tests for org-links outside
+    of org-mode.
+
 2024-04-07  Bob Weiner  <rsw@gnu.org>
 
 * hyrolo.el (hyrolo-consult-grep-paths): Fix handling of zero 'max-matches'
--- a/test/hsys-org-tests.el
+++ b/test/hsys-org-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    23-Apr-21 at 20:55:00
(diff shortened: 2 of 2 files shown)


rswgnu (changes requested):
Look at the values of hsys-org-enable-smart-keys, need to test those to ensure nothing gets written that is conditionalized on that and stops these links from working.

matsl:
> Look at the values of hsys-org-enable-smart-keys, need to test those to ensure nothing gets written that is conditionalized on that and stops these links from working.

Are you saying that we need to run the test with all four values (unset buttons t nil)?

rswgnu:
Yes.-- BobOn Apr 8, 2024, at 4:49 PM, Mats Lidell ***@***.***> wrote:﻿

Look at the values of hsys-org-enable-smart-keys, need to test those to ensure nothing gets written that is conditionalized on that and stops these links from working.

Are you saying that we need to run the test with all four values (unset buttons t nil)?

matsl:
> Yes.

Easy to fix. Update tomorrow.

matsl:
> Easy to fix. Update tomorrow.

Looping over the possible values, repeating the test. PTAL.

rswgnu (changes requested):
How are you trsting that the action key actually opens the link properly if that is the function you are mocking?

you are not trying to test that the variable is set right but to see if setting it affects recognition and execution of the button, i.e. runs the right action.

matsl:
> How are you trsting that the action key actually opens the link properly if that is the function you are mocking?
> 
> you are not trying to test that the variable is set right but to see if setting it affects recognition and execution of the button, i.e. runs the right action.

These are unit test that verifies `org-open-at-point-global` is called as a result of using the action-key (which it was not before your fix.) If the mocked function is not called the test fails. It does not attempt to verify that `org-open-at-point-global` does the right thing since that is an org function.

rswgnu (approved):

rswgnu:
Approved.

PR #522, opened 2024-04-23
fix pathnames in ls R listings

matsl:
# What

- Add test case for delimited paths in "ls -R"-listings
- Allow for space between file names in ls -R listings

# Why

Plain file names in ls -R have stopped to work. I suspect it is
because the file names are not found due to that tab separation is
assumed. I don't know what has changed but in my shell-mode ls -R only
use space between the files.

Removing the special treatment for shell-mode resolves the issue
partially and it is what this PR does. For folders that does not
contain files with names having whitespace this fix works.

# Note 

As pointed out above there are other issues with ls -R listings. If
there are files with space in their name the quotes produces by ls
causes problems since hpath:delimited-possible-path can't handle
that. 

The quotes does further harm in that when clicking on other files the
quote are found by the regexps and we get a list of files instead of
just one file. 

The last problematic case is that the ls listing gets indented one
space because of quotes in file names and the current regexp does not
handle that gracefully. (That might be possible to solve locally but
to much for me today. Might give it a try but since it is part of the
quote problem just fixing that might not make sense.)

## Reflection

Can it be that this and the related functions have to much
responsibilities? Looking for patterns only seen in info, tex, html
while being in shell-mode seems to be redundant. Can we not make the
task easier by earlier realizing that we are in shell-mode and maybe
use adjusted pathname functions that knows about that!?

Anyway, something to discuss. 

I'm submitting this as documentation and as a potential fix for the
case with folders that only contains filenames with no whitespaces. So
a step in the right direction.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,8 @@
 2024-04-23  Mats Lidell  <matsl@gnu.org>
 
+* hpath.el (hpath:delimited-possible-path): Allow space between file names
+    in "ls -R"-listings.
+
 * test/hpath-tests.el (hpath--hpath:delimited-possible-path): Add test
     case for delimited paths in "ls -R"-listings.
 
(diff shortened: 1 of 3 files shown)


matsl:
This was work in progress that has been superseded by work on master branch, See test cases in #536. Closing the PR.

PR #523, opened 2024-04-26
Set pipefail so that makes error code is propagated to the shell

matsl:
# What

Current pipe with tee swallows the error code if any from make.

# Why

The pipe to tee for both showing the output from make and storing it
for processing swallows the error code from make if it fails. Setting shell 
to bash explicitly gets the `pipefail` option set which makes the 
pipe a whole fail if make fails.

# Note

I noticed this incorrectness when looking at why the `rsw` branch with 
the hywiki did not fail since it should. (But it actually did fail. It was not
noticed since the compilation error was swallowed. This PR fixes this.) 

There are some test commits before I got it right but I will squash them 
on merge.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,6 +24,7 @@ jobs:
         run: make clean
 
       - name: Compile
+        shell: bash
         run: |
           make binn 2>&1 | tee MAKE_BIN_OUTPUT
           test 0 -eq ${PIPESTATUS}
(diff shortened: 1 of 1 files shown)


rswgnu (changes requested):
Isn’t ‘sh’ the same as ‘bash’ on Linux?  If so, the former will work on more platforms that do not have bash.

matsl:
> Isn’t ‘sh’ the same as ‘bash’ on Linux? If so, the former will work on more platforms that do not have bash.

No, `sh` is posix compliant whereas `bash` contains enhancements such as `pipefail` that we want to activate in this case. This also only affects the github workflow so platform compatibility is of no concern.

rswgnu (approved):

PR #524, opened 2024-05-05
Quote lambdas used in hact calls

matsl:
# What

Quote lambdas used in hact calls.

# Why

The lambda was not passed as a function to hact. Noticed with recent
versions of Emacs 30.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2024-05-05  Mats Lidell  <matsl@gnu.org>
+
+* test/hui-tests.el: Quote lambdas used in hact calls.
+
 2024-04-20  Bob Weiner  <rsw@gnu.org>
 
 * hui-select.el (hui-c++-defun-prompt-regexp): Add to eliminate an Emacs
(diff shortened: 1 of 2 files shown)


rswgnu (approved):

matsl:
> approved these changes 

Thanks. Maybe I'll look a bit deeper at it first unless you think it is a show stopper now and we should use it as  work around?

rswgnu:
Sure.

rswgnu:
I pushed an update to handle Emacs 30 closures to the rsw branch.

See if you can fix the failing tests here:

https://github.com/rswgnu/hyperbole/actions/runs/9056493310/job/24879104466?pr=521

matsl:
> I pushed an update to handle Emacs 30 closures to the rsw branch. See if you can fix the failing tests here: https://github.com/rswgnu/hyperbole/actions/runs/9056493310/job/24879104466?pr=521


Do you think the test cases needs an update? If so why? They work fine for all other versions and have done that for some time.
Or are you saying that you want me to look into what to **add to your update** so that the test cases pass?

I notice a change in the error though. The test cases now fail due to this (or similar depending on the test case)

>   (wrong-number-of-arguments
     #[nil ((call-interactively 'hui:ibut-rename)) 

It could be due to another Emacs 30 incompatibility change of course. Need to investigate that further to see where it occurs.

PR #525, opened 2024-05-12
Add quotes so make is executed in the container

matsl:
# What

Add quotes so make is executed in the container.

# Why

The point of the target is to run make using the emacs provide by the
docker image. So make must be run within the target.


matsl:
Realizing we have a slight problem here with master version blocking the build due to the lambda-closure issue.  We should probably focus on fixing that rather than working around the issue!?

rswgnu (approved):

PR #534, opened 2024-06-01
Add test cases for hywiki

matsl:
# What

Add tests for hywiki.

# Why

Tests are good! ;-)

# Note

Adding this since we have hywiki in master and this is how far I have
come with the tests. 

I have had problems with the display related test i.e. verifying that
faces are set properly. Don't know yet if that can be tested in batch
mode or if the redisplay is needed. The current two test cases for
faces uses a "hack" to get events processed and needs to be executed
interactively. i.e. the "test-all" target.

There is also one expected failed test. Not sure what is the expected
behavior. Should `hywiki-add-page` with an all lowercase argument
work?

### DIFF
--- a/test/hywiki-tests.el
+++ b/test/hywiki-tests.el
@@ -1,0 +1,138 @@
+;;; hywiki-tests.el --- one line summary                -*- lexical-binding: t; -*-
+;;
+;; Author:       Mats Lidell
+;;
+;; Orig-Date:    18-May-24 at 23:59:48
+;; Last-Mod:      1-Jun-24 at 16:16:00 by Mats Lidell
+;;
+;; SPDX-License-Identifier: GPL-3.0-or-later
+;;
+;; Copyright (C) 2021-2022  Free Software Foundation, Inc.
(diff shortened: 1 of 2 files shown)


test/hywiki-tests.el
```diff
@@ -0,0 +1,138 @@
...
+  "Verify add page requires a WikiWord."
+  :expected-result :failed
+  ;; Should not leave erroneously created file after test but leaving
+  ;; added error cleanup till later if it is even needed!? No file
+  ;; should be created so only happens on error!? (If this is
+  ;; considered an error case that is.)
+  (let ((hywiki-directory (make-temp-file "hywiki" t))
+        (hywiki--pages-hasht nil))
+    (unwind-protect
+        (should-error (hywiki-add-page "notawikiword"))
```
matsl, line 52: @rswgnu Should this call work or err?

rswgnu: We should make it fail.

rswgnu (approved):

PR #548, opened 2024-06-23
Add support for org face types to support org 9.7 and older

matsl:
# What

Add support for org face types to support org 9.7 and older.

# Why

Org 9.7 changes the use of face types in an not compatible way having
multiple properties where we in the past only had one. This PR tries
to adopt for that.

# Note

Filing the PR for feedback at the end of my working day! PTAL

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2024-06-23  Mats Lidell  <matsl@gnu.org>
+
+* hsys-org.el (hsys-org-first-face-type-pos)
+    (hsys-org-last-face-type-pos, hsys-org-region-with-face-type)
+    (hsys-org-is-face-type): Add support for working with org face types
+    to be compatible with org mode 9.7 and older versions.
+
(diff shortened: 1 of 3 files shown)


rswgnu (commented):
Will want a few changes here.  I will likely send a PR against this branch tonight for you to look at.  Finish any other tests you want on this code and I’ll ensure they continue to work.

matsl:
> Will want a few changes here. I will likely send a PR against this branch tonight for you to look at. Finish any other tests you want on this code and I’ll ensure they continue to work.

Great. This is how far I got yesterday. Looking forward to your comments. 

I think there are some edge cases that might not be properly handled that we could add tests for. If possible, point put those and we can write tests for them to ensure it will continue to work over time. I think though current tests could be covering most practical cases.

rswgnu (changes requested):
This has been superseded by PR #550, so it should be closed.

PR #549, opened 2024-06-24
Use basename docker, depend on update to ensure latest image

matsl:
# What

Rename targets from dockerized to the simpler docker. Add dependency
on docker-update to ensure latest docker image is used.

# Why

Dockerized is to long and manually pulling the image when it has been
updated is easily forgotten so better check for it each run. If the
image is already the latest the check is quick.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Author:       Bob Weiner
 #
 # Orig-Date:    15-Jun-94 at 03:42:38
-# Last-Mod:     23-Jun-24 at 00:16:22 by Bob Weiner
+# Last-Mod:     24-Jun-24 at 23:49:29 by Mats Lidell
 #
 # Copyright (C) 1994-2023  Free Software Foundation, Inc.
 # See the file HY-COPY for license information.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -77,14 +77,14 @@
...
-#                   make dockerized-run version=27.1   - run Emacs V27.1
+#               To interactively run a docker version of Emacs with Hyperbole:
+#                   make docker-run              - default to running master
+#                   make docker-run version=27.1 - run Emacs V27.1
 #
 #               To build and test a dockerized version of Emacs with Hyperbole:
-#                   make dockerized                    - defaults: version=master targets='clean bin test'
+#                   make docker                  - defaults: version=master targets='clean bin test'
 #
-#                   make dockerized version=28.2 targets='clean bin' - Emacs 28.2 with Hyperbole byte-compiled (possibly natively compiled)
```
matsl, line 87: @rswgnu The docker image we use now does not support native compilation so I remove the comment about native compilation. We can add that if and when it gets available.

rswgnu (approved):

PR #560, opened 2024-07-09
Use 29.4 for local docker builds

matsl:
# What

Use 29.4 for local docker builds.

# Why

The DOCKER_VERSIONS macro should be in sync with the versions used in
the CI builds.


matsl:
@rswgnu Just a reminder about this one. I should have updated the `DOCKER_VERSIONS` macro when bumping the workflow version but I forgot. This brings the `docker_all_tests` and `docker_batch_tests` target in sync with the workflow versions.

rswgnu (approved):

rswgnu:
Approved. -- Bob

PR #563, opened 2024-07-29
Fix window/minibuffer latency issue

ve-nt:
This (sit-for) causes an annoying bit of latency every time a new window or minibuffer is opened. The extra delay makes Emacs feel very clunky, and almost put me off using Hyperbole more. I'm unsure what the consequences are to removing this, but I would like to see a better solution to any bugs this sleep may be hiding.

### DIFF
--- a/hywiki.el
+++ b/hywiki.el
@@ -832,9 +832,6 @@ the current page unless they have sections attached."
   (walk-windows
    (lambda (window)
      (with-selected-window window
-       ;; Display buffer before `normal-mode' triggers possibly
-       ;; long-running font-locking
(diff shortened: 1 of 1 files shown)


rswgnu:
This is fixed in the latest Github pre-release (master branch) with a (sit-for 0) call.  Give it a try and thanks for pointing it out.  Closing this issue and rejecting the PR since it is already fixed.

rswgnu:
This is fixed in the latest Github pre-release (master branch) with a (sit-for 0) call.  Give it a try and thanks for pointing it out.  Closing this issue and rejecting the PR since it is already fixed.

PR #564, opened 2024-07-30
Add file exist check in ripgrep-msg and a test case

matsl:
# What

Add a check if file in ripgrep output exists before concluding it is a
valid ripgrep-msg.

# Why

A number on a line can get interpreted as ripgrep line number which it
is not. That will block other buts from having their go at it. By
adding a check that the matching "file" exists makes it some what more
robust although I think there can still be problematic cases.

# Note

The error was found by having a iso date, like 2024-07-30, first on a
line. The string on the line above was interpreted as a file name. The
resulting action was a no-op but the action that should have been
triggered was not given its chance.

A test case triggering the bug for an unmodified ripgrep-msg is
supplied.

What do you think of requiring that the file listed in the ripgrep
output exists for it to be a valid button?

### DIFF
--- a/hibtypes.el
+++ b/hibtypes.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    19-Sep-91 at 20:45:31
-;; Last-Mod:     23-Aug-24 at 21:28:35 by Mats Lidell
+;; Last-Mod:     23-Aug-24 at 21:32:09 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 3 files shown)


rswgnu:
Let’s discuss the pros and cons on a call and see where we land.  I would rather more narrowly exclude lines as needed, but am open to discussion.

matsl:
> I would rather more narrowly exclude lines as needed, but am open to discussion.

You need to elaborate so I understand what you are thinking of here. Lets talk on Sunday.

hibtypes.el
```diff
@@ -995,7 +995,8 @@ than a helm completion buffer)."
                       (looking-at "[1-9][0-9]*[-:]\\|--$")))
           (unless (or (looking-at "[1-9][0-9]*[-:]\\|--$")
                       (and (setq file (buffer-substring-no-properties (line-beginning-position) (line-end-position)))
```
rswgnu, line 997: Use this instead:

                      (and (setq file (string-trim (buffer-substring-no-properties (line-beginning-position) (match-beginning 0))))
                           (or (string-empty-p file)
			       (not (file-exists-p file)))))

rswgnu (changes requested):

rswgnu (approved):

PR #566, opened 2024-08-04
Add elisp-compiler-msg-test

matsl:
# What

Add test elisp-compiler-msg-test.

# Discussion

A test is still missing for the byte compiler section of the ibut. I'm not sure the patterns for that in the code is correct. I can't find any byte compile example where these patterns are seen. Neither using batch byte compiler from the make targets nor using the byte compiler functions!? Do you have an example?

Regarding the native byte compiler section it seems that the function reported in the "In function"-section does not have to be part of the file that is byte compiled!? Se example here:

```
Compiling /home/matsl/src/hyperbole/hsys-org.el...
Loading /home/matsl/src/hyperbole/hyperbole-autoloads.el (source)...

In end of data:
hywiki.el:490:25: Warning: the function ‘hsys-org-at-tags-p’ is not known to be defined.

In hyrolo-tags-view:
hyrolo.el:1409:11: Warning: Unused lexical variable ‘org-agenda-buffer-tmp-name’
hyrolo.el:3519:20: Warning: Use keywords rather than deprecated positional arguments to ‘define-minor-mode’
...
```

The file that is compiled is hsys-org.el but the end of data message refers to hywiki.el and `hyrolo-tags-view` is part of hyrolo.el. So it seems this pattern can cause a lot of false positives. In the sunshine case where (verified in the test case) the file referred to in the Compiling line is the file for which the message is for. That works. But in those cases the column and line info on the subsequent lines would be a more direct link to the problem!?

I have tried to look for a spec on how the native compile messages are constructed but so far I have not found anything. Maybe best to ask for that.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2024-08-08  Mats Lidell  <matsl@gnu.org>
+
+* test/hibtypes-tests.el (elisp-compiler-msg-test): Add test.
+
 2024-08-06  Mats Lidell  <matsl@gnu.org>
 
 * test/hproperty-tests.el: Test file for hproperty.
--- a/test/hibtypes-tests.el
+++ b/test/hibtypes-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    20-Feb-21 at 23:45:00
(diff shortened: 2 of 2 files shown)


matsl:
Mocking seems to cause problems for batch run so will close this and look for other ways to do this (if I can't debug the darn thing and understand what is going on. ;-)

rswgnu:
Sounds good.

rswgnu (approved):

PR #576, opened 2024-08-16
Add tree-sitter "programming modes" supported by Hyperbole

matsl:
# What

Add tree-sitter based modes for the programming modes currently
supported by Hyperbole.

# Why

Hyperbole uses buffers current major mode for checking if some
functionality shall be available in the buffer. With tree-sitter comes
modes for many programming languages and those can be used instead of
the old programming modes. These new modes are not recognized by
Hyperbole and if they are used the programming language specific
buttons are not recognized.

# Note

*Was looking into tree-sitter and to use the c++-ts-mode in a c++
file only to realize that Hyperbole smart-c++ was not available in
that case because we don't understand what c++-ts-mode is.*

This just blindly adds the tree-sitter modes to the program language
checks. There are no automatic tests that this works for all cases. On
the other hand we have no tests that it works for all the old modes
either.

It seems some old modes and tree sitter modes use the same base
mode. It could be possible to use that in Hyperbole to check for the
base mode instead to limit the modes to check for. This is not a
general pattern though so can only be applied after checking for each
programming language.

No analysis have been done if there are tree-sitter features that
Hyperbole utilized. This PR only allows the programming language to
derived by looking at the tree-sitter modes.

### DIFF
--- a/hui-mouse.el
+++ b/hui-mouse.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    04-Feb-89
-;; Last-Mod:     12-Jul-24 at 20:48:37 by Mats Lidell
+;; Last-Mod:     16-Aug-24 at 20:59:19 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -444,7 +444,7 @@ Its default value is `smart-scroll-down'.  To disable it, set it to
     ;;
     ;; Python files - ensure this comes before Imenu for more advanced
     ;; definition lookups
-    ((and (or (and (derived-mode-p 'python-mode) buffer-file-name)
+    ((and (or (and (derived-mode-p '(python-mode python-ts-mode)) buffer-file-name)
 	      (and (featurep 'hsys-org) (hsys-org-mode-p)
 		   (equal (hsys-org-get-value :language) "python"))
 	      (let ((case-fold-search))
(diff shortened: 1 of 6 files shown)


hui-mouse.el
```diff
@@ -488,7 +488,7 @@ Its default value is `smart-scroll-down'.  To disable it, set it to
 		     (looking-at "@see[ \t]+")))))
      . ((smart-java) . (smart-java nil 'next-tag)))
     ;;
-    ((and (memq major-mode '(js2-mode js-mode js3-mode javascript-mode html-mode web-mode))
+    ((and (memq major-mode '(js2-mode js-mode js3-mode javascript-mode html-mode web-mode js-ts-mode))
```
rswgnu, line 491: Keep these mode names in alphabetical order.

hui-select.el
```diff
@@ -119,7 +119,8 @@
 ;;; ************************************************************************
 
 (defcustom hui-select-brace-modes
-  '(c++-mode c-mode java-mode objc-mode perl-mode tcl-mode)
+  '(c++-mode c-mode java-mode objc-mode perl-mode tcl-mode
+             c++-ts-mode c-ts-mode java-ts-mode)
```
rswgnu, line 123: Same comment as above.  Check the whole PR for this ordering issue.

rswgnu (changes requested):
Fix the comments and then it will be good.  -- rsw

hui-select.el
matsl: > Same comment as above. Check the whole PR for this ordering issue.

The mode lists should now be sorted.

rswgnu (approved):

PR #590, opened 2024-09-11
Verify hyrolo--expanded-file-list is updated when a file is added

matsl:
# What

Verify `hyrolo--expanded-file-list` is updated when a file is added.

# Why

Test marked as expected-result failed since cache currently is not
updated which seems like an error. Or what do you say?

# Note

Noticed this when I was adding a file to a directory in my
`hyrolo-file-list` and it was not found when I searched for things in
it a moment later. Surprising effect.

Caching is hard, and this is one of the subtle user problems it can
cause. Currently the cache is updated when `hyrolo-file-list` is
changed but the cache is on the file level so adding or removing a
file is not noticed.

Maybe there is a hook that can be used to invalidate the cache when a
file is created within folders in the `hyrolo-file-list`? (But feels a
bit clumsy.)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2024-09-11  Mats Lidell  <matsl@gnu.org>
+
+* test/hyrolo-tests.el (hyrolo-tests--get-file-list): Verify
+    `hyrolo--expanded-file-list' is updated when a file is added.
+
 2024-09-10  Bob Weiner  <rsw@gnu.org>
 
 * hywiki.el (hywiki-org-make-publish-project-alist): Add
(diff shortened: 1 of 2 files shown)


rswgnu:
What are you suggesting should triggercthe cache invalidation?  Any file creation within any directory in the list?

matsl:
> What are you suggesting should triggercthe cache invalidation? Any file creation within any directory in the list?

Yes. The system should work with the cache as if the file list was computed with each search. The cache is an optimization and should not conflict with expected functionality,

rswgnu (approved):

PR #592, opened 2024-10-03
add test for add and remove wiki page

matsl:
# What

- Add hywiki-tests--get-page-list-after-add-and-delete
- Set mod time and checksum on file creation

# Why

Did an attempt to write a test case that verifies that the pages
returned by hywiki-get-page-list when adding and removing pages. I
know removing might be something we are working on but it felt that
the caching is not perfect. So the test case might not be worth adding
in its present state but can be worth discussing. So filing it as PR
for review and discussion.


rswgnu (changes requested):
Obsoleted by latest push.  Please close.

PR #597, opened 2024-10-27
fix html contents

matsl:
# What

- Add top node for contents generation
- Use updated texinfo-7.css, replaces texinfo-klare.css
- Check in updated files stored under version control
- Convenience targets for working with doc files

# Why

The contents was missing.

# Note

I preferred to display the top node before the contents since
otherwise the top node is hard to see. I think that intro is useful
for new users with a picture and all. WDYT?

I also updated the css file since it was updated by the author and I
added a long wanted target that allows us to force regeneration of the
docs which is useful at times.

You can preview the result [here](https://htmlpreview.github.io/?https://raw.githubusercontent.com/rswgnu/hyperbole/refs/heads/fix-html-contents/man/hyperbole.html)


man/hyperbole.texi
```diff
@@ -141,7 +141,8 @@ Texinfo markup language.
 @c @setchapternewpage odd
 
 @node Top, Introduction, (dir), (dir)
-@unnumbered GNU Hyperbole
+@top GNU Hyperbole
```
rswgnu, line 144: Interesting @top must be a fairly new tag.  In the past, I think the @node Top would take care of this.  Great you found that.

rswgnu (approved):

PR #600, opened 2024-11-02
add more hywiki tests

matsl:
# What

- Add hywiki-word-is-p test
- ~Use plain star wildcard and add a test~
- Add test case for verifying property is updated on edit
- Add macros and functions for handling hooks functions to emulate redisplay behavior.

# Why

Tests are good.

# Note 

@rswgnu For just picking a file from a PR and run it locally you can click on `Files changed` / View file / Download.

### DIFF
--- a/hywiki.el
+++ b/hywiki.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    21-Apr-24 at 22:41:13
-;; Last-Mod:     28-Oct-24 at 01:43:34 by Bob Weiner
+;; Last-Mod:      1-Nov-24 at 00:00:19 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -836,7 +836,7 @@ these are handled by the Org mode link handler."
 (defun hywiki-directory-edit ()
   "Display and edit HyWiki pages in current `hywiki-directory'."
   (interactive)
-  (dired (concat hywiki-directory "[[:upper:]][[:alpha:]]*"
+  (dired (concat hywiki-directory "*"
 		 (regexp-quote hywiki-file-suffix))))
 
 (defun hywiki-directory-get-checksum ()
--- a/hywiki.el
+++ b/hywiki.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    21-Apr-24 at 22:41:13
-;; Last-Mod:      1-Nov-24 at 00:00:19 by Mats Lidell
+;; Last-Mod:     28-Oct-24 at 01:43:34 by Bob Weiner
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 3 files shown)


hywiki.el
```diff
@@ -836,7 +836,7 @@ these are handled by the Org mode link handler."
 (defun hywiki-directory-edit ()
   "Display and edit HyWiki pages in current `hywiki-directory'."
   (interactive)
-  (dired (concat hywiki-directory "[[:upper:]][[:alpha:]]*"
+  (dired (concat hywiki-directory "*"
```
matsl, line 839: @rswgnu  Character classes are not supported by other shells than bash.

rswgnu: Bummer.  Will have to use English chars then, * is insufficient here.

matsl: Why? Will it not list all possible files plus files not being WikiWord files?

test/hywiki-tests.el
```diff
@@ -283,6 +301,31 @@ Both mod-time and checksum must be changed for a test to return true."
...
+         (wikipage2 (hywiki-add-page "AnyWord")))
+    (unwind-protect
+        (progn
+          (find-file wikipage2)
+          (hywiki-mode 1)
+          (insert "Wikiord")
+	  (newline nil t)
+          (goto-char 5)
+          (should (looking-at-p "ord"))
+          (should-not (hproperty:but-get (point) 'face hywiki-word-face))
```
rswgnu, line 319: Have to have done some redisplay before this line, I think.

rswgnu (changes requested):

test/hywiki-tests.el
matsl: An alternative would be to more directly test the code that does the highlighting to stay away from depending on redisplay. But that seems not be that straight forward.

hywiki.el
rswgnu: Just HyWiki pages, as the doc indicates.

test/hywiki-tests.el
rswgnu: You have to get the hooks set by HyWiki to run before you check the results but since your test is running during the command, this doesn't happen.  I think we make a function that invokes the appropriate hooks and then just use that before the should and should-not calls.

hywiki.el
matsl: > Just HyWiki pages, as the doc indicates.
Maybe I misunderstand. I'm saying that using the shell pattern '*' will match all files in the hywiki-directory. So more than HyWiki-pages but is portable. Users will not likely have other files in that directory so is a small price to pay for supporting different shells.

test/hywiki-tests.el
matsl: That is an alternative and seems to work fine.

matsl: Added calls to `hywiki-maybe-highlight-page-names` before checking the face plus added a test case for deleting the first char in the WikiWord. That test case oddly works but running it manually does not.

So maybe I need to add calls to all hook functions as you suggest so behavior is consistent with interactive use.

hywiki.el
matsl: I have removed this test case and will update it when the modified version gets merged.

matsl:
@rswgnu This seems to run more inline with what I see when I run it manually in interactive mode. All tests succeed except the one with removing the leading capital character. I also had to mark some of the test cases as only valid when interactive as well. PTAL.

rswgnu (approved):

PR #605, opened 2024-11-12
remove warnings

matsl:
# What

- Remove unused warning, comment out all occurences of new-buf-flag
- Add defvar and declare-function to silence warnings

# Why

We should minimize the number of warnings.

# Note

I'm keeping the new-buf-flag but commenting it put everywhere to be in
company with the comment later in the code. I would opt for removing
all of these comments.

### DIFF
--- a/hui-treemacs.el
+++ b/hui-treemacs.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    19-Nov-17
-;; Last-Mod:     10-Nov-24 at 17:27:30 by Bob Weiner
+;; Last-Mod:     12-Nov-24 at 20:17:24 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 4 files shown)


hyrolo.el
```diff
@@ -1973,7 +1973,8 @@ Return number of matching entries found."
   ;; Save pattern as last rolo search expression.
   (setq hyrolo-match-regexp pattern)
   ;;
-  (let ((new-buf-flag) (actual-buf)
+  (let (;;(new-buf-flag)
```
matsl, line 1976: I propose we remove this altogether with the comment on line 2002. But the suggested change is a middle ground if the comment should stay.

hui-treemacs.el
```diff
@@ -45,6 +45,8 @@
 (declare-function treemacs-node-buffer-and-position "ext:treemacs-mouse-interface")
 (declare-function treemacs-quit "ext:treemacs-interface")
 (declare-function treemacs-toggle-node "ext:treemacs-interface")
+(declare-function treemacs-add-and-display-current-project-exclusively "ext:treemacs")
```
rswgnu, line 48: Alphabetize these declarations.

hywiki.el
```diff
@@ -139,6 +139,7 @@
 (defvar org-agenda-buffer-tmp-name)  ;; "org-agenda.el"
 (defvar org-export-with-broken-links);; "ox.el"
 (defvar org-publish-project-alist)   ;; "ox-publish.el"
+(defvar action-key-modeline-buffer-id-function)  ;; "hui-mouse"
```
rswgnu, line 142: Alphabetize.

hyrolo.el
rswgnu: So you are saying it is not used anywhere?  If so, just remove it.  If there are no other references to it anywhere in the code and the return value of its so setq is not used then it is just a leftover from some prior version and you should remove it.

rswgnu (changes requested):

rswgnu (approved):

Issue #603, opened 2024-11-11
Can't use `org-meta-return` in org buffer even after configuring


edenworky:
I tried using `C-h h c o`, but none of the available options (including "None") disables the default `M-RET` in org buffers, i.e. still doesn't create a new heading when invoked at the end of a line.

matsl:
Hi @edenworky Thanks for the report. I can recreate the described behavior. Looking into it. :pray:

matsl:
Hi @edenworky With #607 `None` and `Hyperbole-Buttons-Only` should give you the desired behavior. With `All-Hyperbole-Contexts` smart-eolp behavior will be active. _(In the later case you will still get the org behavior if at the last line of a file but I'm looking into closing that gap.)_

rswgnu:
Hi @edenworky:

I'm pretty sure this is resolved as of today with commit #a7aefb122.  Hyperbole's end-of-line behavior should now defer to Org-mode when the config option is set to None, meaning Hyperbole's M-RET control is never active or if set to :buttons, meaning it is active only when on text Hyperbole recognizes as a hyperbutton.  Additionally, we have improved end-of-buffer handling to be consistent with end-of-line handling so on the last line of the buffer if there is any non-whitespace text, behavior will match that of any other line.

Get the latest from melpa or elpa-devel and let us know after restarting your Emacs if it works as you expect and then we'll close this issue.

PR #607, opened 2024-11-15
Dispatch smart-eolp case to smart-org

matsl:
# What

When Hyperbole context in org-mode should not be active, drop to later
handlers in hkey-alist i.e. possibly smart-org handler.

# Why

When we say that a nil value `hsys-org-enable-smart-keys` make M-RET
behave as it normally does is suggests that the smart-eol function
should not be active in org mode. Or!?

# Note

Issue comes from report in #603.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2024-11-16  Mats Lidell  <matsl@gnu.org>
+
+* test/hsys-org-tests.el (hsys-org--meta-return-on-end-of-line): Add test
+    case for eol action in org-mode.
+
+* hui-mouse.el (hkey-alist): Guard action-key-eol-function from acting
+    when we should be delegate to org-mode.
+
 2024-11-13  Bob Weiner  <rsw@gnu.org>
 
 * hywiki.el (hywiki-maybe-dehighlight-page-name,
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,7 +4,7 @@
     case for eol action in org-mode.
 
 * hui-mouse.el (hkey-alist): Guard action-key-eol-function from acting
-    when we should be delegate to org-mode.
+    when we should delegate to org-mode.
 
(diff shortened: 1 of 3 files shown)


ChangeLog
```diff
@@ -1,3 +1,11 @@
+2024-11-16  Mats Lidell  <matsl@gnu.org>
+
+* test/hsys-org-tests.el (hsys-org--meta-return-on-end-of-line): Add test
+    case for eol action in org-mode.
+
+* hui-mouse.el (hkey-alist): Guard action-key-eol-function from acting
+    when we should be delegate to org-mode.
```
rswgnu, line 7: ‘should delegate’, no ‘be’.

test/hsys-org-tests.el
```diff
@@ -225,6 +225,58 @@ This is independent of the setting of `hsys-org-enable-smart-keys'."
...
+        (goto-char 1)
+        (end-of-line)
+        (with-mock
+          (mock (hsys-org-meta-return) => t)
+          (should (equal hsys-org-enable-smart-keys v)) ; Ert traceability
+          (should (action-key))))))
+  (let ((hsys-org-enable-smart-keys t)
+        (v t))
+    ;;; One line no return
+    ;; At end of line is also end of file so smart-eolp filters out
```
matsl, line 255: @rswgnu Any thoughts about this corner case?

matsl:
> Since with the mouse, you can select our points in the window beyond the end of the buffer and expect that to set the buffer to point-max, in this case, we should treat EOL as EOB, that is prioritizing the end of buffer over end of line, whatever that means for the code.

Yes, I understand that but the behavior is likely confusing for the user. Not sure what is the best solution. We could signal an error if at EOB and hsys-org-enable-smart-keys=t. 

Anyway, that is for another PR if we want to fix that. Thanks for approving. It has been merged and pushed.

PR #613, opened 2024-11-23
Use default value of debug-on-error

matsl:
# What

Was set to produce a backtrace but it is not required. See
"(elisp)Batch Mode".

# Why

Emacs 31 introduces calling man using `with-demoted-errors` which
causes an error when requiring man when `debug-on-error` is set to
`t`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,8 @@
 2024-11-23  Mats Lidell  <matsl@gnu.org>
 
-* Makefile (BATCHFLAGS): Use debug-on-error's default value nil.
+* Makefile (BATCHFLAGS): Use debug-on-error's default value nil. Emacs 31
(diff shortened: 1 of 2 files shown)


ChangeLog
```diff
@@ -1,3 +1,7 @@
+2024-11-23  Mats Lidell  <matsl@gnu.org>
+
+* Makefile (BATCHFLAGS): Use debug-on-error's default value nil.
+
```
rswgnu, line 4: In addition to above text, add this:

Emacs 31 introduces calling man using with-demoted-errors which
causes an error when requiring man when debug-on-error is set to
t.

rswgnu (changes requested):

rswgnu (approved):

PR #614, opened 2024-11-23
Test sections after HywikiWords

matsl:
# What

Add test of sections after HywikiWords.

# Why

Document current behavior.

# Note

This is WIP. Started to write tests for sections with dash and space. Feel free to suggest cases I should address. You need to fill me in on the details of the "space to dash", or was it "dash to space", conversion, so I can write those as well, Or if you chose to add those when you work on that.


matsl:
Moved the test to #620 Closing this PR.

PR #618, opened 2024-11-25
Use kotl-kview

matsl:
# What

Use kotl-kview when changing label type from menu.

# Why

This was likely missed during refactoring when the prefix kotl was
added to the kview local variable.

### DIFF
--- a/kotl/kmenu.el
+++ b/kotl/kmenu.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    28-Mar-94 at 11:22:09
-;; Last-Mod:      3-Oct-23 at 16:43:12 by Mats Lidell
+;; Last-Mod:     25-Nov-24 at 22:11:56 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


kotl/kmenu.el
```diff
@@ -91,11 +91,11 @@
       ["First-Cell"          kotl-mode:beginning-of-buffer  t]
       ["Last-Cell"           kotl-mode:end-of-buffer        t])
     '("Label-Type"
-      ["Alphanumeric (Default)"  (kview:set-label-type kview 'alpha)  t]
-      ["Legal"                   (kview:set-label-type kview 'legal)  t]
```
rswgnu, line 95: Shouldn’t we have gotten unbound dynamic variable warnings here?

rswgnu (approved):
See my inline question.

kotl/kmenu.el
matsl: Yes, at a first look you would think that. Can it be that `INITVALUE` of the defconst is just an s-exp with no real meaning until it is used as a menu? 

The `kmenu.el` file is free of requiring any packages that define functions used in the menus.

rswgnu: Yes, the menu expressions are quoted, so not evaluated until used, so I would just expect these to fail when used and therefore need testing to see if that were true prior to applying the suggested fix.

PR #621, opened 2024-11-30
Rename remaining hywiki-get-page references to hywiki-get-referent

matsl:
# What

Rename remaining hywiki-get-page to hywiki-get-referent.

# Why

Some hywiki-get-page references are still left in comments.


rswgnu:
This will be taken care of by the next rsw branch push, so it is unneeded.

PR #625, opened 2024-12-22
Fix tests, make result path absolute

matsl:
# What

Tests were dependent on where the suite was run. Set default-directory
for the two tests where this caused problems to force an absolute path
to be produced.

# Why

Tests should work wherever they are executed.


rswgnu (approved):

rswgnu:
This was applied to master.  -- rsw

PR #627, opened 2024-12-24
add hywiki words with delimiter tests

matsl:
# What

hywiki tests for highlighting and identification of WikiWords.

# Why

WikiWords are not identifies as such in all cases. In some cases
org-links are produced and in others WikiWords are not identified
since adjacent text or symbols takes precedence. So we need to have
test cases that checks there are no regressions.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2024-12-24  Mats Lidell  <matsl@gnu.org>
+
+* test/hywiki-tests.el (hywiki-tests--a-wikiword-in-hywiki-directory)
+    (hywiki-tests--wikiword-identified-with-delimiters): Add tests for
+    hywiki words with delimiters.
+
 2024-12-23  Mats Lidell  <matsl@gnu.org>
 
 * test/hywiki-tests.el (hywiki-tests--add-activity)
--- a/test/hywiki-tests.el
+++ b/test/hywiki-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell
 ;;
 ;; Orig-Date:    18-May-24 at 23:59:48
-;; Last-Mod:     24-Dec-24 at 00:15:00 by Mats Lidell
+;; Last-Mod:     24-Dec-24 at 00:48:15 by Mats Lidell
 ;;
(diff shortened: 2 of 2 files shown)


matsl:
Hi @rswgnu 

I was cleaning my workspace and found this branch about identifying `WikiWords` when there are different adjacent characters etc. I have some slight memory of us discussing this but I don't remember the outcome of that discussion. So I rebased it on top of master and made it succeed for all cases. So it covers the current behavior. Can you take a look if the tests make sense? Any more cases that should be added?

rswgnu (approved):

rswgnu:
In a second set of tests, we’ll need to add testing of #this-is-a-section and (WikiWord#Section2), delimited tests.  Ignore spaces for now since we are going to disallow thise in section names and convert to dashes somehow, at which point we can add those tests. -- BobOn Dec 24, 2024, at 3:38 AM, Mats Lidell ***@***.***> wrote:﻿
Hi @rswgnu
I was cleaning my workspace and found this branch about identifying WikiWords when there are different adjacent characters etc. I have some slight memory of us discussing this but I don't remember the outcome of that discussion. So I rebased it on top of master and made it succeed for all cases. So it covers the current behavior. Can you take a look if the tests make sense? Any more cases that should be added?

matsl:
> In a second set of tests, we’ll need to add testing of #this-is-a-section and (WikiWord#Section2), delimited tests.  Ignore spaces for now since we are going to disallow thise in section names and convert to dashes somehow, at which point we can add those tests.

Yes. I think there is a small start on that but will try to expand on it.

rswgnu:
Sounds good.  Focus on ensuring proper behavior within balanced delimiters or sparsted by punctuation characters.-- BobOn Dec 24, 2024, at 5:52 PM, Mats Lidell ***@***.***> wrote:﻿

In a second set of tests, we’ll need to add testing of #this-is-a-section and (WikiWord#Section2), delimited tests.  Ignore spaces for now since we are going to disallow thise in section names and convert to dashes somehow, at which point we can add those tests.

Yes. I think there is a small start on that but will try to expand on it.

PR #632, opened 2024-12-27
silence obsolete when let et al warnings for emacs 31

matsl:
# What

Add EXTRA_COMP_FLAGS to control extra flags for byte and native
compilation.

# Why

when-let et al are marked as obsolete from 31.1 and causes a lot of
warning messages that disturbs the output. As a quick fix introduce
the make macro so we can add settings to turn these warnings off. When
31.1 becomes closer to release we can deal with this.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2024-12-27  Mats Lidell  <matsl@gnu.org>
+
+* Makefile (EXTRA_COMP_FLAGS): Macro for extra flags to Emacs for byte and
+    native compilation.
+
 2024-12-27  Bob Weiner  <rsw@gnu.org>
 
 * test/hywiki-tests.el (hywiki-tests--add-activity,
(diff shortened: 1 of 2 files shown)


rswgnu (approved):

rswgnu:
What replaces these two calls that is not obsolete?  If the others have been around for a while, we should just switch.

matsl:
> What replaces these two calls that is not obsolete? If the others have been around for a while, we should just switch.

You are right. I was fooled by the deprecation warning. It says: *This macro is obsolete since 31.1; use ‘when-let*’ or ‘and-let*’
instead.* And when looking at `when-let*` it said *Probably introduced at or before Emacs version 31.1*. But the later is not really true. I checked with older versions of Emacs and they have `when-let*` and `and-let*` (for some versions they are in `subr-x.el`)

So it should be possible to replace these in a portable manner. I'll come back with another PR using that instead.

matsl:
Maybe it was amplified for me by the discussion in bug: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=+73853 where the deprecation got some criticism.

PR #635, opened 2024-12-28
Verify creation of pages and movement to pages

matsl:
# What

Added or modified basic tests for creating pages and moving to pages
with sections.  Misc new tests and refactoring of old tests.  ~Update
hywiki to use sections.~

# Why

Moving to pages and section in pages is basic HyWiki functionality so
we need to cover that in our tests.

# Note

~Hywiki was missing using the sections with hpath:find. Added that in
order to write the test cases. Feel free to modify.~

Rebased to use the update on the master branch.

### DIFF
--- a/test/hywiki-tests.el
+++ b/test/hywiki-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell
 ;;
 ;; Orig-Date:    18-May-24 at 23:59:48
-;; Last-Mod:     28-Dec-24 at 16:58:48 by Mats Lidell
+;; Last-Mod:     28-Dec-24 at 22:14:33 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -53,7 +53,7 @@
         (should-not (cdr (hywiki-add-page "notawikiword")))
       (hy-delete-dir-and-buffer hywiki-directory))))
 
-(ert-deftest hywiki-tests--action-key-on-wikiword-displays-page ()
+(ert-deftest hywiki-tests--action-key-on-hy-wikiword-displays-page ()
   "Verify `action-key' on a prefixed WikiWord, outside of hywiki-directory, creates a new page."
   (let ((hsys-org-enable-smart-keys t)
         (hywiki-directory (make-temp-file "hywiki" t))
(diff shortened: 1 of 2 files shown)


hywiki.el
```diff
@@ -1108,8 +1108,10 @@ an existing or new HyWikiWord."
       (call-interactively hywiki-display-page-function)
     (when (null wikiword)
       (setq wikiword (hywiki-word-read-new "Find HyWiki page: ")))
-    (let ((file (hywiki-get-file (or file-name wikiword))))
-      (funcall hywiki-display-page-function file)
+    (let ((file (hywiki-get-file (or file-name wikiword)))
+          (section (when (string-match "#[^#]+$" wikiword)
+                     (substring wikiword (match-beginning 0)))))
+      (funcall hywiki-display-page-function (concat file section))
```
rswgnu, line 1114: On this fix, you'll have to wait for me to push my update to this function and then drop yours, as it is more comprehensive.

test/hywiki-tests.el
```diff
@@ -65,21 +65,72 @@
...
+              (goto-char (point-min)))))
+      (hy-delete-file-and-buffer hywiki-page-file)
+      (hy-delete-dir-and-buffer hywiki-directory))))
+
+(ert-deftest hywiki-tests--action-key-on-wikiword-and-section-displays-page ()
+  "Verify `action-key' on a WikiWord with section moves to the section."
+  (let* ((hsys-org-enable-smart-keys t)
+         (hywiki-directory (make-temp-file "hywiki" t))
+	 (hywiki-page-file (expand-file-name "WikiWord.org" hywiki-directory))
+         (sections '("* Header" "** SubHeader" "*** SubSubHeader")))
```
rswgnu, line 117: I would add a single line body within each header section, to ensure the tests work in such cases.

test/hywiki-tests.el
```diff
@@ -132,6 +183,15 @@
               (goto-char 6)
               (should (string= "WikiWord" (hywiki-word-at)))))
 
+          ;; Does not match as a WikiWord
+          (dolist (v '("WikiWord#"))
```
rswgnu, line 187: An Action Key press on a WikiWord with just a trailing # should probably either go to the first Org doc section or trigger an error.  What do you prefer?

hywiki.el
matsl: > On this fix, you'll have to wait for me to push my update to this function and then drop yours, as it is more comprehensive.

That is OK and expected. Do your update and I'll remove this change. It is mostly there to allow me to work on these tests.

test/hywiki-tests.el
matsl: > An Action Key press on a WikiWord with just a trailing # should probably either go to the first Org doc section or trigger an error. What do you prefer?

I think it currently raises an error. _(Hyperbole Action Key): No action defined for this context; try another location_ That is OK. Maybe just handle it as WikiWord would cause less questions from a user but will think about it.

hywiki.el
rswgnu: Fix has  been pushed to master.

test/hywiki-tests.el
matsl: Thinking more on this I think `WikiWord#` should be treated just like `WikiWord#non-existing-section`. 

Currently we just move to the WikiWord-page if the section does not exists, There is an error message that the section can't be found.  With an _empty_ section we should do something similar. We could even strip of the hash mark and treat it as a normal WikiWord.

rswgnu: I think a wikiword that links to a page will display an existing buffer page with point wherever it was before the button was activated.  We need a way to say, display the beginning or first section of the page, so this is as good a way as any.

rswgnu: I changed my thought on this.  With my next update, you'll be able to add :L1 at the end of a HyWikiWord referent to go to the first line of a wiki page, so if a # is given without a section name, it the HyWikiWord reference will just be ignored for now.

matsl: OK. I will updated these tests when that change goes in.

test/hywiki-tests.el
```diff
@@ -53,7 +53,7 @@
         (should-not (cdr (hywiki-add-page "notawikiword")))
       (hy-delete-dir-and-buffer hywiki-directory))))
 
-(ert-deftest hywiki-tests--action-key-on-wikiword-displays-page ()
+(ert-deftest hywiki-tests--action-key-on-hy-wikiword-displays-page ()
```
rswgnu, line 56: No need for a dash, just use 'hywikiword' instead of 'hy-wikiword'.

test/hywiki-tests.el
```diff
@@ -65,21 +65,72 @@
           (goto-char 4)
           (action-key)
 	  (should (equal (cons 'page wikifile) (hywiki-get-referent "WikiWord"))))
-      (hy-delete-file-and-buffer wikifile))))
+      (hy-delete-file-and-buffer (expand-file-name wikifile hywiki-directory))
+      (hy-delete-dir-and-buffer hywiki-directory))))
 
-(ert-deftest hywiki-tests--assist-key-on-wikiword-displays-help ()
+(ert-deftest hywiki-tests--assist-key-on-hy-wikiword-displays-help ()
```
rswgnu, line 71: Same as prior comment.

test/hywiki-tests.el
rswgnu: That update has been made, so you can finish this PR.

rswgnu (changes requested):

matsl:
WikiWord# still errors. I got the impression that a single '#' would now be ignored so that WikiWord# still would link to the page?

rswgnu:
Proper behavior right now is that WikiWord# is not recognized nor
highlighted as a HyWikiWord.  Therefore, it is left for any other Action
Key action to process or the generic 'no action' error to occur.  That is
how you should test it right now.  It is too much effort to try to catch
that as a misformed HyWikiWord right now.  -- Bob

matsl:
OK. The test is in line with that behavior so nothing needs to be updated.

rswgnu (approved):

matsl:
@rswgnu There is a regression on master. Moving to wikiword-sections and also line-column with wikiwords does not work. The test that verifies sections fails so this PR is not possible to merge. 

I have also verified manually on master that sections and line-column does not work so it is not necessarily a bad test.

matsl:
@rswgnu I added the test case with line and column numbers on WikiWords to this PR and marked that and the section (or suffix) test case as `expected failed.`  It was easiest in this case I think. You can merge this to master and work from there to fix the regressions since the error is already on master.

PR #638, opened 2024-12-30
Ensure info index is correct

matsl:
# What

Ensure info index is correct.

# Why

We generate the info pages with texi2any and patch them for
compatibility. If this would fail due to changes in tex2any or our
Makefile the indexes for the hyperbole info pages could be broken.


test/hypb-tests.el
```diff
@@ -74,6 +75,15 @@
   (should (= (hypb:oct-to-int 2000) 1024))
   (should-error (hypb:oct-to-int 8) :type 'error))
 
+(ert-deftest hypb--verify-info-index-is-correct ()
+  "Verify that Hyperbole info page indexes are identified.
+See Emacs bug#74042 related to usage of texi2any."
+  (unwind-protect
+      (progn
+        (Info-goto-node "(Hyperbole)Top")
+        (should (set:equal '("Key Index" "Function" "Concept Index") (Info-index-nodes))))
```
rswgnu, line 84: Shouldn’t this be “Function Index”?

rswgnu (changes requested):

test/hypb-tests.el
matsl: Currently it is just named "Function" as the test case shows. Se picture below. 

But it has looked strange to me to just call it function so in that sense you are right. I could look into how to change it to "Function index". (Ought to be simple.)

Skärmbild_2024-12-30_23-22-40

matsl: > Shouldn’t this be “Function Index”?

Renamed it to `Function index` in the manual and updated the test case to reflect that.

rswgnu (approved):

PR #639, opened 2025-01-01
Use link-to-Info-index-item for info index display

matsl:
# What

Use link-to-Info-index-item.

Add test that verifies link-to-Info-index-item is called for
displaying the referent.

# Why

Bug fix.

### DIFF
--- a/test/hactypes-tests.el
+++ b/test/hactypes-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    30-Jan-21 at 12:00:00
-;; Last-Mod:      2-Jan-25 at 22:32:05 by Mats Lidell
+;; Last-Mod:      3-Jan-25 at 23:52:37 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -44,7 +44,8 @@
         (actypes::link-to-Info-index-item "(hyperbole)hyperb:dir")
         (should (string-prefix-p "*info*" (buffer-name)))
         (should (string= "hyperbole" (file-name-nondirectory Info-current-file)))
-        (should (string= "Documentation" Info-current-node)))
+        (should (string= "Documentation" Info-current-node))
+        (should (looking-at-p "The Hyperbole Manual is a reference manual, not a simple introduction\\.")))
     (kill-matching-buffers "^\\*info\\*" nil t)))
 
 (provide 'hactypes-tests)
--- a/test/hactypes-tests.el
+++ b/test/hactypes-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    30-Jan-21 at 12:00:00
-;; Last-Mod:      8-Jul-23 at 14:11:49 by Bob Weiner
+;; Last-Mod:      2-Jan-25 at 22:32:05 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -19,6 +19,7 @@
 ;;; Code:
 
 (require 'ert)
+(require 'el-mock)
 (require 'hactypes)
 (require 'hy-test-helpers "test/hy-test-helpers")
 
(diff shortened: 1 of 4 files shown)


test/hywiki-tests.el
```diff
@@ -725,11 +725,15 @@ Note special meaning of `hywiki-allow-plurals-flag'."
...
-	  (hywiki-add-info-index "WikiWord")
-	  (should (equal '(info-index . "(info)index-name") (hywiki-get-referent "WikiWord"))))
+        (progn
+	  (mocklet (((info) => t)
+		    ((Info-read-index-item-name "Info index item: ") => "index-name")
+		    ((Info-current-filename-sans-extension) => "info"))
+	    (hywiki-add-info-index "WikiWord")
+	    (should (equal '(info-index . "(info)index-name") (hywiki-get-referent "WikiWord"))))
+          ;; Verify display of referent
+          (mocklet (((hact 'link-to-Info-index-item "(info)index-name") => t))
```
rswgnu, line 735: Don’t like this part.   Ould you actually do the display and then check the buffer and current line that it matches expectations.  It could be a parial match to the line that wnsures it contains the index entry.

rswgnu (changes requested):

test/hywiki-tests.el
matsl: I was afraid you would say that ;-)

I think we should avoid  "integration tests". Instead try to do unit tests as much as possible. 

In this case I would prefer to keep this test as it is and add a test that `link-to-Info-index-item` does its job. This test ensures that our code under test calls the function we intend it to call. For this test we trust `link-to-Info-index-item` to do its job as we cover that in other tests (admittedly that has to be written, since we currently only cover info nodes but you see the idea?)

rswgnu: Ok, with a separate test, that will work.  I’ll approve and watch for the second one.

test/hactypes-tests.el
```diff
@@ -32,5 +33,19 @@
   (should (actypes::display-boolean nil))
     (hy-test-helpers:should-last-message "Result = nil; Boolean value = False"))
 
+(ert-deftest hactypes-tests--link-to-Info-index-item ()
```
matsl, line 36: Added this test that both does a unit test that we do call id-info-item but also goes the full way and checks the right info page is displayed. That might be non orthodox, and arguable partly against what I said about unit tests, but is a way to balance the number of test we need to write with the amount of code they verify.

test/hactypes-tests.el
```diff
@@ -32,5 +33,19 @@
...
+  (should-error (actypes::link-to-Info-index-item "(unknown-file)unknown-index-item") :type 'error)
+  (mocklet (((id-info-item "(infofile)index-item") => t))
+    (actypes::link-to-Info-index-item "(infofile)index-item"))
+  (unwind-protect
+      (progn
+        (actypes::link-to-Info-index-item "(hyperbole)hyperb:dir")
+        (should (string-prefix-p "*info*" (buffer-name)))
+        (should (string= "hyperbole" (file-name-nondirectory Info-current-file)))
+        (should (string= "Documentation" Info-current-node)))
+    (kill-matching-buffers "^\\*info\\*" nil t)))
```
rswgnu, line 48: Still need to verify point goes to the line linked to by the Info ondex entry within the node.  Otherwise, you are just testing a link to a node itself.

rswgnu (changes requested):

test/hactypes-tests.el
matsl: We look at this differently. I'm thinking we use Emacs function `Info-index` for moving the point etc (The implementation of `actypes::link-to-Info-index-item.` uses that.) We don't have to verify that it is doing its job. It is up to Emacs to ensure that. Besides we know that the index `(hyperbole)hyperb:dir` goes to the Documentation node.

We could add a unit test of our function `id-info-item` that verifies we actually are calling `Info-index` with the right `topic`.

I will add a test that shows we are at the line in the Document node but not sure it really makes the test better.

rswgnu: Pick an index entry that points somewhere other than the first line of a node and test that one, to prove index entries can behave differently than node links.

rswgnu (changes requested):

rswgnu (approved):

test/hactypes-tests.el
matsl: Thanks for approving. 

I think the current test meets the criteria. For a node, point is not on the first paragraph but on the first empty line after the first breadcrumbs-line (if its called that.) In addition to that there is no node for (hyperbole)hyperb:dir. It only exists in the index.

PR #645, opened 2025-01-09
add test for c w

matsl:
# What

Add test for hui-kill-region

# Why

New function.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,7 @@
 2025-01-09  Mats Lidell  <matsl@gnu.org>
 
+* test/hui-tests.el (hui--kill-region): Test.
+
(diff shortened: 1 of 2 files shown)


matsl:
I'm not super happy about how these tests turned out. Yet another case where batch vs interactive execution causes different behavior. Which is critical here since hui-kill-region depends on that. I think that funcall-interactively does not work as expected and I'm trouble shooting this now. I still believe that the tests as they are now works so could go in and be later refactored when I understand the problem better. (But we can also let this be WIP for now.)

matsl:
@rswgnu Tests are now simplified and works both in batch and interactive mode. PTAL

matsl:
See #646 for another go at this.

PR #646, opened 2025-01-14
add test for \C-w using prefix arg

matsl:
# What

- Add test for hui-kill-region
- Refactored and simplified
- Use prefix arg for interactive call
- Refactor to have base behavior as kill-region
- Fix non-interactive behavior

# Why

Tests.

# Note

The inspiration for using an argument for identification of
interactive call or not comes from the comment in 
```
{C-hf called-interactively-p RET}
```

It says:

    Instead of using this function, it is cleaner and more reliable to
    give your function an extra optional argument whose ‘interactive’ spec
    specifies non-nil unconditionally ("p" is a good way to do this), or
    via (not (or executing-kbd-macro noninteractive)).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,7 @@
 2025-01-09  Mats Lidell  <matsl@gnu.org>
 
+* test/hui-tests.el (hui--kill-region): Test.
+
 * test/MANIFEST: Add hycontrol-tests.el
 
 * hycontrol.el (hycontrol-framemove-direction): Quit hycontrol before
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,9 @@
-2025-01-09  Mats Lidell  <matsl@gnu.org>
+2025-01-19  Mats Lidell  <matsl@gnu.org>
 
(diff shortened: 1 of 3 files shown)


rswgnu:
Be sure to test for 2 kills in a row that they append as one kill properly:
  case 1:  kill based on matching pair at point followed by a regular C-k line kill
  case 2: regular C-k line kill followed by a matching pair at point kill.

rswgnu:
And we'll review this when we chat.

matsl:
> Be sure to test for 2 kills in a row that they append as one kill properly: case 1: kill based on matching pair at point followed by a regular C-k line kill 

There is no validation of what is actually saved in the current tests so will add that in one or more separate tests I think.

> case 2: regular C-k line kill followed by a matching pair at point kill.

That is an interesting case. Will take a look.

rswgnu:
I think the signature of our overriding kill-region command should match the signature of kill-region as best as possible as well.

matsl:
Added some tests for `kill-region-type` of functionality.

PR #651, opened 2025-01-20
make failing test case succeed on test all

matsl:
# What

- Add helper for making random WikiWord suitable for tests
- Modify test to succeed on test-all

# Why

Test fails on make test-all and this PR installs a work-around by
creating an unused initial WikiPage which seems to fix the issue. The
failing test case is left commented out in the code for later
investigation and ideas what can cause this.

### DIFF
--- a/test/hy-test-helpers.el
+++ b/test/hy-test-helpers.el
@@ -3,11 +3,11 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    30-Jan-21 at 12:00:00
-;; Last-Mod:     20-Jan-25 at 20:21:34 by Mats Lidell
+;; Last-Mod:     21-Jan-25 at 17:02:36 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
-;; Copyright (C) 2021-2024  Free Software Foundation, Inc.
+;; Copyright (C) 2021-2025  Free Software Foundation, Inc.
 ;; See the "HY-COPY" file for license information.
 ;;
 ;; This file is part of GNU Hyperbole.
--- a/test/hy-test-helpers.el
+++ b/test/hy-test-helpers.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    30-Jan-21 at 12:00:00
-;; Last-Mod:     20-Jan-25 at 00:24:32 by Mats Lidell
+;; Last-Mod:     20-Jan-25 at 20:13:23 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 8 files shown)


test/hy-test-helpers.el
```diff
@@ -100,5 +100,20 @@ Checks ACTYPE, ARGS, LOC, LBL-KEY and NAME."
       (kill-buffer buf))
     (delete-directory dir)))
 
+(defun hy-make-random-wikiword (&optional length word-length)
+  "Return a random WikiWord of LENGTH and with WORD-LENGTH.
```
rswgnu, line 104: Change word-length to word-count, easier to understand.

rswgnu (changes requested):

test/hy-test-helpers.el
matsl: Hmm... `word-count` suggest to me the number of words in the WikiWord rather than the length of each word!?

rswgnu:
So what is the length argument if not the length of a word?  Just need clarity on each arg.-- BobOn Jan 20, 2025, at 12:28 PM, Mats Lidell ***@***.***> wrote:﻿



In test/hy-test-helpers.el:
> @@ -100,5 +100,20 @@ Checks ACTYPE, ARGS, LOC, LBL-KEY and NAME."
       (kill-buffer buf))
     (delete-directory dir)))
 
+(defun hy-make-random-wikiword (&optional length word-length)
+  "Return a random WikiWord of LENGTH and with WORD-LENGTH.

Hmm... word-count suggest to me the number of words in the WikiWord rather than the length of each word!?

matsl:
> So what is the length argument if not the length of a word?  Just need clarity on each arg.

Docstring is more verbose now.

Thought I would throw in a test of the function too but that turned into something, lest say not hard, but more work than expected. PTAL.

test/hy-test-helpers.el
```diff
@@ -100,5 +100,22 @@ Checks ACTYPE, ARGS, LOC, LBL-KEY and NAME."
       (kill-buffer buf))
     (delete-directory dir)))
 
+(defun hy-make-random-wikiword (&optional length word-length)
+  "Create a random WikiWord (string).
+The WikiWord will have a total of LENGTH characters.  Each word part of
+the WikiWord will have WORD-LENGTH characters.  The default LENGTH is 12
```
rswgnu, line 106: What is ‘each word part of the WikiWord’?  Doesn’t that still sound unclear to you?  The first part is better where you say each WikiWord has LENGTH characters.  Maybe you mean the #section part of a WikiWord reference.  We should define terms.

test/test-helpers-tests.el
```diff
@@ -0,0 +1,44 @@
+;;; test-helpers-tests.el --- tests of the test helpers -*- lexical-binding: t; -*-
+;;
+;; Author:       Mats Lidell
+;;
+;; Orig-Date:    20-Jan-25 at 20:22:05
+;; Last-Mod:     20-Jan-25 at 20:56:51 by Mats Lidell
+;;
+;; SPDX-License-Identifier: GPL-3.0-or-later
+;;
+;; Copyright (C) 2021-2022  Free Software Foundation, Inc.
```
rswgnu, line 10: Update to 2025.

matsl: Thanks. Good catch. Need to update the templates or make the copyright year being generated.

test/hy-test-helpers.el
matsl: Yes. My suggestion is that a WikiWord is not a word. It is words written in a camel case style. Just like an identifier is not a word but can be built up from words stacked together.

So WikiWord-length is the length of the whole WikiWord. And word length is the length or the words that together form the WikiWord. I feel the specification with part or section o component is not really necessary but I can live with that.  (For camel cased identifiers Gemini suggests word, component or part with a preference for word.)

(In a real WikiWord the words have any length of course, but for a random WikiWord, having the same length on all words is a compromise. The last word might have a different length though to make up the WikiWord length. And there are other edge cases. But you get all that.)

rswgnu:
I get it now.  The initial phrasing threw me off.
Thanks.

rswgnu (approved):

PR #654, opened 2025-01-26
Remove deploy static content to pages action

matsl:
# What

Remove deploy static content to pages action.

# Why

It is not needed. The default, deploy from a branch, works for us. We link to the manual from the readme.

# Note

The setting under Settings/Pages need to be set back to "Deploy from a branch" together with merging this.


rswgnu:
All set with this now.

PR #659, opened 2025-01-30
Use replace-regexp-in-string for compatibility with Emacs 27.2

matsl:
# What

Use replace-regexp-in-string since string-replace is not available in
Emacs 27.2.

# Why

We have just one call using string-replace so instead using the older
replace-regexp-in-string seems like a good candidate.

# Note

~We have no tests for this so it is just a textual substitution with
no extensive verification.~ A test has been added.

We could consider using compat for using string-replace. We have done
this change before so it seem the simpler string-replace meets our
needs. Or we could take a middle step and reintroduce the code we had
from compat and use that in a hypb-compat file to be required when we
use functions that we need forward compatibility for!?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2025-01-30  Mats Lidell  <matsl@gnu.org>
+
+* hui-mini.el (hui:menu-read-from-minibuffer): Use
+    replace-regexp-in-string for compatibility with Emacs 27.2.
+
 2025-01-29  Mats Lidell  <matsl@gnu.org>
 
 * Makefile (HYPB_GEN, HYPB_ELC, HYPB_ELC_ELN, HYPB_at): Verbosity macros.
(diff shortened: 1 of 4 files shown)


rswgnu:
Did this appear as an undefined function in byte-compiler warnings?  Why didn’t we see it before?

matsl:
> Did this appear as an undefined function in byte-compiler warnings? 

Yes, apparently. See example of a [warning summary](https://github.com/rswgnu/hyperbole/actions/runs/12979385839) from a recent master build:  

> Why didn’t we see it before?

Good question. I think we have not made it a habit of checking the warning summary page.

The issue here is that string-replace is not available in subr in Emacs 27.2. We have had issues with that in the past but it crept in again.

By the way. We have a warning for 29.4: _"hywiki.el:1091:26: Warning: `org-id-get' called with 4 arguments, but accepts only 0-3". I think we have talked about it but we did not get around to fix it._ I have added it now to the `Todo.`

rswgnu:
Thanks.

PR #666, opened 2025-02-06
Return hasht when rebuilding from load data

matsl:
# What

Return hasht when rebuilding from load data and add test cases that
depends on the fix.

# Why

The result of hywiki-make-referent-hasht is used as a hash table in
some case so it is important that the new value is returned.

The successful test case depends on this change.

The expected failed test case illustrates that the key series is
inserted in the buffer as well as in the definition of the
referent. This can be shown manually too so it does not only happen in
the test case. This test case is only activated when run interactively
like test-all.

### DIFF
--- a/test/hywiki-tests.el
+++ b/test/hywiki-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell
 ;;
 ;; Orig-Date:    18-May-24 at 23:59:48
-;; Last-Mod:     26-Jan-25 at 18:22:58 by Bob Weiner
+;; Last-Mod:      7-Feb-25 at 00:46:50 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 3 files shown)


test/hywiki-tests.el
```diff
@@ -922,5 +922,56 @@ Note special meaning of `hywiki-allow-plurals-flag'."
...
+
+          ;; Simulate reload from cache
+          (setq hywiki--referent-hasht nil)
+          (hywiki-make-referent-hasht)
+          (should (equal '(key-series . "{ABC}") (hywiki-get-referent wiki-referent))))
+      (hy-delete-files-and-buffers (list wiki-page (hywiki-cache-default-file)))
+      (hy-delete-dir-and-buffer hywiki-directory))))
+
+(ert-deftest hywiki-tests--save-referent-use-hyperbole-menu ()
+  "Verify saving and loading a referent works when using Hyperboles menu."
```
rswgnu, line 949: Hyperbole’s

test/hywiki-tests.el
```diff
@@ -922,5 +922,56 @@ Note special meaning of `hywiki-allow-plurals-flag'."
...
+  :expected-result :failed
```
rswgnu, line 950: I don’t understand the expected fail.  Are you showing that you found an issue with key series as a referent that your hasht fix does not resolve?  If so, you should also comment the code with the problem rather than just letting the failed test demonstrate it, for my small mammalian brain.

rswgnu (changes requested):

test/hywiki-tests.el
```diff
@@ -922,5 +922,56 @@ Note special meaning of `hywiki-allow-plurals-flag'."
...
+  ;; The entered key series is inserted into the WikiWord file.
```
matsl, line 951: > I don’t understand the expected fail. Are you showing that you found an issue with key series as a referent that your hasht fix does not resolve? If so, you should also comment the code with the problem rather than just letting the failed test demonstrate it, for my small mammalian brain.

Here is the comment that tries to explain it but I can move it down to the should check that fails.

rswgnu (approved):

PR #667, opened 2025-02-08
Fix hywiki publishing

matsl:
# What

Highlight wiki page names before looking for overlays to turn into
org-links. Let org handle the sitemap file.

# Why

The wiki pages are copied during export so the overlays are not
immediately available so we need to force the highlighting. 

The sitemap is already handled by org export so we do not need to
touch that. This is no optimisation but rather that adding the
hy-prefix trips org export and outputs link with `hy:` verbatim in the
link text which is not what we want.

# Note

No tests are added yet. Working on that.

What do you think of not using the overlays for the finding the wiki
words to convert? Maybe a more direct way to find the page names and
convert them would be preferred? Instead of going pass inserting
overlays and then find them!? (But this seems to work so for now I'm
happy with it.)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2025-02-08  Mats Lidell  <matsl@gnu.org>
+
+* hywiki.el (hywiki--sitemap-file): Helper function for getting the sitemap
+    file-name.
+    (hywiki-org-export-function): Only use org publishing for the sitemap file.
+    (hywiki-convert-words-to-org-links): Highlight wiki page names so overlays
+    are defined before looking for them.
+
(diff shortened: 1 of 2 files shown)


rswgnu:
Great.  It is strange as I had verified that the overlays were cooied to the temp buffer Org creates but maybe this is a timing issue or performs differently in different Emacs versions.  Clearly it was broken prior to your changes which I’ll also test.  If need be, we could use the code we have that reproduces the overlays as a way of finding wiki words to convert to html links but if Org can properly handle that aside from the hy: prefix, I would prefer for the export to do the least work.

rswgnu (approved):

Issue #669, opened 2025-02-15
For lisp modes, hywiki-buttonize-non-character-commands calls  (delete-char -1)


simendsjo:
I.e. deleting `a` in `(a)` will also delete `(`.

Here is a discussion on the Doom Emacs Discord: https://discord.com/channels/406534637242810369/1339703648980308030/1339703648980308030

matsl:
Hi @simendsjo 

Thanks for the report. Would it be possible for you to outline here steps to reproduce the problem? I don't have access to Discord and I fail to recreate the problem.

matsl:
Hi @simendsjo, I manged to recreate the problem. We are working on a fix. Thanks again.

simendsjo:
Thanks! Sorry, I wasn't around to file a more detailed report, but looks like you found it anyway. The bug has been haunting me for a while, and I had no idea it was caused by hyperbole :)

rswgnu:
Fixed in the pre-release a while back.  Closing.

PR #670, opened 2025-02-16
Verify issue with delete char

matsl:
# What

Verify delete char only deletes the char.

# Why

Reported issue. See #669


matsl:
@rswgnu Added a test case that verifies the reported behavior, Set as expected failed so can be merged and worked on to fix.

PR #680, opened 2025-02-28
Do not build and test for Emacs 27.2

matsl:
# What

Do not build 27.2 in CI.

# Why

We are bumping the required version of Emacs to 28.1 due to markdown
mode not being available for Emacs 27.2 any more.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


matsl:
@rswgnu If you remove 27.2 as expected check (You'll find it in settings/branch rules I think) this PR can be merged. (Maybe you need to retrigger the build or something too. I don't remember the details from lats time we removed a required check.)

rswgnu (approved):

PR #682, opened 2025-03-03
Do not include 27.2 for local docker based CI tests

matsl:
# What

Do not use version 27.2 on local docker based CI tests.

# Why

We are dropping support for 27.2.


rswgnu (approved):
Because markdown mode now requires Emacs 28.

PR #685, opened 2025-03-07
fix warning messages

matsl:
# What

- **Fix docstring and add test cases**
- **Public declare var**
- **Fix docstrings**

# Why

The goal is to have zero warnings.

### DIFF
--- a/test/hpath-tests.el
+++ b/test/hpath-tests.el
@@ -466,18 +466,26 @@ dir/subdir:
 (ert-deftest hpath--spaces-to-dashes-markup-anchor ()
   "Verify `hpath:spaces-to-dashes-markup-anchor'."
   (with-temp-buffer
+    ;; Normal case: Replace space wih dash.
     (should (string= "a-b-c" (hpath:spaces-to-dashes-markup-anchor "a b c")))
+    ;; Mixed case: No conversion.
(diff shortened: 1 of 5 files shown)


test/hpath-tests.el
```diff
@@ -463,6 +463,24 @@ dir/subdir:
       (should (= (line-number-at-pos) 1))
       (should (= (current-column) 0)))))
 
+(ert-deftest hpath--spaces-to-dashes-markup-anchor ()
+  "Verify `hpath:spaces-to-dashes-markup-anchor'."
+  (with-temp-buffer
+    (should (string= "a-b-c" (hpath:spaces-to-dashes-markup-anchor "a b c")))
```
rswgnu, line 469: Go through these test cases and make sure the comparison string is right in all cases.

rswgnu (changes requested):

test/hpath-tests.el
matsl: Well the tests succeed so either the tests are wrong or the function under test is wrong :smile: 

Anyway, I have expanded the test a little and added comments for why the test is there. I hope it makes things clear. If you think there is some comparison that is wrong or some test is missing please tell me and I'll fix it.

rswgnu (approved):

PR #686, opened 2025-03-17
Use hywiki-tests--command-execute

matsl:
# What

Replace use of with-hywiki-buttonize-hooks et al macros with
hywiki-tests--command-execute.

# Why

It is a more direct way to call the hooks setup by hywiki-mode rather
than explicitly calling each function as the macros did. This
hopefully makes it more close to how it actually works and makes the
test cases clearer.


test/hywiki-tests.el
```diff
@@ -65,6 +76,7 @@
           (goto-char 4)
           (action-key)
 	  (should (equal (cons 'page wikifile) (hywiki-get-referent "WikiWord"))))
+      (hywiki-mode 0)
```
rswgnu, line 79: Each test should leave hywiki-mode in the state it was before the test ran.  You can build a macro to wrap test code with that manages this.

rswgnu (approved):
Please handle the hywiki-mode comment when you have time.  I have approved so you can merge now and desl with that later.

test/hywiki-tests.el
matsl: Hmm... I see your point. Just afraid that it will complicate the test a bit ensuring the macro is debuggable plus add some extra overhead to the test. Will see what I can do. 

Until it is in place we will have to make sure to set the hywiki-mode to whatever we want it to be for our manual testing or usage too of course.

matsl: See #687. It was easier than expected and did not make the test cases unreadable as I first thought they might. PTAL.

PR #687, opened 2025-03-17
Add macro for restoring hywiki-mode to state before ert-test

matsl:
What

Add macro for restoring hywiki-mode to state before ert-test.  Use if
for all tests that manipulate the state of hywiki-mode.

Why

The state of hywiki-mode in the current Emacs should not be changed by
running a test case that manipulates hywiki-mode.

### DIFF
--- a/test/hywiki-tests.el
+++ b/test/hywiki-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell
 ;;
 ;; Orig-Date:    18-May-24 at 23:59:48
-;; Last-Mod:     18-Mar-25 at 16:19:07 by Mats Lidell
+;; Last-Mod:     19-Mar-25 at 17:00:58 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


test/hywiki-tests.el
```diff
@@ -36,6 +36,24 @@ This is for simulating the command loop."
     (command-execute cmd))
   (run-hooks 'post-command-hook))
 
+(defmacro hywiki-tests--preserve-hywiki-mode (&rest body)
+  "Restore hywiki-mode after running BODY."
+  (declare (indent 0) (debug t))
+  `(let ((current-hywiki-mode hywiki-mode))
+     (unwind-protect
+	 (progn ,@body)
+       (hywiki-mode (if current-hywiki-mode 1 0)))))
```
rswgnu, line 45: If the hywiki mode variable here matches the let variable value, then you should not in oke the hywiki mode function.  This prevents multiple runs of the hywiki mode hooks among other things.

rswgnu (changes requested):

test/hywiki-tests.el
matsl: Maybe. This should just be a very small optimization since calling hywiki-mode multiple times should not matter? If active or not calling it again should be a no op if already in the wanted state, right?

rswgnu:
Maybe you want to disable any hywiki-mode-hooks during test runs to ensure behavior is always consistent.-- BobOn Mar 18, 2025, at 5:36 AM, Mats Lidell ***@***.***> wrote:﻿



In test/hywiki-tests.el:
> @@ -36,6 +36,24 @@ This is for simulating the command loop."
     (command-execute cmd))
   (run-hooks 'post-command-hook))
 
+(defmacro hywiki-tests--preserve-hywiki-mode (&rest body)
+  "Restore hywiki-mode after running BODY."
+  (declare (indent 0) (debug t))
+  `(let ((current-hywiki-mode hywiki-mode))
+     (unwind-protect
+	 (progn ***@***.***)
+       (hywiki-mode (if current-hywiki-mode 1 0)))))

Maybe. This should just be a very small optimization since calling hywiki-mode multiple times should not matter? If active or not calling it again should be a no op if already in the wanted state, right?

matsl:
> Maybe you want to disable any hywiki-mode-hooks during test runs to ensure behavior is always consistent.

Can you elaborate why this would be necessary?

test/hywiki-tests.el
rswgnu: Again, unless you remove the mode hooks, they can execute any state changing code, so potentially far from a no-op.  I wouldn’t assume you control the env the tests are executed in, only what you code in the tests themselves.

rswgnu (changes requested):

matsl:
>  I wouldn’t assume you control the env the tests are executed in

I think we should only consider tests run through the make targets or in the CI/CD. In these cases we do control the environment. Does not the '--quick' command line option ensure that? Am I missing something? 

Test executed in a running Emacs is another issue. We do that as developers when developing code and tests, but the environment is not really possible to control fully. Emacs is a too customizable environment for that. Who knows what packages a user might have loaded etc. I don't think even trying is worth it.

rswgnu (approved):
Leaving to handle as you like.

matsl:
> Leaving to handle as you like.

Thanks. I'm open for revisiting this later.

PR #689, opened 2025-03-20
speed up docker runs by using local elpa volume

matsl:
# What

- **Include 30.1 for local docker based tests**
- **Use volume to keep elpa files between docker runs**
- **Add clean target for docker**

# Why

By keeping or sharing the required elpa packages between runs we can
speed up local docker runs considerable. No more waiting for download
from the package repos, except the first time. This makes it even more
useful to use the docker target for running tests since this ensures a
well defined environment and standard Emacs config.

# Note

Slides in the 30.1 update to the DOCKER_VERSIONS as well since we did
not add that after 30.1 was released.

Running "make docker-batch-test" takes now 54 secs on my machine. That
is full batch test for four versions of Emacs in less than one minute.

Similar for the simple "make docker targets='clean test'". That is
batch test for the development version of Emacs. It completes in less
that 12 seconds.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,6 +5,8 @@
     runs.
     (recompile-docker-elpa): Recompile elpa folder to ensure files are
     compiled with the same Emacs as used.
+    (docker-clean): Remove the elpa volume, useful for starting with a
+    fresh download of the required packages.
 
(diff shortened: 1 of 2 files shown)


rswgnu:
Wonderful.  — rsw

rswgnu (approved):

PR #692, opened 2025-03-31
Add support for running tests that are known to fail

matsl:
# What

Add support for running tests that are known to fail.

# Why

We want to be able to add failing test cases to master so that they
can be run from the command line, i.e. using make, but not affect the
regular CI builds.

# Note

Test that are known to fail can for normal runs be skipped by starting
the test with (skip-unless hy-test-failure). This skips the test on
regular execution of the tests either in batch or interactive mode.

For running these test cases add 'failure=t' on the command line. This
can be combined with 'test=<selector>' to further select which test
cases to run.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,7 @@
 2025-04-06  Mats Lidell  <matsl@gnu.org>
 
-* test/hy-test-helpers.el (hy-test-failure): Set to non-nil to activate
-    failing test cases. Used by HYPB_ERT_FAILURE.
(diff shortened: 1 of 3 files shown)


matsl:
@rswgnu What do you think of something like this for keeping failing test cases on the master branch?

rswgnu:
@matsl - Looks good except all Lisp booleans should end with '-flag', so please change 'hy-test-failure' to 'hy-test-run-failing-flag'.

matsl:
@rswgnu Updated. PTAL.

rswgnu (approved):

PR #693, opened 2025-04-06
change DSL to insert whole string before checking

matsl:
# What

- **Change DSL to verify after each operation rather than each
  character or deletion**
- **Check hywiki-word-is-p agrees with hywiki-word-at**

# Why

Trying another way to design the tests where the state after each
operation is checked that is easier to review and read.

# Note

The old tests are still valid with the current change but it allows
the tests to be a bit simplified which is also part of the PR.

### DIFF
--- a/test/hywiki-tests.el
+++ b/test/hywiki-tests.el
@@ -1382,27 +1382,32 @@ Each test case consists of cons cells with an operation and the expected
 state of the WikiWord being constructed.  Operations are either a string
 to be inserted, a number of chars to be deleted or a symbol p<number>
 for where to move point.  The expected state is either nil for not a
-wikiword or non-nil for a wikiword.  If equal to a string it is checked
-for match with the wikiword.  Movement of point is relative to point
-when the function is called."
+wikiword or non-nil for a wikiword.  The state is checked after all
+chars of the string is inserted.  If equal to a string it is checked for
+match with the wikiword.  Movement of point is relative to point when
+the function is called."
   (let ((origin (point)))
-    (should (listp test-case))           ; For traceability
+
+    ;; For traceability when looking through the list of should
+    ;; clauses in a failing test.
+    (should (listp test-case))
+
     (dolist (steps test-case)
       (let ((step (car steps))
             (vfy (cdr steps)))
         (cond ((stringp step)
                (dolist (ch (string-to-list step))
-                 (hywiki-tests--command-execute #'self-insert-command 1 ch)
-                 (save-excursion
-                   (goto-char (1- (point)))
-                   (hywiki-tests--verify-hywiki-word vfy))))
+                 (hywiki-tests--command-execute #'self-insert-command 1 ch))
+               (save-excursion
+                 (goto-char (1- (point)))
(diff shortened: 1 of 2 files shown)


test/hywiki-tests.el
```diff
@@ -1372,37 +1372,44 @@ If EXPECTED is a string also verify that the wikiword matches the
...
   "Run the TEST-CASE from point.
 Each test case consists of cons cells with an operation and the expected
 state of the WikiWord being constructed.  Operations are either a string
 to be inserted, a number of chars to be deleted or a symbol p<number>
 for where to move point.  The expected state is either nil for not a
-wikiword or non-nil for a wikiword.  If equal to a string it is checked
-for match with the wikiword.  Movement of point is relative to point
-when the function is called."
+wikiword or non-nil for a wikiword.  The state is checked after all
+chars of the string is inserted.  If equal to a string it is checked for
```
rswgnu, line 1388: 'are inserted' instead of 'is inserted'.

test/hywiki-tests.el
```diff
@@ -1372,37 +1372,44 @@ If EXPECTED is a string also verify that the wikiword matches the
 string."
   (if (not expected)
       (should-not (hywiki-word-at))
-    (if (stringp expected)
-        (should (string= expected (hywiki-word-at)))
-      (should (hywiki-word-at)))))
+    (let ((hywiki-found (hywiki-word-at)))
```
rswgnu, line 1375: Call this 'hywiki-word-found' for clarity.

test/hywiki-tests.el
```diff
@@ -1414,19 +1421,20 @@ when the function is called."
...
-    (("(H") ("iHo" . t) ("#") ("s" . "HiHo#s") (")" . "HiHo#s)")) ; Delimiter part of WikiWord. See below too.
-    (("(H") ("iHo" . t) ("#") ("s" . "HiHo#s") ("-" . "HiHo#s-") ("n" . "HiHo#s-n") (")" . "HiHo#s-n)"))
+    (("(HiHo#s" . "HiHo#s") (" " . "HiHo#s"))
+    (("(HiHo#s" . "HiHo#s") (")" . "HiHo#s)")) ; Delimiter part of WikiWord. See below too.
+    (("(HiHo#s" . "HiHo#s") ("-" . "HiHo#s-") ("n" . "HiHo#s-n") (")" . "HiHo#s-n)"))
     ;; Insert and delete between WikiWords
-    (("H") ("iHo" . t) (p3 . t) (" " . "Hi") (p4 . "Ho") (-1 . "HiHo"))
-    (("H") ("iho" . t) (p3 . t) (" " . "Hi") (p4) (-1 . "Hiho"))
+    (("HiHo" . t) (p3 . t) (" " . "Hi") (p4 . "Ho") (-1 . "HiHo"))
+    (("Hiho" . t) (p3 . t) (" " . "Hi") (p4) (-1 . "Hiho"))
```
rswgnu, line 1437: All these test cases are easier to read now, so this is a good thing.

rswgnu (changes requested):

rswgnu (approved):

PR #694, opened 2025-04-07
accept M RET on all input with vertico

matsl:
# What

- **Treat M-RET on input as by vertico when active**
- **Add hui-mouse-tests for hkey-alist actions**

# Why

Vertico uses the keybinding M-RET for vertico-exit-input for exiting
the minibuffer using the current minibuffer input. For example when
creating a new file where there is no candidate to select. An
alternative is to use the uparrow to go to the input prompt and use
RET there. That fails however for the case when there is no input and
the empty input is used to terminate. The user gets stuck in this case
and only way out is to use M-x vertico-exit-input or similar tricks.

This PR takes the approach to always respect vertico's view when
reading an argument.

The PR is also a start for verifying that the hkey-alist will call the
expected actions from a given state and not something else.

### DIFF
--- a/hui-mouse.el
+++ b/hui-mouse.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    04-Feb-89
-;; Last-Mod:     22-Feb-25 at 16:18:02 by Bob Weiner
+;; Last-Mod:     16-Mar-25 at 00:15:09 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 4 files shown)


rswgnu:
Interesting you decided to work on this now.  Certainly sounds useful.  -- rswOn Apr 7, 2025, at 6:16 PM, Mats Lidell ***@***.***> wrote:﻿
@matsl requested your review on: #694 accept M RET on all input with vertico.

hui-mouse.el
```diff
@@ -256,8 +256,7 @@ Its default value is `smart-scroll-down'.  To disable it, set it to
     ;; If in the minibuffer and reading an argument with vertico
     ;; run the vertico command on {M-RET} which accepts the first
     ;; line of minibuffer input, rather than any candidate.
-    ((and hargs:reading-type
```
rswgnu, line 259: I’m fine with everything here except the first line.  We’ll have to think about how this could impact other things, i.e. the above line was put there for a reason.

matsl: Can hargs:reading-type and vertico-mode be non-nil at the same time? If I remember correct they can not so the check on hargs:reading-type effectively ignores vertico reading the input.

Anyway I'm far from sure here. Could be a case where vertico is prompting for input where we want the action key to kick in instead? So I added a test case for hkey-alist so we can verify that different scenarios are handled properly. If we can identify the case we can add tests for that so we can verify that both scenarios works as expected.

rswgnu, line 259: Why don't we just add an appropriate check for whether the vertico-mode variable is true or not and dispatch as needed (defer to vertico)  rather than touching the hargs:reading-type clause?

rswgnu (changes requested):

rswgnu (commented):
I have manually integrated the needed code from here and will push it to the rsw branch today.  We can close this PR.

PR #698, opened 2025-04-14
reduce tarball size

matsl:
# What

- **Remove big png-file that is not used in any docs**
- **Remove another big png-file that is not used in any docs**
- **Remove eps files**

# Why

The hyperbole tarball is big. There are a few things that can be done
to reduce the size. 

One thing is to remove files from git that is not really needed. The
files removed from git in this PR are files that I can not see being
referenced in any way from Hyperbole.

There are more things that can be done but they are not addressed by
this PR.


rswgnu:
Are you sure the.eps files are not included in the .pdf file that TeX produces?  — rsw

matsl:
> Are you sure the.eps files are not included in the .pdf file that TeX produces? — rsw

I thinks so. Regenering the pdf using this branch works well and I can't see anything missing. The eps files themselves are generated using ImageMagic and dates back to 2016. I think they were manually created by converting the png to eps.

rswgnu:
> > Are you sure the.eps files are not included in the .pdf file that TeX produces? — rsw
> 
> I thinks so. Regenering the pdf using this branch works well and I can't see anything missing. The eps files themselves are generated using ImageMagic and dates back to 2016. I think they were manually created by converting the png to eps.

They must have been used when we used to generate PostScript versions of the manual but now that we use PDF, they must no longer be required.

rswgnu (approved):

PR #708, opened 2025-04-20
Remove byte compile warnings

matsl:
# What

Remove warnings.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2025-04-21  Mats Lidell  <matsl@gnu.org>
+
+* hywiki.el (ibtypes::pathname, ibtypes::pathname-line-and-column):
+    Declare functions from hpath.
+
+* hsys-org.el (hsys-org-link-at-p): Reinsert let binding of label-start-end.
+
+* hbut.el (ibut:insert-text): Remove shaddowed pcase.
+
 2025-04-20  Mats Lidell  <matsl@gnu.org>
 
 * test/hywiki-tests.el (hywiki-tests--filename-same-as-wiki-word): Add
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,7 +5,7 @@
 
 * hsys-org.el (hsys-org-link-at-p): Reinsert let binding of label-start-end.
 
-* hbut.el (ibut:insert-text): Remove shaddowed pcase.
+* hbut.el (ibut:insert-text): Remove shadowed pcase.
 
 2025-04-20  Mats Lidell  <matsl@gnu.org>
 
--- a/hbut.el
+++ b/hbut.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    18-Sep-91 at 02:57:09
-;; Last-Mod:     20-Apr-25 at 15:16:34 by Bob Weiner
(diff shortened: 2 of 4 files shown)


hsys-org.el
```diff
@@ -537,9 +537,10 @@ a non-Org buffer type."
 	    (save-match-data
 	      ;; If this Org link matches a potential HyWiki word, ignore it.
 	      (when (not (and (fboundp 'hywiki-word-at) (hywiki-word-at)))
-		(if (setq label-start-end (ibut:label-p t "[" "]" t))
-		    (cons (nth 1 label-start-end) (nth 2 label-start-end))
-		  t)))))))))
+                (let ((label-start-end (ibut:label-p t "[" "]" t)))
```
matsl, line 540: For some reason this change was lost so reapplying it.

hbut.el
```diff
@@ -2734,13 +2734,6 @@ Summary of operations based on inputs (name arg from \\='hbut:current attrs):
 	(if (eq arg3 0)
 	    (format "\"%s:%d\"" (hpath:shorten arg1) arg2)
 	  (format "\"%s:%d:%d\"" (hpath:shorten arg1) arg2 arg3))))
-      ('actypes::link-to-file
```
matsl, line 2737: There is a recent inserted actypes::link-to-file above that shadows this. See  line: 2703. I suppose that is the preferred version so removing this!?

hywiki.el
```diff
@@ -175,6 +175,8 @@
 (declare-function bookmark-completing-read "bookmark" (prompt &optional default))
 (declare-function bookmark-location "bookmark" (bookmark-name-or-record))
 (declare-function hsys-org-at-tags-p "hsys-org")
+(declare-function ibtypes::pathname "hpath")
```
matsl, line 178: hpath is not required so these functions create warnings.

I normally just add declare-function in these cases with the goal that one day we will refactor how modules relate and we can replace these declare-function with requires. (Or at least minimize the use of these for internal dependencies.)

ChangeLog
```diff
@@ -1,3 +1,12 @@
+2025-04-20  Mats Lidell  <matsl@gnu.org>
+
+* hywiki.el (ibtypes::pathname, ibtypes::pathname-line-and-column):
+    Declare functions from hpath.
+
+* hsys-org.el (hsys-org-link-at-p): Reinsert let binding of label-start-end.
+
+* hbut.el (ibut:insert-text): Remove shaddowed pcase.
```
rswgnu, line 8: shadowed

hsys-org.el
rswgnu: Why do we need this if we have the let* up above?

hywiki.el
rswgnu: That sounds good.  Requiring it creates a circular dependency, so I left it out, knowing you would catch it when you compiled.

rswgnu (changes requested):

hsys-org.el
matsl: Same reason as before. That let is in the "then" part of the if. This is in the "else" part. An alternative is to move the let for this var out one level so that it covers for both usages of the var but IMHO this is better by keeping it more local to the usage. It is just a temporary var used for the calculation.

rswgnu (approved):

PR #709, opened 2025-04-20
Add test - filename same as WikiWord.

matsl:
# What

Add test for filename being the same as a WikiWord.

# Why

This is a tricky edge case so want to avoid regressions.


rswgnu (approved):
I have some ideas on how to improve on this but lets get it in, I can make edits and then discuss with you.

PR #710, opened 2025-04-21
Add variation to WikiWord splitting

matsl:
# What

Add variation to WikiWord splitting.

# Why

We need to cover more ways a WikiWord can be modified.

### DIFF
--- a/test/hywiki-tests.el
+++ b/test/hywiki-tests.el
@@ -1343,6 +1343,10 @@ the function is called."
     (("HiHo" . t) (p3 . t) (" " . "Hi") (p4 . "Ho") (-1 . "HiHo"))
     (("Hiho" . t) (p3 . t) (" " . "Hi") (p4) (-1 . "Hiho"))
     (("hiHo") (p3) (" ") (p4 . "Ho") (-1))
+    ;; With double quotes
+    (("\"HiHo\"" . t) (p4 . t) (" " . "Hi") (p5 . "Ho") (-1 . "HiHo"))
+    (("\"Hiho\"" . t) (p4 . t) (" " . "Hi") (p5) (-1 . "Hiho"))
+    (("\"hiHo\"") (p4) (" ") (p5 . "Ho") (-1))
     )
   "List of test cases for WikiWords.")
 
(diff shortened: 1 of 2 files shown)


test/hywiki-tests.el
```diff
@@ -1339,9 +1339,10 @@ the function is called."
     (("(HiHo#s" . "HiHo#s") (" " . "HiHo#s"))
     (("(HiHo#s" . "HiHo#s") (")" . "HiHo#s")) ; Delimiter part of WikiWord. See below too.
     (("(HiHo#s" . "HiHo#s") ("-" . "HiHo#s-") ("n" . "HiHo#s-n") (")" . "HiHo#s-n"))
-    ;; Insert and delete between WikiWords
+    ;; Insert and delete between WikiWords and non WikiWords
     (("HiHo" . t) (p3 . t) (" " . "Hi") (p4 . "Ho") (-1 . "HiHo"))
     (("Hiho" . t) (p3 . t) (" " . "Hi") (p4) (-1 . "Hiho"))
+    (("hiHo") (p3) (" ") (p4 . "Ho") (-1))
```
rswgnu, line 1345: How about testing without any double quotes too,
E.g. HihoHiho and Hihop3, then split with a space.  Delimited words is a whole other area to build test cases around.

rswgnu (approved):
Approved but see my comment for more tests.

rswgnu (approved):

matsl:
@rswgnu Thanks for approving. I took the liberty to add a few more tests before merging. 

I'm not 100% sure about what is the expected behavior when it comes to WikiWords surrounded by double quotes. On the quote before the word the WikiWord is not identified but if on the quote after the word it is!? :shrug: Is that expected?

Anyway. Since the tests pass as written they document the current behavior and I felt that was OK for merging.

rswgnu:
-- BobOn Apr 23, 2025, at 6:16 AM, Mats Lidell ***@***.***> wrote:﻿
  @rswgnu Thanks for approving. I took the liberty to add a few more tests before merging.
I'm not 100% sure about what is the expected behavior when it comes to WikiWords surrounded by double quotes. On the quote before the word the WikiWord is not identified but if on the quote after the word it is!? 🤷 Is that expected?
Anyway. Since the tests pass as written they document the current behavior and I felt that was OK for merging.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>



matsl left a comment (rswgnu/hyperbole#710)
On the quote before the word the WikiWord is not identified but if on the quote after the word it is!? 🤷 Is that expected?Yes, point is always between characters and before the character its cursor appears over, so outside the string on an opening quote and inside on a trailing one.
Anyway. Since the tests pass as written they document the current behavior and I felt that was OK for merging.Great.Bob

PR #711, opened 2025-04-23
Add face verification for hywiki--word-face

matsl:
# What

Add face verification for hywiki--word-face.

Make a copy of the step check that does the same verification but adds
a check of the face of the matching WikiWord.

# Why

WikiWords should not only be recognized as text but should also have
the right face for visual feedback to the user. 

This combines the check of the text with a check that an overlay
exists with the right face.

# Note

The test does fails due to the face check. Is it thus put behind the
failure flag to not be executed in the CI.

### DIFF
--- a/test/hy-test-helpers.el
+++ b/test/hy-test-helpers.el
@@ -136,7 +136,7 @@ and the default WORD-LENGTH is 4."
         (cl-return t))))
 
 (defun hy-test-word-face-at-region (beg end)
-  "Non-nil if all chars in region [BEG, END] has `hywiki--word-face'."
+  "Non-nil if all chars in region [BEG, END] have `hywiki--word-face'."
   (interactive "r")
   (let (no-face)
     (while (and (< beg end) (not no-face))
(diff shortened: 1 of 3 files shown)


test/hy-test-helpers.el
```diff
@@ -127,5 +127,23 @@ and the default WORD-LENGTH is 4."
...
+(defun hy-test-word-face-at-point (&optional pos)
+  "Non-nil if `hywiki--word-face' at POS."
+  (unless pos
+    (setq pos (point)))
+  (cl-dolist (ol (overlays-at pos) nil)
+    (if (equal (overlay-get ol 'face) 'hywiki--word-face)
+        (cl-return t))))
+
+(defun hy-test-word-face-at-region (beg end)
+  "Non-nil if all chars in region [BEG, END] has `hywiki--word-face'."
```
rswgnu, line 139: ‘have’ instead of ‘has’.

rswgnu (approved):
Fix the one typo noted in the review and then approved.

PR #713, opened 2025-04-24
add missing wikipage in test

matsl:
# What

- **Add missing WikiPage required for successful test**
- **Refactor: Reuse hywiki-word-face-at-p**
- **Add skip-unless guard for face test**

# Why

The good news: The with face test was failing due to that the WikiPage
Hiho was not created. So now plain textual WikiWord verification works
as well when the overlay face is checked.  

The bad news: The test of course needs to be run interactively only to
work.

The even odder and possibly worse news: I have a simple test, that
when I run manually in my Emacs fails, but works in the CI or if I run
it in dockerized Emacs!? The test recipe is simple:
  1. Enter a WikiWord (WikiWord is highlighted)
  2. Add a space after (WikiWord is still highlighted)
  3. Remove the space (WikiWord is still highlighted)
  4. Add space again after. WikiWord is not highlighted.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,10 @@
 2025-04-26  Mats Lidell  <matsl@gnu.org>
 
+* hywiki.el (hywiki-word-face-at-p): Add optional pos to enable reuse.
+
+* test/hy-test-helpers.el (hy-test-word-face-at-region): Simplify by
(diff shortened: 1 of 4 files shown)


matsl:
@rswgnu PTAL The PR should be good to go in.

rswgnu (approved):

PR #714, opened 2025-04-24
Use ert test resources

matsl:
# What

Remove dependency on test file for test data.

# Why

Using a file use for testing as test data in test causes many odd test
failures when the file needs updates. That is bad. This replaces that
construct with ert test resources. That is a feature we have not been
using that we might want to explore further in the future.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,18 @@
+2025-04-25  Mats Lidell  <matsl@gnu.org>
+
+* test/MANIFEST: Add resource folders.
+
+* test/hmouse-drv-resources/test-data.el:
+  test/hmouse-drv-resources/TAGS:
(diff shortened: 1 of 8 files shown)


rswgnu (approved):
I don’t really understand why there are so many files or how this improves things but if you think it is important, I’ll defer to you on this.

matsl:
> I don’t really understand why there are so many files or how this improves things but if you think it is important, I’ll defer to you on this.

The number of files comes from that each test file has its own resource folder. That is some kind of duplication now but it is the default for ert-resources so good to start out like that. The files are small.

The improvement is that hy-test-helper.el is now not used as test data anymore. This means changes to that file will not break tests. We can edit it without thinking about if a new line will make a test break. It is better to have test data separated from the files actually used for running the tests.

PR #715, opened 2025-04-25
Use ert-simulate-keys

matsl:
# What

Refactor: Use ert-simulate-keys.

# Why

Replaces external dependency on with-simulated-input with similar
functionality from the ert-x package, which is part of Emacs.


rswgnu:
Love it.  I have hated that dependency.  — rsw

rswgnu (approved):

PR #727, opened 2025-05-12
refactor hkey execute

matsl:
# What

- **Base hkey-execute on hkey-actions**
- **Add tests for hkey-actions and hkey-execute**

# Why

We have tests for hkey-actions with hkey-alist to ensure the right
action/assist it selected in the given state or situation. 

This PR refactors hkey-execute to use hkey-actions instead of using a
copy of the same, or almost the same code.

# Note

There is some extra complication in the that hkey-debug call has been
moved into hkey-actions so that the loop vars does not have to be
returned from hkey-alist but can be local to hkey-actions.

There is also a small code duplication in that the point-marker both
needs to be present in hkey-actions and hkey-execute. 

There might be other ways to solve this including changing what
information hkey-debug should display. I have left is as it is for not
making an unnecessary large change.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2025-05-12  Mats Lidell  <matsl@gnu.org>
+
+* test/hmouse-drv-tests.el (hmouse-drv--hkey-actions): Test hkey-actions.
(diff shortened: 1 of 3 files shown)


rswgnu (approved):
Fix the one typo before committing.

PR #728, opened 2025-05-17
fix prefix arg issue and silence some warnings

matsl:
# What

- **Prefix arg with "p" spec is a number**
- **Silence warnings**

# Why

Fixing a problem with kotl-mode:add-prior-cell with prefix args. It
seems interactive spec "p" gives a prefix arg that is a number where
"P" gives the raw input. For C-u that means 4 vs '(4). So changing the
code to check for a number instead of a list.

Not sure which form is preferred but it matters for the check in the
function.

Some warnings are also fixed in a separate commit.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2025-05-18  Mats Lidell  <matsl@gnu.org>
+
+* kotl/kotl-mode.el (kotl-mode:add-prior-cell): Prefix arg with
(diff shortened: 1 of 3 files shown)


matsl:
@rswgnu I have modified this PR after your fix pushed to master. PTAL.

rswgnu (approved):

PR #729, opened 2025-05-18
Add interactive spec for actypes::org-link

matsl:
# What

Add interactive spec for actypes::org-link.

# Why

Allows org-link to be used as an ebut action. See bug#77838.

# Notes

Seems we did not fix this. 

On the other hand I can't see that org-link is available as a choice
if not the interactive spec is there!? So this error situation should
not really occur!? Or I might miss something.

Anyway, did not investigate this further. Can it be that it was
designed like this since there is less use of interactively adding an
org-link in a ebut!? Would not hyperbole links cover all cases?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,6 @@
 2025-05-31  Mats Lidell  <matsl@gnu.org>
 
-* hsys-org.el (org-link): Add interactive spec for promping user for an
+* hsys-org.el (org-link): Add interactive spec for prompting user for an
     org link.
 
 2025-05-27  Bob Weiner  <rsw@gnu.org>
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2025-05-31  Mats Lidell  <matsl@gnu.org>
+
+* hsys-org.el (org-link): Add interactive spec for promping user for an
+    org link.
+
 2025-05-27  Bob Weiner  <rsw@gnu.org>
 
 * hypb.el (hypb:in-string-p): Fix 'texinfo-mode' string not returning a list
--- a/hsys-org.el
+++ b/hsys-org.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     2-Jul-16 at 14:54:14
-;; Last-Mod:     26-May-25 at 00:15:37 by Bob Weiner
+;; Last-Mod:     31-May-25 at 15:35:55 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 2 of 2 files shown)


rswgnu:
We made this update instead, filtering out non-interactive actypes:

```
* hui.el (hui:actype): Fix to ensure only interactive actypes are returned
    so when an explict button is created, it can prompt for the parameters
    associated with the actype (otherwise, it will just get nil).
         (hui:link-possible-types): Update link to outline heading handler
    to use 'link-to-file' instead of 'link-to-string-match'.  Ignore the
    column number of point and for instance number of the #section, count
    only matches to heading lines, i.e. with 'outline-regexp'.
  hbut.el (ibut:insert-text): Add support for 'link-to-file'.
```

rswgnu:
So let's maybe work together on this one, make it better and then apply it.

matsl:
> We made this update instead, filtering out non-interactive actypes:

I see. I did not remember. So if I get it right we made it impossible to use org-link in the context where the user had the problem. (That is consistent with what I got when trying of it was fixed.)  So is this PR then needed?

matsl:
> So let's maybe work together on this one, make it better and then apply it.

Given that we made a fix to not let the user get into this situation, what do you think is good about allowing the user to add an org-link? I'm thinking a user might know an org link and would just want to use it. So it would make sense to support that with accepting entering a link manually.

If so, what do you think is missing in the PR? I'm thinking at least update to the docstring and tests would not hurt either while we are at this function.

rswgnu:
I think we need to think about the usage scenarios more before doing
anything.  What is the desire?  To link to an Org headline, an existing Org
link or some other type of entity.  Then we look at what mechanisms we have
already:  links to Org headlines or IDs, links to existing ibuttons, etc.
Then we can figure out what additional functionality might make sense.

-- rsw

rswgnu:
I need more clarity here.  It seems like you want to store the active text from an org link as the action of an explicit button.  Is that it?   I think that coukd be confusing, though I can see reasonable utility too.   Or do we want links to org buttons (we already have that with ilinks).

matsl:
> I need more clarity here. It seems like you want to store the active text from an org link as the action of an explicit button. Is that it? I think that coukd be confusing, though I can see reasonable utility too. Or do we want links to org buttons (we already have that with ilinks).

Yes, that is I believe, the net result of adding an interactive spec to the `defhact org-link`.  You could make user input slightly easier by inserting the brackets [[ and ]] around the user input to form a legit org-link.  

Whether supporting this makes sense is another issue but it would be in line with how other actypes work.

The originating bug report is [here](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=77838) for reference.

ChangeLog
```diff
@@ -1,3 +1,8 @@
+2025-05-31  Mats Lidell  <matsl@gnu.org>
+
+* hsys-org.el (org-link): Add interactive spec for promping user for an
```
rswgnu, line 3: Missing t in prompting.

rswgnu (approved):
I'll approve with the fix above but I have a more generalized idea to use completion that we will apply later.

PR #732, opened 2025-05-22
Add basic test of quote handling of hypb:in-string-p

matsl:
# What

Add test of basic quote handling in hypb:in-string-p

# Why

We need tests.

# Notes

This is just a start for testing `hypb:in-string-p` still I could not get it to work for markdown-mode. I tried it manually as well with the same result. Within triple backticks is it still not regarded as a string in markdown-mode!? Can you take a look?

I had so use python-mode for the python quotes so I assume the function is mode specific, i.e. the tests needs to be done in the proper mode. Am I right?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2025-05-23  Mats Lidell  <matsl@gnu.org>
+
+* test/hypb-tests.el (hypb--in-string-p): Add basic test of quote handling.
+
 2025-05-23  Bob Weiner  <rsw@gnu.org>
 
 * hywiki.el (hywiki-word-with-optional-suffix-regexp): Change #section
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,6 @@
-2025-05-23  Mats Lidell  <matsl@gnu.org>
+2025-05-24  Mats Lidell  <matsl@gnu.org>
+
+* test/hypb-tests.el (hypb--in-string-p--max-lines): Add test of max-lines.
 
 * test/hypb-tests.el (hypb--in-string-p): Add basic test of quote handling.
 
--- a/test/hypb-tests.el
+++ b/test/hypb-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:     5-Apr-21 at 18:53:10
-;; Last-Mod:     23-May-25 at 15:31:39 by Mats Lidell
+;; Last-Mod:     23-May-25 at 23:35:04 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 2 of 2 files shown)


rswgnu:
Hi Mats:  Looks like you should list out the set of test cases you see, let’s agree on that, then you can go off and develop them at your own pace.  I don’t want to add or review this until it handles a good number of the more complex scenarios since that is what we really need the tests for.  And you should have a good picture of what cases it handles and doesn’t handle in terms of even/odd strings before this string,  string delimiters on separate lines, where point is relative to blank lines within a string, etc.

matsl:
> Yes, mode-specific. 

Thanks for confirming.

> I meant to remove the markdown backtick handling since that handles formatted code rather than strings. So don’t worry about it.

OK. I'll remove that test then.

matsl:
Hi Bob,

> Hi Mats: Looks like you should list out the set of test cases you see, let’s agree on that, then you can go off and develop them at your own pace. 

I understand that independent test development can be good but it is also very time consuming. I have not the detailed knowledge that you have gained while developing this. It would provide valuable feedback to me if you too could list what test you see would be relevant.
 
> I don’t want to add or review this until it handles a good number of the more complex scenarios since that is what we really need the tests for.

The test is not only for verifying that complex scenarios work. It is also, and maybe even more important, for verifying that no regressions are introduces while we work on the code base. Hence the simple tests make a lot if sense to fast and with precision point out there is a problem.  

> And you should have a good picture of what cases it handles and doesn’t handle in terms of even/odd strings before this string, string delimiters on separate lines, where point is relative to blank lines within a string, etc.

I'm afraid I don't. The task involves re-engineering the code to find what "edge" cases there are that could require a test. It is here I find it would be time saving if you could hint what problems you have had to solve.

It is here test driven development shines. Every time you spot a problem you turn that into a test. Then you fix the problem and use the test to verify you got it right. With our approach do to the tests afterwards we are not sure the test actually covers the problem. The test is only verified by inspection and can be wrong ie not testing what we think it tests. (As have happened at times.) Sometimes we can roll back the code, while keeping the test to ensure it catches the problem. Most times that is not possible though.

matsl:
Hi Bob,

> Hi Mats: Looks like you should list out the set of test cases you see, let’s agree on that, then you can go off and develop them at your own pace. 

I understand that independent test development can be good but it is also very time consuming. I have not the detailed knowledge that you have gained while developing this. It would provide valuable feedback to me if you too could list what test you see would be relevant.
 
> I don’t want to add or review this until it handles a good number of the more complex scenarios since that is what we really need the tests for.

The test is not only for verifying that complex scenarios work. It is also, and maybe even more important, for verifying that no regressions are introduces while we work on the code base. Hence the simple tests make a lot if sense to fast and with precision point out there is a problem.  

> And you should have a good picture of what cases it handles and doesn’t handle in terms of even/odd strings before this string, string delimiters on separate lines, where point is relative to blank lines within a string, etc.

I'm afraid I don't. The task involves re-engineering the code to find what "edge" cases there are that could require a test. It is here I find it would be time saving if you could hint what problems you have had to solve.

It is here test driven development shines. Every time you spot a problem you turn that into a test. Then you fix the problem and use the test to verify you got it right. With our approach do to the tests afterwards we are not sure the test actually covers the problem. The test is only verified by inspection and can be wrong ie not testing what we think it tests. (As have happened at times.) Sometimes we can roll back the code, while keeping the test to ensure it catches the problem. Most times that is not possible though.

rswgnu:
> @rswgnu Added tests using the range-flag, It fails for texinfo-mode where there is no sequence return. I have thus marked the test as expected failed.
> 
> Plus a test of max-lines.

Thanks, will fix it.

PR #742, opened 2025-06-05
Add test for yanking in WikiWord

matsl:
# What

Add test for WikiWord that is yanked in together with following words.

# Why

The current behavior sets the WikiWord face on the whole yanked
text. It should only set the WikiWord face on the WikiWord. The test
is marked as expected failed.


rswgnu (approved):
Nice!

PR #743, opened 2025-06-07
protect org publish project alist from change by test

matsl:
# What

- **Do no let test change org-publish-project-alist**
- **Simplify target text and use quotes for subsection**

# Why

The test case, if run in an live Emacs, should not alter the users
org-publish-project-alist setting. 

The test data is also updated to be more precis i.e. there is no need
to let the WikiWord-links to be in headers but can rather be plain
body text.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,7 +2,7 @@
 
 * test/hywiki-tests.el
     (hywiki-tests--published-html-links-to-word-and-section): Do no let test
-    change org-publish-project-alist.
+    change org-publish-project-alist and update test.
 
 2025-06-06  Mats Lidell  <matsl@gnu.org>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2025-06-07  Mats Lidell  <matsl@gnu.org>
+
+* test/hywiki-tests.el
+    (hywiki-tests--published-html-links-to-word-and-section): Do no let test
+    change org-publish-project-alist.
+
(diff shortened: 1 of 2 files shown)


rswgnu:
Please fix typos in the PR description and first commit summary.

ChangeLog
```diff
@@ -1,3 +1,9 @@
+2025-06-07  Mats Lidell  <matsl@gnu.org>
+
+* test/hywiki-tests.el
+    (hywiki-tests--published-html-links-to-word-and-section): Do no let test
```
rswgnu, line 4: Change ‘no’ to ‘not’.

test/hywiki-tests.el
```diff
@@ -808,9 +809,9 @@ body B
             (save-buffer))
           (with-current-buffer (find-file-noselect wikipage)
             (insert "\
-* WikiWord
-* WikiWord#Asection
-* WikiWord#Bsection-subsection
+WikiWord
+WikiWord#Asection
+\"WikiWord#Bsection subsection\"
```
rswgnu, line 814: Add the case with the dash back in as a fourth test.  Will be good to have it.

rswgnu (changes requested):

test/hywiki-tests.el
matsl: > Add the case with the dash back in as a fourth test. Will be good to have it.

OK. Added it back as a Csection-subsection test.

rswgnu (approved):

PR #744, opened 2025-06-07
Verify creating a file for WikiWord highlights WikiWord

matsl:
# What

Verify creating a file for WikiWord highlights the instances of the WikiWord
already present in another WikiWord file.

# Why

WikiWords inserted before the WikiWord file is created should be
highlighted at the time the WikiWord file is created.

### DIFF
--- a/test/hywiki-tests.el
+++ b/test/hywiki-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell
 ;;
 ;; Orig-Date:    18-May-24 at 23:59:48
-;; Last-Mod:     22-Jun-25 at 22:32:29 by Bob Weiner
+;; Last-Mod:     24-Jun-25 at 09:36:49 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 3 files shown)


rswgnu (changes requested):
The test is to check whether creating a wikipage and saving it highlights existing instances of the associated wikiword already displayed in another window and buffer prior to the wikipage creation.  All buffers displayed with the wikiword should update and highlight at that point.

matsl:
> The test is to check whether creating a wikipage and saving it highlights existing instances of the associated wikiword already displayed in another window and buffer prior to the wikipage creation. All buffers displayed with the wikiword should update and highlight at that point.

I understand. The question I'm posing is whether using hywiki-add-page is OK or not. hywiki-add-page both creates a wikipage that is saved to disk. I'm just wondering if some more decoupled way to create and save the file is wanted. (Using hywiki-add-page is a convenient way to create a wikipage.)

rswgnu:
Yes, hywiki-add-page is the way to go.

test/hywiki-tests.el
```diff
@@ -1667,6 +1667,25 @@ Insert test in the middle of other text."
...
+           (hywiki-tests--with-face-test t)
+           wikiHo)
+      (unwind-protect
+          (progn
+            (hywiki-mode 1)
+            (with-current-buffer (find-file wikiHi)
+              (insert "Ho")
+              (save-buffer)
+              (setq wikiHo (cdr (hywiki-add-page "Ho")))
+              (goto-char 2)
```
rswgnu, line 1684: Need to have a wikiword reference in a window first, before creating the page for the wikiword.  Then add the wikiword page with any arbitrary content and save it.  The wikiword in the other window should highlight.  You may also want to have a reference to the wikiword in another frame to demonstrate that windows in multiple frames highlight as well.

rswgnu (changes requested):

rswgnu:
Let me know if you are updating this and want to get it approved or going another way.

rswgnu (approved):

PR #767, opened 2025-07-29
add kotl mode add cell tests

matsl:
# What

Add: 
 * kotl-mode--add-cell
 * kotl-mode--add-after-parent
 * kotl-mode--add-before-parent
 * kotl-mode--add-below-parent
 * kotl-mode--add-child

# Why

Functions are lacking tests.

# Note

## Using a temp buffer

These tests uses a new way of testing kotl-mode by combining `erase-buffer`
and `cl-flet` with `with-temp-buffer`. A local function is used that erases the
buffer and then calls `kotl-mode` which initializes the buffer. That way one
temp buffer can be used over and over again with greater speed and less
file creation and deletion. 

If there are no problems with using a temp buffer with `kotl-mode`
then I think that can be applied for more of the kotl-mode tests and
maybe even for more of our tests.

### DIFF
--- a/test/kotl-mode-tests.el
+++ b/test/kotl-mode-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell <matsl@gnu.org>
 ;;
 ;; Orig-Date:    18-May-21 at 22:14:10
-;; Last-Mod:     29-Jul-25 at 16:05:57 by Mats Lidell
+;; Last-Mod:     30-Jul-25 at 23:55:27 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


rswgnu:
All visible kotl cells are descendents of the too-level invisible 0 cell.  So the first child of the 0 cell is the #1 cell, first visible in the outline.

You can attach global properties like :title to the 0 cell.

-- Bob

rswgnu:
Please rewrite the Note and code if need be with the info I just provided and then I’ll review it.  Thanks.

-- Bob

test/kotl-mode-tests.el
```diff
@@ -1121,7 +1121,171 @@ marked with :ignore t")
...
+      (ert-info ("Error: Called with non positive argument")
+        (init)
+        (should-error (kotl-mode:add-before-parent 0)))
+      (ert-info ("If on the first level of the outline, insert cells at the start of the outline.")
+        (init)
+        (kotl-mode:beginning-of-buffer)
+        (should (= (kcell-view:level) 1))
+        (kotl-mode:add-before-parent 1 "start of outline")
+        ;; (should (kotl-mode:first-cell-p))) ;; <= EXPECTED
+        (should (string= (kcell-view:label (point)) "2"))) ;; FAIL: Actual.
```
matsl, line 1257: The cell should be added at the start of the outline. i.e. label=1 but is added after i.e. label=2.

PR #783, opened 2025-09-18
move out global text scale adjust

matsl:
# What

- **Use hycontrol-zoom-all-frames-{in,out} for all versions of Emacs**
- **Add zoom reset bound to X in the hycontrol menu**

# Why

global-text-scale-adjust is not fit for use from elisp. It is supposed
to be used interactively bound to a keymap which does not work well
with Hyperboles menu system.

Temporary roll back use of global-text-scale-adjust and just use
hycontrol-zoom-all-frames-* for all versions to solve the immediate
problem.

I noticed the problem above while trying out the zoom reset
functionality. Sorry for not testing more thoroughly to being
with. I'll look into a better long term solution for zooming in
harmony with global-text-scale-adjust.

The PR also adds zoom reset. It is bound to capital X since it is
close to capital Z.  This is under the assumption that doing a zoom
reset it likely to happen close to zooming in. WDYT?

### DIFF
--- a/hycontrol.el
+++ b/hycontrol.el
@@ -787,11 +787,9 @@ Screen bottom edge is adjusted based on `hycontrol-screen-bottom-offset'."
   "Change the font size of all faces by INCREMENT.
 Use `global-text-scale-adjust' if it exists.  If not fall back to
 implementation in hycontrol-zmfrm.el."
-  (if (fboundp #'global-text-scale-adjust)
-      (global-text-scale-adjust increment)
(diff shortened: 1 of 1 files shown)


rswgnu:
All sounds good.  I’ll review it now.-- BobOn Sep 18, 2025, at 4:05 PM, Mats Lidell ***@***.***> wrote:﻿
@matsl requested your review on: #783 move out global text scale adjust.

PR #787, opened 2025-09-21
Add actypes:activities for external package activities

matsl:
# What

Add actypes:activity for the external package activities.

# Why

It is convenient so switch between activities using global buttons or
per project ebuts.

# Notes

This is work in progress. 

WDYT?
 - Good idea? 
 - Should it be in a separate file, hsys-activities.el like other
   external packages?

Related: Would be nice to be able to use the assist key for getting
activity with the default state. But for ebuts as for ibuts the assist
key invokes help. With current scheme we would have to prompt user on
creation and have two types of buttons. With and without state
reset. Ideas?

This is a PR in Draft mode for the discussion.


rswgnu:
After looking at this, I am not sure we need it.  The activities package should provide the actions we need as autoloads.  Then we can create activities with <activities-new ...>  and restore them with <activities-resume ...> buttons.  If you have an idea of some other syntax to use to recognize activities names implicitly, let me know.

matsl:
> I would suggest first asking alphapapa to add the missing function that you need to the package.

Yes. I'm on it. :pray: 

I've been using my own functions with for action-buttons with activities for a couple of days and it works fine.

What do you think is the criteria for pointing users to use action-buttons rather than Hyperbole providing support through explicit buttons? What was previously available only through ebuts can now be done with labeled ibuts and  action-buttons (abuts?)!?

PR #788, opened 2025-09-21
Add actypes:combine-buttons

matsl:
# What

Prompts user for existing buttons to be executed in the order entered.

# Why

Sometimes it can be convenient to run a sequence of buttons. This
actypes allows for such a meta button.

### DIFF
--- a/hactypes.el
+++ b/hactypes.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    23-Sep-91 at 20:34:36
-;; Last-Mod:     21-Sep-25 at 23:47:17 by Mats Lidell
+;; Last-Mod:     22-Sep-25 at 23:59:47 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


rswgnu:
I think there is a good idea here. Let’s work to flesh it out together.  I think it should be more feneral than global buttons alone and maybe support per buffer named buttons too.

hactypes.el
```diff
@@ -792,6 +792,21 @@ SECTION is a string and can be just the leading part of a section heading."
...
     (recenter 0)))
 
+(defact combine-buttons (buts)
+  "Combine multiple buttons given by their labels into one button.
+Prefer buffer local buttons over global buttons."
+  (interactive
+   (list (completing-read-multiple "Button labels: "
+                                   (append (ebut:list)
+                                           (ibut:list)
+                                           (gbut:label-list))
```
rswgnu, line 802: Need to handle when the current buffer is the global button buffer and not append the same buttons twice.

hactypes.el
```diff
@@ -792,6 +792,21 @@ SECTION is a string and can be just the leading part of a section heading."
     (forward-line 1)
     (recenter 0)))
 
+(defact combine-buttons (buts)
```
rswgnu, line 795: How about naming this 'button-sequence' instead to match with 'key-sequence'?

hactypes.el
```diff
@@ -792,6 +792,21 @@ SECTION is a string and can be just the leading part of a section heading."
...
+  "Combine multiple buttons given by their labels into one button.
+Prefer buffer local buttons over global buttons."
+  (interactive
+   (list (completing-read-multiple "Button labels: "
+                                   (append (ebut:list)
+                                           (ibut:list)
+                                           (gbut:label-list))
+                                   nil t)))
+  (dolist (b buts)
+    (let ((but (hbut:get (hbut:label-to-key b))))
```
rswgnu, line 805: You want the let outside of the dolist so it doesn't push on the stack for each iteration.  Use setq within the loop.

rswgnu (changes requested):

PR #789, opened 2025-09-24
Check all caps on filename, reorder mode tests, in anchor search

matsl:
- Change setup referent-regexp to look only on filename part of path for all-caps.
- Extend all-caps to also include dash and underscore so that files like HY-NEWS are expected to have star-outlines.
- Change order of test so that markup is not masked by text-mode.

### DIFF
--- a/hpath.el
+++ b/hpath.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     1-Nov-91 at 00:44:23
-;; Last-Mod:     21-Sep-25 at 09:14:07 by Bob Weiner
+;; Last-Mod:     24-Sep-25 at 19:06:49 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


hpath.el
```diff
@@ -1677,19 +1677,21 @@ of the buffer."
 				    (anchor-name (hpath:dashes-to-spaces-markup-anchor anchor))
 				    (referent-regexp (format
 						      (cond ((or (derived-mode-p 'outline-mode) ;; Includes Org mode
-								 ;; Treat all caps filenames without suffix like outlines, e.g. README, INSTALL.
+								 ;; Treat all caps filenames without suffix like outlines,
+                                                                 ;; e.g. README, INSTALL, HY-NEWS, ...
 								 (and (hypb:buffer-file-name)
-								      (string-match-p "\\`[A-Z][A-Z0-9]+\\'" (hypb:buffer-file-name))))
+								      (string-match-p "\\`[A-Z0-9][A-Z0-9_-]*\\'"
```
rswgnu, line 1683: It used to be that - had to be the first char in the grouping to match.  Seems like your testing shows that the last char works too.   Ut have you checked compat across all our supported emacs versions?

PR #802, opened 2025-10-16
Add hywiki tests

matsl:
# What

Add more hywiki tests.

# Why

These tests covers the remaining hywiki functions that can be
activated by the hui-mini menu. It does not cover calls that goes
directly to other modules or set variables.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,7 +3,7 @@
 * test/hywiki-tests.el (hywiki-tests--consult-grep)
     (hywiki-tests--hkey-help, hywiki-tests--add-path-link-v2)
     (hywiki-tests--tags-view, hywiki-tests--directory-dired-edit)
-    (hywiki-tests--interactive-hywiki-mode-toogles): Add hywiki tests.
+    (hywiki-tests--interactive-hywiki-mode-toggles): Add hywiki tests.
 
 2025-10-13  Mats Lidell  <matsl@gnu.org>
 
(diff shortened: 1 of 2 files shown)


ChangeLog
```diff
@@ -1,3 +1,10 @@
+2025-10-16  Mats Lidell  <matsl@gnu.org>
+
+* test/hywiki-tests.el (hywiki-tests--consult-grep)
+    (hywiki-tests--hkey-help, hywiki-tests--add-path-link-v2)
+    (hywiki-tests--tags-view, hywiki-tests--directory-dired-edit)
+    (hywiki-tests--interactive-hywiki-mode-toogles): Add hywiki tests.
```
rswgnu: -toggles is misspelled.

test/hywiki-tests.el
```diff
@@ -1898,6 +1898,111 @@ Insert test in the middle of other text."
...
+          (progn
+            (hywiki-mode 1)
+            (with-current-buffer (find-file wikiHo)
+              (insert "123"))
+            (with-current-buffer (find-file wikiHi)
+              (hywiki-add-path-link "HoRef" wikiHo 3)
+              (should (string= (concat (file-name-nondirectory wikiHo) ":L1:C2")
+			       (cdr (hywiki-get-referent "HoRef"))))))
+        (hy-delete-files-and-buffers (list wikiHi wikiHo))
+        (hy-delete-dir-and-buffer hywiki-directory)))))
```
rswgnu: Separately, please add a check in the function that is called above so it refuses to delete the default value of hywiki-directory in case it is ever accidentally called outside the let of hywiki-directory.

rswgnu (changes requested):

test/hywiki-tests.el
matsl: You mean `hy-delete-dir-and-buffer`?

PR #806, opened 2025-10-29
First stab at publishing with org ids

matsl:
# What

Test case works for regular WikiWords and with sections.
Needs further work with tests and possibly refactoring.

Sharing for feedback.

I have today experimented with your hywiki folder. All files except one can be
exported with out any runtime error. The BB.org files causes an error which I
will look into.

I have not doubled checked all links but to my surprise I found rather few
proper links in the material. So not sure how I can use it to create more
types of tests from it. (Except for the BB.org file that is.)

Further updates.

I was able to find what input causes the error and also found two cases where
the generated output does not match the expected output. This is for plain
WikiWord generation, i.e. no sections involved.

It is documented in a test case in the mentioned PR.

Next step is to figure out what is happening and fix it.

### DIFF
--- a/hywiki.el
+++ b/hywiki.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    21-Apr-24 at 22:41:13
-;; Last-Mod:     29-Oct-25 at 23:22:22 by Mats Lidell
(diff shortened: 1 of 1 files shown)


rswgnu:
Please add a description of anything not yet working with the HyWiki html export command.

rswgnu (changes requested):

PR #809, opened 2025-11-07
Remove warnings

matsl:
# What

Silence two byte compilation warnings.

# Why

Byte compile should be as clean as possible.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,8 +1,8 @@
 2025-11-07  Mats Lidell  <matsl@gnu.org>
 
 * hsys-www.el (eww--dwim-expand-url): Declare function to silence warning.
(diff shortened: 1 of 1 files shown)


hsys-www.el
```diff
@@ -226,6 +228,7 @@ instead of `browse-url-new-window-flag'."
       (format "*eww-%s*" (url-host (url-generic-parse-url
                                     (eww--dwim-expand-url url)))))))
   (eww-mode)
+  (defvar url-allow-non-local-files)    ; Defined from Emacs 29.
```
rswgnu: Fix the Changelog comment that references a different variable name.

rswgnu (approved):

PR #819, opened 2025-11-27
Update hpath and add tests

matsl:
# What

Add tests for hpath. Correct some problems.

# Why

More tests are good and helps spot problems.

# Note

- First commit contains new hpath tests and a few updates to hpath itself.
- Second commit removes an old unused functions
- Third removes a warning

I added some comments in the PR  with the hope it will make the review easier.

### DIFF
--- a/hpath.el
+++ b/hpath.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     1-Nov-91 at 00:44:23
-;; Last-Mod:     27-Nov-25 at 21:31:34 by Mats Lidell
+;; Last-Mod:     27-Nov-25 at 21:46:58 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


hpath.el
```diff
@@ -131,22 +131,22 @@ The format is ${variable}.  Match grouping 1 is the name of the variable.")
...
+(declare-function hbut:label-to-key "hbut")
+(declare-function hmail:editor-p "hmail")
+(declare-function hypb:decode-url "hypb")
 (declare-function hypb:object-p "hypb")
-(declare-function Info-find-node "info")
-
+(declare-function kbd-key:key-series-to-events "hib-kbd")
 (declare-function kcell-view:indent "kcell-view")
 (declare-function klink:act "klink")
+(declare-function mm-mailcap-command "mm-decode")
```
matsl: Sorted the lines. Only new item is `hmail:editor-p.`

hpath.el
```diff
@@ -207,17 +207,16 @@ converted path with the value of `hpath:mswindows-mount-prefix'."
   (when (and (stringp path) (not (equal path "\\\\")))
     (setq path (hpath:mswindows-to-posix-separators path))
     (when (string-match hpath:mswindows-drive-regexp path)
-      (when (string-match hpath:mswindows-drive-regexp path)
```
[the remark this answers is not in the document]
matsl: The removed line, the when clause, is a duplicate of the previous line.

test/hpath-tests.el
```diff
@@ -266,14 +335,60 @@
               (hy-test-helpers:action-key-should-call-hpath:find (expand-file-name py-file hyperb:dir)))))
       (setenv "PYTHONPATH" old-python-path))))
 
+(ert-deftest hpath--absolute-to ()
+  "Verify `hpath:absolute-to'."
+  :expected-result :failed
```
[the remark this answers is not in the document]
matsl: To be sure what is the root cause within `hpath:call` would require some more work. Did not want to hold this PR because of that since even if I would describe it it will not fix it. That would require even more work. What is up for review here is that the tests are OK.

If you think the test is proper then we know there is a problem somewhere. If the test is wrong then that needs to be fixed.

hpath.el
```diff
@@ -1409,20 +1403,22 @@ If PATH is absolute, return it unchanged."
 (defun hpath:file-line-and-column (path-line-and-col)
   "Return list of parts from PATH-LINE-AND-COL string of format path:line:col.
 Parse out the parts and return a list, else nil."
-  (when (and (stringp path-line-and-col)
-	     (string-match hpath:section-line-and-column-regexp path-line-and-col))
```
matsl: The regexp `hpath:section-line-and-column-regexp` does not match a file without a line. The regexp can maybe be changed to account for that but I felt it was more local change to add support for the filename with no line as a default case in a cond.
The old function has code for returning a list with just a file **which hints at his being an expected result**. If that is wrong and the function should return nil, if there are no line/column info, then this fix is wrong and the test case needs to be changes as well of course.

hpath.el
```diff
@@ -2559,23 +2555,6 @@ from path or t."
 	    (or suffix t)
 	  return-path))))
 
-;; Next function from: 2006-11-02  Mats Lidell
```
matsl: This is mentioned in the ChangeLog as being removed from use everywhere so it can go.

PR #823, opened 2025-11-30
Add tests for hpath

matsl:
# What

Add tests for hpath.el.

# Why

More tests are good and helps spot problems.


matsl:
@rswgnu I forgot to fix the paths. A new update will come with hopefully better names.

PR #826, opened 2025-11-30
Generate manual for getting the published date right

matsl:
# What

Regenerated the manual using updated `before-save-hook`.

# Why

The hook did not account for the `@set UPDATED Month Day, Year`-field. The day is also now trimmed and does not include a leading zero.

### DIFF
--- a/man/hyperbole.texi
+++ b/man/hyperbole.texi
@@ -7,7 +7,7 @@
 @c Author:       Bob Weiner
 @c
 @c Orig-Date:     6-Nov-91 at 11:18:03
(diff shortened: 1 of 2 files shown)


man/hyperbole.texi
```diff
@@ -31,7 +31,7 @@
 @set txicodequotebacktick
 
 @set UPDATED November 30, 2025
```
matsl: @rswgnu Here is the problem. That field is not updated as it should and I know why. Bad regexp. Will be fixed tomorrow.

rswgnu (approved):

PR #846, opened 2026-02-01
Matsl fix save referent find use menu

matsl:
# What

Use the preserve hywiki-mode macro and utilize that in the test

# Why

The test creates a WikiPage so should do that in a temporary directory. The macro fixes that and we can avoid creating the file again in the test.


matsl:
OOPS! Made the PR to the master branch which is wrong. Will remove this PR:

PR #849, opened 2026-02-01
Make hywiki-tests--word-at more robust

matsl:
# What

Further improvement.

PR #850, opened 2026-02-02
new branch

matsl:
# What

Request to merge on master branch instead of rsw. Should be same as
#849 (after rebase with rsw at the time) with an update of the ChangeLog.

# Why

rsw branch was merged to master so work continues here.

Includes further improvements. More tests are working now.

I also rolled back to the previous version of `hywiki-tests--word-at`
and **changed**  it so that if `hywiki-word-at` finds a wikiword then if we test
highlighting `hywiki-highligted-word-at` must also find a wikiword. 
So that not missing highlighting indicates there is no wikiword.

That rollback causes a test to fail. I think this points to a problem
in hywiki rather than in the test, I can reproduce manually, 
so I have marked it as expected failed.

### DIFF
--- a/test/hywiki-tests.el
+++ b/test/hywiki-tests.el
@@ -1637,7 +1637,7 @@ comparison with expected overlays stable."
   "Choose what test to perform based on value of `hywiki-tests--with-face-test'."
   (let* ((range (hywiki-word-at :range)))
     (when (hywiki-get-referent (car range))
-      (when (and range hywiki-tests--with-face-test)
+      (when hywiki-tests--with-face-test
         (save-excursion
           (goto-char (round (/ (+ (cadr range) (caddr range)) 2.0)))
           (should (equal range (hywiki-highlighted-word-at :range)))))


test/hywiki-tests.el
```diff
@@ -1831,6 +1826,7 @@ face is verified during the change."
 
 (ert-deftest hywiki-tests--wikiword-identified-in-emacs-lisp-mode ()
   "Verify WikiWord is identified when surrounded by delimiters in `emacs-lisp-mode'."
+  :expected-result :failed
```
matsl: This is the test case that fails.

With hywiki-mode :all and WikiWord page created this in emacs-lisp-mode below 
```
;; [[WikiWord]]
```
There WikiWord is not highlighted but hywiki-word-at says it is a WikiWord. Is that the expected behavior? It seems wrong. It can be reproduced manually.

Ideally tests with or without face check should behave the same.

test/hywiki-tests.el
```diff
@@ -1640,12 +1634,14 @@ comparison with expected overlays stable."
...
-  "Test if there is a HyWikiWord reference at point with a referent.
-Choose what test to perform based on value of `hywiki-tests--with-face-test'."
-  (when (hywiki-referent-exists-p)
-    (if hywiki-tests--with-face-test
-	(hywiki-highlighted-word-at)
-      (hywiki-word-at))))
+  "Choose what test to perform based on value of `hywiki-tests--with-face-test'."
+  (let* ((range (hywiki-word-at :range)))
+    (when (hywiki-get-referent (car range))
+      (when (and range hywiki-tests--with-face-test)
```
matsl: The range check is not needed here since we must have a range for hywiki-get-referent to be non nil.

test/hywiki-tests.el
```diff
@@ -1640,12 +1634,14 @@ comparison with expected overlays stable."
...
+        (save-excursion
+          (goto-char (round (/ (+ (cadr range) (caddr range)) 2.0)))
+          (should (equal range (hywiki-highlighted-word-at :range)))))
```
matsl: Here we verify that the highlighted range is the same as what is returned by hywiki-word-at. Point is though adjusted to be within the range since hywiki-word-at does accept point to be outside of the range and still finds the wikiword.

Issue #883, opened 2026-03-03
hycontrol-move-frame: frame jumps to right edge if you attempt to move past left edge


Bodertz:
Same is true for attempting to move up when already at the top: the frame will jump to the bottom.

This is the case because `hycontrol-move-frame` does not ensure that X and Y are positive when passed to `set-frame-position`, and `set-frame-position` interprets negative X or negative Y as the right-most edge or top-most edge.

I have the following advice which fixes it, but it does that by replacing the entire function:

```elisp
(define-advice hycontrol-move-frame
    (:override (arrow pixels &rest r) ensure-positive-position)
  "Ensure the position passed to `set-frame-position' is positive."
  (let ((x (car (frame-position)))
        (y (cdr (frame-position))))
    (pcase arrow
      ('up    (set-frame-position nil x (max 0 (- y pixels))))
      ('down  (set-frame-position nil x (max 0 (+ y pixels))))
      ('left  (set-frame-position nil (max 0 (- x pixels)) y))
      ('right (set-frame-position nil (max 0 (+ x pixels)) y)))))
```

This ensure that x and y can't go below 0.  I imagine that isn't needed for moving right or down, but I added it there anyway.

matsl:
Closing as fixed. See #887

PR #887, opened 2026-03-04
Only use non negative values to set-frame-position

matsl:
# What

Only use non negative values to set-frame-position.

* hycontrol.el (hycontrol-move-frame): Ensure move past left and upper
    boarder is not possible.
* test/hycontrol-tests.el (hycontrol-tests--move-frame): Test case.

# Why

Non negative values means position is on the other side of the
frame. With this change the frame stays at the upper of left border it
is moving towards.

Thanks @Bodertz for the report in issue #883

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,7 @@
 2026-03-05  Mats Lidell  <matsl@gnu.org>
 
 * hycontrol.el (hycontrol-move-frame): Ensure move past left and upper
-    boarder is not possible.
+    border is not possible.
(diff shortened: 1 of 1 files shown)


ChangeLog
```diff
@@ -1,3 +1,9 @@
+2026-03-05  Mats Lidell  <matsl@gnu.org>
+
+* hycontrol.el (hycontrol-move-frame): Ensure move past left and upper
+    boarder is not possible.
```
rswgnu: Spell it, border.

rswgnu (changes requested):

PR #888, opened 2026-03-04
Refactor Makefile and autoloads generation

matsl:
# What

Refactor Makefile and autoloads generation

* Makefile (.PHONY): Set the .PHONY declarations at each phony target.
    (HYPB_BYTE_COMP_INFO, HYPB_BYTE_COMP_FUNC): Use macros so %.elc rule
    can de defined once.
    (new-bin): Remove not needed dependency on autoloads.
    (eln): Remove not needed dependency on src and echo.
    (kotl/kotl-autoloads.el, hyperbole-autoloads.el): Use
    hload-path--make-directory-autoloads in both rules and make them
    consistent. Touch target after rebuild to ensure new timestamp.
    (bin): Run remove-elc before src target.
    (TOUCH): Add touch(1) for setting new timestamp.

# Why

The phony targets were added a bit randomly so this makes it in better
order. While at it I simplified autoloads and made autoloads if
triggered always to update the timestamp.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,16 @@
 2026-03-04  Mats Lidell  <matsl@gnu.org>
 
+* Makefile (.PHONY): Set the .PHONY declarations at each phony target.
+    (HYPB_BYTE_COMP_INFO, HYPB_BYTE_COMP_FUNC): Use macros so %.elc rule
+    can de defined once.
+    (new-bin): Remove not needed dependency on autoloads.
+    (eln): Remove not needed dependency on src and echo.
+    (kotl/kotl-autoloads.el, hyperbole-autoloads.el): Use
+    hload-path--make-directory-autoloads in both rules and make them
+    consistent. Touch target after rebuild to ensure new timestamp.
+    (bin): Run remove-elc before src target.
+    (TOUCH): Add touch(1) for setting new timestamp.
+
 * test/hargs-tests.el (hargs-get-verify-extension-characters): Make test run
     in batch.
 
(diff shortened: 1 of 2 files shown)


ChangeLog
```diff
@@ -1,5 +1,16 @@
 2026-03-04  Mats Lidell  <matsl@gnu.org>
 
+* Makefile (.PHONY): Set the .PHONY declarations at each phony target.
+    (HYPB_BYTE_COMP_INFO, HYPB_BYTE_COMP_FUNC): Use macros so %.elc rule
```
rswgnu: Help me understand the need for all the dot phony lines. It's a lot of noise to add to the file visually so it should have a lot of value if we add it.

matsl: Each phony target should be defined. In small Makefiles it is customary to do that at one place. In bigger Makefiles, such as ours, that pauses a problem since it is easy to miss a declaration. And the declarations gets added at ad hoc places as we currently have. With this change I adopt the policy to declare each phony target as phony close to where the target is defined. That makes it clear what target it is and easier to maintain.

matsl:
> Just wondering if empirically you can point to any problem that you saw that was solved by this update. Yes, someone could add a file named ‘clean’ or ‘bin’ and that would interfere with a target name without a .phony, but have you seen that happen?

No, it is not based on that we have been hit by this problem. 
 
> I will approve but this is what I am trying to understand.

Thanks.

While doing the other improvements I noticed some phony targets were missing. While adding them I decided to go over the whole file to make the placement of the phony line more logical. The intent is to make the file easier to maintain when adding or removing targets.

PR #932, opened 2026-04-08
Add hyrolo test case for moving in fgrep result

matsl:
# What

Add hyrolo test case for moving in fgrep result.

* test/hyrolo-tests.el (hyrolo-tests--fgrep-move-test): Add test case
for moving in *HyRolo* buffer after fgrep search. Test is set as
expected failed since there is a problem with moving over hidden
sections.

# Why

More tests and shows a possible bug.


rswgnu:
I found the movement problem.  I'll deploy your PR, remove the expected fail, apply my movement fix and then push to master with your test working.

rswgnu (approved):

matsl:
> Why don't you put the tests that fail at the beginning of the test sequence so when edebugging you don't have to step through all the known good tests to see one that is failing.

When debugging I just comment out the test that works so only the failing test is remaining.  It works quite well when using ert-info to wrap each test so it is easy to comment out.

I think it is good to have a set of working tests that sort of proves that the test setup is doing the right thing. An alternative could be to add the failing test as a separate ert-test that we later merge with the ert-test with the good test cases!? That would make it possible to have more than one failing test. WDYT?

PR #942, opened 2026-04-24
Add test for hyrolo-add

matsl:
# What

Add test for hyrolo-add

* test/hyrolo-tests.el (hyrolo-tests--add--error-cases):
    (hyrolo-tests--add-item-to-new-file): Add tests.
    (hyrolo-tests--add-items-at-multiple-levels):
    (hyrolo-tests--add-items-interactive): Fix quotes in docstring.

# Why

More test coverage for hyrolo.

# Note

Testing write and read properties of files using root as user is
problematic since it behaves different for root than for a regular
user. This PR solved it by mocking the checks. 

This shows a bigger problem. In the best of worlds we should be
running the tests as a regular user. That will require some work to
implement in our current setup.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2026-05-10  Mats Lidell  <matsl@gnu.org>
+
+* test/hyrolo-tests.el (hyrolo-tests--add--error-cases):
+    (hyrolo-tests--add-item-to-new-file): Add tests.
+    (hyrolo-tests--add-items-at-multiple-levels):
+    (hyrolo-tests--add-items-interactive): Fix quotes in docstring.
+
 2026-05-10  Bob Weiner  <rsw@gnu.org>
 
 * hyrolo.el (hyrolo-set-date): Change edit date to replace date without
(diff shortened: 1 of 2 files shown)


rswgnu:
If the first file in hyrolo-file-list does not exist, the file should be created and the entry added when hyrolo-add is called.  Hyperbole should initialize that variable to non-null if needed as well.

matsl:
> If the first file in hyrolo-file-list does not exist, the file should be created and the entry added when hyrolo-add is called. Hyperbole should initialize that variable to non-null if needed as well.

That is a good test to add and it breaks my fix. Good catch. I'll add that test and fix hyrolo-add.

Since we have a defcustom for hyrolo-file-list that initializes it, that should already be covered by that!?

But coming back to the test. **What should be the behavior when adding an item to a file that does not exist but is not the first file in hyrolo-file-list? Shall it too be created or shall hyrolo-add signal an error?**

Related: Do we need to check better that the first file in hyrolo-file-list is a "user-specific hyrolo file" and not just the first file that hyrolo-get-file-list returns?

matsl:
@rswgnu  Updated the test to treat adding an item to a new file as a legit operation. Added a new test for that. PTAL.

rswgnu (approved):

PR #954, opened 2026-05-14
Move up code so symbols are defined for later use in this file

matsl:
# What

Move up code so symbols are defined for later use in this file.

# Why

This prepares for removing loading of autoloads as part of byte
compilation.


matsl:
@rswgnu Does the moved code call any functions during load time? Is it not only definitions that will be called later when the rest of the file already has been loaded?

rswgnu (approved):

matsl:
@rswgnu Thanks for approving. I'll add a ChangeLog item about the code move too before merging.

PR #968, opened 2026-06-05
prepare for emacs 31

matsl:
# What

- **Handle functions marked as obsolete in Emacs 31**
- **Add type and group to defcustom**
- **Declare used functions**

# Why

Remove warnings.

### DIFF
--- a/hypb.el
+++ b/hypb.el
@@ -63,6 +63,7 @@
 (defvar hsys-org-enable-smart-keys)     ; "hsys-org.el"
 (declare-function kview:char-invisible-p "kotl/kview")
 (declare-function smart-eobp "hui-mouse")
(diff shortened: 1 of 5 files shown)


hypb.el
```diff
@@ -63,6 +63,7 @@
 (defvar hsys-org-enable-smart-keys)     ; "hsys-org.el"
 (declare-function kview:char-invisible-p "kotl/kview")
 (declare-function smart-eobp "hui-mouse")
+(declare-function smart-ancestor-tag-files "hmouse-tag")
```
rswgnu, line 65: Add above prior line so alphabetized.

rswgnu (changes requested):

rswgnu (approved):

PR #973, opened 2026-06-09
Use vertico-mode command to control if mode is active

matsl:
# What

* test/hywiki-tests.el (hywiki-tests--referent-test): Disable
vertico-mode during the test if active.
(hywiki-tests--save-referent-command-use-menu):
(hywiki-tests--save-referent-find-use-menu): Do not set vertico-mode
in let.
(hywiki-tests--save-referent-info-node-use-menu): Change expected
results to match result without vertico-mode active. Test now passes
both when in docker and native.

# Why

The vertico-mode variable is not used by the mode to control if mode
is active or not. It is just for providing the state of the mode. This
PR uses vertico-mode command instead for disabling vertico for the
relevant test cases.

This also resolved a test case,
hywiki-tests--save-referent-info-node-use-menu, that had the expected
result if vertico-mode was active. Now that test works both in native
and docker Emacs.

# Note

There can be issues with completion using different completion
frameworks that can trip us. When I manually test creating a
WikiReferent the name of the WikiWord is not inserted. However it is
defined so if I add the WikiWord manually it has the proper
definition. So that needs to be investigated further.

I think there are other places where we use let to set vertico-mode
but, as described above, that is not the proper way to control
vertico-mode I think.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,14 @@
+2026-06-10  Mats Lidell  <matsl@gnu.org>
+
+* test/hywiki-tests.el (hywiki-tests--referent-test): Disable
+    vertico-mode during the test if active.
+    (hywiki-tests--save-referent-command-use-menu):
+    (hywiki-tests--save-referent-find-use-menu): Do not set vertico-mode
+    in let.
+    (hywiki-tests--save-referent-info-node-use-menu): Change expected
+    results to match result without vertico-mode active. Test now passes
+    both when in docker and native.
+
 2026-06-09  Mats Lidell  <matsl@gnu.org>
 
 * test/hbut-tests.el (hbut-tests--ibut-at-p-identifies-a-remote-pathname):
--- a/test/hywiki-tests.el
+++ b/test/hywiki-tests.el
@@ -34,17 +34,21 @@ The template runs the PREPARE body, and that must add the HyWikiWord
 named WikiReferent with a non-page referent type."
   (declare (indent 0) (debug t))
   `(let* ((hsys-consult-flag nil)
-	  (vertico-mode 0)
+	  (local-vertico-mode (bound-and-true-p vertico-mode))
 	  (hywiki-directory (make-temp-file "hywiki" t))
 	  (wiki-word-non-page "WikiReferent")
           (mode-require-final-newline nil))
      (unwind-protect
          (save-excursion
+           (when local-vertico-mode
+             (vertico-mode -1))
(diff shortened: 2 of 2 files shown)


rswgnu:
I have tested the letter of vertico-mode and it has worked a ross versions.  Maybe somwthing has changed.  Please add bug todos wherever you find any.Thanks,-- BobOn Jun 9, 2026, at 6:15 PM, Mats Lidell ***@***.***> wrote:﻿

@matsl requested your review on: rswgnu/hyperbole#973 Use vertico-mode command to control if mode is active.

matsl:
> I have tested the letter of vertico-mode and it has worked a ross versions.  Maybe somwthing has changed.  Please add bug todos wherever you find any.Thanks,

Not sure anything has changed. This is the doc string for vertico-mode that clearly states the command should be used:
```
vertico-mode is a variable defined in ‘vertico.el’.

Its value is t
Original value was nil

Non-nil if Vertico mode is enabled.
See the ‘vertico-mode’ command
for a description of this minor mode.
Setting this variable directly does not take effect;
either customize it (see the info node ‘Easy Customization’)
or call the function ‘vertico-mode’.

  You can customize this variable.
  Setting this variable directly with ‘setq’ may not take effect;
  use either customize or ‘setopt’, or call ‘custom-set-minor-mode’.
```

I'm thinking that since only one the hywiki test case was really punished by vertico-mode being active, and that worked as it was defined if you run it in an environment where vertico-mode was present, it could easily go unnoticed depending on in which environment you run the interactive tests.

Anyway this PR is only about the changing the tests and does not alter anything else.

rswgnu (approved):

rswgnu:
Maybe the let works differently than a setq but I have approved your PRs
anyway.  We obviously want the tests to work and it looks like you found
why at least one was always causing trouble.

Bob

matsl:
> Maybe the let works differently than a setq but I have approved your PRs anyway.

Thanks.

I think the docstring for vertico-mode tries to say that there are more to turning vertico-mode off than to set the variable. Hence the advice to use customize or setopt. I have not looked into details but we can come to that eventually. 

> We obviously want the tests to work and it looks like you found why at least one was always causing trouble.

Yes. I found that turning off vertico-mode using the command worked for the problematic test case. Then I realized we already tried to do that in the base macro used in the test but only by setting the variable there. So not really working. So I moved the change into the macro instead of just doing that in the specific test.

PR #990, opened 2026-06-25
Extend tests of hywiki-display-referent-type

matsl:
# What

Extend tests of hywiki-display-referent-type

* test/hywiki-tests.el (hywiki-display-hywiki-test): Add helper
display function.
(hywiki-tests--display-referent-type): Verify right display function
is called and hbut:current attributes are set.

# Why

The call to the display function was not covered directly.

### DIFF
--- a/test/hywiki-tests.el
+++ b/test/hywiki-tests.el
@@ -3,7 +3,7 @@
 ;; Author:       Mats Lidell
 ;;
 ;; Orig-Date:    18-May-24 at 23:59:48
-;; Last-Mod:     25-Jun-26 at 13:00:41 by Bob Weiner
+;; Last-Mod:     26-Jun-26 at 00:03:34 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


rswgnu (changes requested):
I'm thinking if you use a link-to-bookmark referent type, the display test verifies that the buffer and point match the ones linked to, and so on for all of the referent types.

matsl:
> I'm thinking if you use a link-to-bookmark referent type, the display test verifies that the buffer and point match the ones linked to, and so on for all of the referent types.

I'll respond on that tomorrow. 

This PR is a general test that the function `hywiki-display-referent-type` handles error and builds the right display function plus setting the attributes. That does not have to be verified for each type. Not claiming that the test handles everything but is a good test of the function.

matsl:
> I'll respond on that tomorrow.

It took longer than one day and I replied in a separate mail. What it boiled down to is that this is a unit test of the `hywiki-display-referent-type` function and does not replace testing the functions performing the display action. That will be covered in future planned work.

test/hywiki-tests.el
```diff
@@ -2071,11 +2071,28 @@ expected result."
 		       (list #'hywiki-tags-view t nil bn)))
         (should (= (line-number-at-pos) 3))))))
 
+(defun hywiki-display-hywiki-test (wikiword value)
```
rswgnu, line 2074: Rename to hywiki-tests--format-referent-value so does not get in the regular HyWiki function namspace.

rswgnu (changes requested):

rswgnu:
> > I'll respond on that tomorrow.
> 
> 
> 
> It took longer than one day and I replied in a separate mail. What it boiled down to is that this is a unit test of the `hywiki-display-referent-type` function and does not replace testing the functions performing the display action. That will be covered in future planned work.

Ok.

test/hywiki-tests.el
matsl: `hywiki-display-referent-type` builds the display function name by appending referent type to the prefix 'hywiki-display-', that is hardcoded,  so it must be within the namespace. Can this function cause any problems?

rswgnu: How is building your own referent type to check the mechanics of the display-referent-type code any different than just choosing an existing type and testing against that?  Wouldn't that work just as well or is yours somehow more generic?

matsl: @rswgnu  I could probably get away without using the helper function if I mock one of the regular display functions and use that in the test. I find that solution less ideal thought than the suggested one.

rswgnu: Alright, let's try it your initial way.

rswgnu (approved):

test/hywiki-tests.el
matsl: > How is building your own referent type to check the mechanics of the display-referent-type code any different than just choosing an existing type and testing against that? Wouldn't that work just as well or is yours somehow more generic?

It would work as well but I don't see the harm in the test defun. Anyway. Thanks for approving. It can always be tweak later as we go.

PR #1007, opened 2026-07-11
Make install check aware of alternative package managers

matsl:
# What

Make install check aware of alternative package managers and avoid doing
package-install for those. Adding a `defcustom` so users can choose to get an
automated install if using the standard `package.el.`

# Why

There are multiple package managers so we can't assume users use
package.el. If it is not used then using it to install a package would likely
be an error an could damage a users setup or at least not be what the user
expects.

Supporting different package managers for an automatic install is beyond
where we want to go. 

# Note

The idea is to use `hypb:ensure-dependency` in all places where we today to `package-install-p`. That would be

```
$ grep package-installed-p *.el
hsys-consult.el:  (unless (package-installed-p 'consult)
hsys-consult.el:  (unless (package-installed-p 'org-roam)
hui-treemacs.el:    (and (package-installed-p 'treemacs)
hypb.el:  (unless (package-installed-p 'interaction-log)
hypb.el:           (package-installed-p package))
hyrolo.el:    (unless (package-installed-p 'markdown-mode)
hyrolo.el:  (unless (package-installed-p 'markdown-mode)
```
Before implementing that could you, @rswgnu,  take a look at what you think of
using this idea? If that makes sense I can complete this Draft.

### DIFF
--- a/hypb.el
+++ b/hypb.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     6-Oct-91 at 03:42:38
-;; Last-Mod:     10-Jul-26 at 17:16:28 by Bob Weiner
+;; Last-Mod:     15-Jul-26 at 21:01:28 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -91,6 +91,11 @@ preceded by a backslash are ignored.  The wrap regexp has group 1 that
 matches the beginning of line or a backslash character.  Group 2 matches the
 delimiter."
   :type 'sexp
+  :group 'hyperbole-commands)
+
+(defcustom hypb:ask-to-install-package-flag t
+  "Non-nil if user shall be queried before installing a package."
+  :type 'boolean
   :group 'hyperbole-commands)
 
 (defvar hypb:mail-address-mode-list
@@ -181,8 +186,7 @@ Uses the newer \"nadvice\" elisp library, not \"advice\"."
 This displays a clean log of Emacs keys used and commands executed."
   (interactive)
   ;; Ensure package is installed
-  (unless (package-installed-p 'interaction-log)
-    (package-install 'interaction-log))
+  (hypb:require-package 'interaction-log)
 
   ;; Ensure interaction-log-mode is disabled to removes its command
   ;; hooks which are replaced below.
@@ -1169,23 +1173,45 @@ WINDOW pixelwise."
 	     (string-join unreadable-dirs "\n"))))
   dirs)
 
+(defun hypb:users-package-manager ()
+  "Return the package manager in use.
+Current supported package managers are `straight', `elpaca', and `package'."
+  (cond ((featurep 'straight) 'straight)
+        ((featurep 'elpaca) 'elpaca)
+        (t 'package)))
+
+(defun hypb:package-el-install (package)
+  "Install PACKAGE using the default package manager `package.el'.
+If `hypb:ask-to-install-package-flag' is non-nil query user if package should
+be installed."
+  (when (or (not hypb:ask-to-install-package-flag)
+            (y-or-n-p (format "Install `%s' to enable this feature? " package)))
+    (package-install package)
+    (require package)))
+
+(defun hypb:notify-manual-install-needed (package manager)
+  "Notify user that a manual install is needed for PACKAGE.
+Suggest user to use the package MANAGER."
+  (user-error "Package '%s' is required by this command.  Use your package manager '%s' to install it" package manager))
+
+(defun hypb:ensure-dependency (package)
+  "Ensure PACKAGE is available.
+Returns non-nil if the feature can be used."
+  (or (require package nil t)
+      (pcase (hypb:users-package-manager)
+        ('package (hypb:package-el-install package))
+        (manager (hypb:notify-manual-install-needed package manager)))))
+
 ;;;###autoload
 (defun hypb:require-package (package)
-  "Prompt user to install, if necessary, and require the Emacs PACKAGE-NAME.
-PACKAGE-NAME may be a symbol or a string."
+  "Prompt user to install, if necessary, and require the Emacs PACKAGE.
+PACKAGE may be a symbol or a string."
   (when (stringp package)
     (setq package (intern package)))
   (unless (symbolp package)
     (error "(hypb:require-package): package must be a symbol or string, not '%s'" package))
-  (unless (or
-           ;; Allow for alternative package managers like elpaca that don't
-           ;; show up with a `package-installed-p' check
-           (require package nil t)
-           (package-installed-p package))
-    (if (y-or-n-p (format "Install package `%s' required by this command?" package))
-	(package-install package)
-      (keyboard-quit)))
-  (require package))
+  (unless (hypb:ensure-dependency package)
+    (error "(hypb:require-package): package '%s' could not be found" package)))
 
 ;; Adapted from cl--do-remf in "cl-extra.el" but uses 'equal' for comparisons.
 ;;;###autoload
--- a/hypb.el
+++ b/hypb.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     6-Oct-91 at 03:42:38
-;; Last-Mod:     15-Jul-26 at 21:01:28 by Mats Lidell
+;; Last-Mod:     16-Jul-26 at 17:10:28 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
@@ -94,7 +94,7 @@ delimiter."
   :group 'hyperbole-commands)
 
 (defcustom hypb:ask-to-install-package-flag t
-  "Non-nil if user shall be queried before installing a package."
+  "Non-nil means `hypb:require-package' queries the user before installing it."
   :type 'boolean
   :group 'hyperbole-commands)
 
@@ -1190,13 +1190,14 @@ be installed."
     (require package)))
 
 (defun hypb:notify-manual-install-needed (package manager)
-  "Notify user that a manual install is needed for PACKAGE.
-Suggest user to use the package MANAGER."
+  "Prompt with an error that PACKAGE must be manually installed using MANAGER."
   (user-error "Package '%s' is required by this command.  Use your package manager '%s' to install it" package manager))
 
 (defun hypb:ensure-dependency (package)
-  "Ensure PACKAGE is available.
-Returns non-nil if the feature can be used."
+  "Return non-nil if PACKAGE is installed and ready for use.
+If PACKAGE is not installed and the package.el package manager is in
+use, an install attempt may be made.  When any other package manager is
+in use, an error is triggered prompting to manually install the PACKAGE."
   (or (require package nil t)
       (pcase (hypb:users-package-manager)
         ('package (hypb:package-el-install package))
--- a/hyrolo.el
+++ b/hyrolo.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:     7-Jun-89 at 22:08:29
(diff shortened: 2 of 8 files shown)


hypb.el
```diff
@@ -1169,23 +1174,45 @@ WINDOW pixelwise."
 	     (string-join unreadable-dirs "\n"))))
   dirs)
 
+(defun hypb:users-package-manager ()
+  "Return the package manager in use.
+Current supported package managers are `straight', `elpaca', and `package'."
+  (cond ((and (featurep 'straight) (fboundp 'straight-use-package)) 'straight)
```
rswgnu: Add quelpa too.

rswgnu (commented):
You are on the right track.  I would finish it off and add quelpa.

hypb.el
matsl: I read up on quelpa and it seems it is based on top of package.el. Users can mix installing using quelpa and package.el directly. So it is not clear a user would prefer to install the packages we dynamically depend on using quelpa. So we can let package.el handle a potential install for quelpa users. It will not cause any big problem for them if I understand things correctly.

hyrolo.el
```diff
@@ -3231,13 +3229,17 @@ files to search."
...
+      (pcase (hypb:users-package-manager)
+        ('package (when (or (not hypb:dependency-install-ask)
+                            (y-or-n-p (format "Install `%s' to enable this feature? " 'markdown-mode)))
+                    (unless (assoc "nongnu" package-archives)
+                      (setq package-archives (cl-copy-list package-archives))
+                      (add-to-list 'package-archives '("nongnu" . "https://elpa.nongnu.org/nongnu/")
+		                   t))
+                    (package-refresh-contents)
+                    (package-install 'markdown-mode)))
+        (manager (hypb:notify-manual-install-needed 'markdown-mode manager)))))
```
matsl: This is implemented on its own to allow the insertion of nongnu if that is missing only in the case where package.el is used. An alternative would be to do that either always  or if package.el is in use and then call `hypb:require-package`. Seeing it now it might more sense to do like that. WDYT?

hyrolo.el
```diff
@@ -38,8 +38,6 @@
 (require 'hsys-consult)
 (require 'hsys-org) ;; For `hsys-org-cycle-bob-file-list'
 (require 'hypb)     ;; For `hypb:mail-address-regexp' and `hypb:add-to-invisibility-spec'
-(eval-when-compile
-  `(hyrolo-install-markdown-mode))
```
matsl, line 42: I removed this. Does not seem like it does anything. Notice the backquote!? Am I missing something here?

rswgnu: Put a markdown file with a hyrolo entry in it and add this file to hyrolo-file-list.  Then start an emacs 28.2 with your changed hyperbole code and do a hyrolo-fgrep for that entry.  Then jump to the found entry with M-RET and you should be in markdown mode without an error.  If so, your change is good.

hypb.el
```diff
@@ -93,6 +93,11 @@ delimiter."
   :type 'sexp
   :group 'hyperbole-commands)
 
+(defcustom hypb:dependency-install-ask t
```
rswgnu: Let's rename this to 'hypb:ask-to-install-package-flag' before the name becomes entrenched as a defcustom.

rswgnu (dismissed):
Let's see together if we can reduce the number of these added support functions before merging this.  We can go through it together when you have time.

hyrolo.el
matsl: Tested this and it works as expected. Reported in separate mail.

matsl:
> Let's see together if we can reduce the number of these added support functions before merging this. We can go through it together when you have time.

What functions do you think makes sense to merge? Do you think the number of functions makes the code hard to understand?

rswgnu:
Just felt like a lot of functions when I first looked at it. I'll have to look at it more to figure out where we could merge in consultation with you.

matsl:
> Just felt like a lot of functions when I first looked at it. I'll have to look at it more to figure out where we could merge in consultation with you.

There are four new functions. But shall the number of functions matter? Yes, they are small but with a clear purpose. The ChangeLog is a bit hard to read but I tried to get all in there which might make it look more than it is!? Happy talk about it.

matsl:
@rswgnu I've rebased it so it is possible to merge it so it will be possible for you to add require of denote if you want.

hyrolo.el
matsl: Removed the insertion of nongnu based on that it is default in all Emacs versions we support.

rswgnu:
I don't understand your reference to denote here.  Could you explain or point me to some other message you are referring to.  I do see that the latest merge build here failed on 28.2.  Are you fixing that?

test/hypb-tests.el
```diff
@@ -157,6 +157,56 @@ See Emacs bug#74042 related to usage of texi2any."
...
+  (let* ((hypb:ask-to-install-package-flag nil)
+         (err (should-error (hypb:package-el-install 'hypb-dummy-package-name) :type 'error)))
+    ;;FIXME: Emacs-28 includes a '-' when there is no version
+    ;;number. Can be removed when we drop support for Emacs 28.
+    (should (string-match-p (rx "Package " punct "hypb-dummy-package-name" (optional "-") punct " is unavailable") (cadr err)))))
+
+(ert-deftest hypb--notify-manual-install-needed ()
+  "Verify `hypb:notify-manual-install-needed'.
+Shall raise and user error with name of the package and the name of the
+package manger we think the user is using."
```
rswgnu: Fix typos and simplify last 2 lines of doc string: Raise a `need to install' package manager error.

matsl:
@rswgnu Removing the nongnu archive insertion that we talked about yesterday did not work. Reason was that in one call path the whole package-archives was locally set to nil. I suppose for not tampering with users package-archives. I had missed that. So I inserted that back but slightly different. As we said yesterday we possibly can remove this juggling since with nongnu being the default we could just try to install. However  I did not want to do that in this long and winding PR. Happy to work on that in a separate PR though.

I also added unit tests for the new functions.

Since the PR had many commits I squashed it and tried to make a better ChangeLog for the changes.

PTAL.

hypb.el
```diff
@@ -93,6 +93,11 @@ delimiter."
   :type 'sexp
   :group 'hyperbole-commands)
 
+(defcustom hypb:ask-to-install-package-flag t
+  "Non-nil if user shall be queried before installing a package."
```
rswgnu: "Non-nil means `hypb:require-package' queries the user before installing it when not found."

hypb.el
```diff
@@ -1169,23 +1173,45 @@ WINDOW pixelwise."
...
+If `hypb:ask-to-install-package-flag' is non-nil query user if package should
+be installed."
+  (when (or (not hypb:ask-to-install-package-flag)
+            (y-or-n-p (format "Install `%s' to enable this feature? " package)))
+    (package-install package)
+    (require package)))
+
+(defun hypb:notify-manual-install-needed (package manager)
+  "Notify user that a manual install is needed for PACKAGE.
+Suggest user to use the package MANAGER."
```
rswgnu: Replace whole doc string with:

"Prompt with an error that PACKAGE must be manually installed using MANAGER."

hypb.el
```diff
@@ -1169,23 +1173,45 @@ WINDOW pixelwise."
...
+  (user-error "Package '%s' is required by this command.  Use your package manager '%s' to install it" package manager))
+
+(defun hypb:ensure-dependency (package)
+  "Ensure PACKAGE is available.
+Returns non-nil if the feature can be used."
```
rswgnu: Replace whole doc string with:

"Return non-nil if PACKAGE is installed and ready for use.
If PACKAGE is not installed and the package.el package manager is in use, an install attempt may be made.  When any other package manager is in use, an error is triggered prompting to manually install the PACKAGE."

test/hy-test-dependencies.el
```diff
@@ -26,7 +28,10 @@
 (when (fboundp #'markdown-ts-mode)
   (add-to-list 'auto-mode-alist '("\\.md\\'" . markdown-mode)))
 
-(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/"))
+(setq package-pinned-packages
+      '((markdown-mode . "nongnu")))
+
+(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/") t)
```
rswgnu: What part of your tests requires melpa?  Does straight or elpaca require it?  Are you actually testing calls to melpa or mocking them?

rswgnu (changes requested):

rswgnu:
> As we said yesterday we possibly can remove this juggling since with
nongnu being the default we could just try to install. However I did not
want to do that in this long and winAs we said yesterday we possibly can
remove this juggling since with nongnu being the default we could just try
to install. However I did not want to do that in this long and winding PR.ding
PR.

If you are just going to rely on default behavior, I would deal with that
now and then have your tests cover it, so all of these updates are tested
in one go.  I have sent my other review comments.  I'll try to wrap this up
with you today.

test/hy-test-dependencies.el
matsl: No, that is old. el-mock requires it. See the diff. We had it before.

I did change it so melpa would be last in the list but was really not planning in the end to have it as part of the PR. I missed to revert it when I was cleaning the PR. Thanks for the catch. I'll revert.

The change was triggering some thoughts about how we should deal with our testing setup which is complicated.

test/hy-test-dependencies.el
```diff
@@ -53,5 +58,9 @@
 ;; Ignore publishing-related errors that don't affect the tests.
 (ignore-errors (hsys-org-log-and-fix-version))
 
+;; Allow dynamic loading with no questions asked of dependencies
+;; i.e. markdown-mode
+(setq hypb:ask-to-install-package-flag nil)
```
rswgnu, line 90: Do you think we should make this 't' by default since installing a package can be a heavy-weight operation?

matsl: This is for allowing CI to run without prompting. That means when markdown-mode is required in CI it will be installed in the CI environment. We don't have any other tests that drags in dynamic dependencies what I know. But maybe we want that in the future.

With this on mind I'm thinking we need to look over our test strategy. With now over 700 tests that we want to run efficient in Github, local docker, in personal development Emacs we have a pretty entangled system that might not serve us well in all cases. (This is after the release.)

matsl:
> If you are just going to rely on default behavior, I would deal with that now and then have your tests cover it,

Well, I don't think the tests need any change. It is two line change and all tests works when I tried it. I can push it so you can review.

matsl:
> I don't understand your reference to denote here.

I understood it as you said you were pausing adding the dynamic install check for denote until this PR was merged so you would not make this work harder. Rather to be able to use hybp:install-package directly when it was ready. Maybe I misunderstood.

rswgnu (approved):

matsl:
Don't know if we can trust this but this time the build worked for all versions. PTAL

rswgnu:
I approved so you can merge when you have time to test.

matsl:
@rswgnu This was the least intrusive change I could come up with. `package-install` is adviced to retry on the specific error that we get. It is only applied for version 32.0.50 i.e will only matter for CI build and those who build their Emacs from source. And it is only applied in the unit tests when hy-test-dependencies is loaded.

Slightly sad things is that with this advice the error has never occurred. It is like [Heisenberg's indeterminacy principle](https://en.wikipedia.org/wiki/Uncertainty_principle). If we try to catch the error it is gone.

I know the PR is approved but want you view on this way to move forward since that was added after the approval.

rswgnu (approved):
Advice with retry seems like a lot but i am sure you have your reasons.  Approved.

matsl:
> Advice with retry seems like a lot but i am sure you have your reasons. Approved.

It was the leanest I could come up with that actually tries to just act when the error occurs. i.e. only in Emacs 32.0.50 or later, if package-install signals a stringp nil error (which looks like a problem in package-install rather than our code) and only after loading hy-test-dependencies. A normal user session will not se this retry advice. At least that is the intent.

PR #1017, opened 2026-07-19
Update hypb:in-string-p test

matsl:
# What

Update hypb:in-string-p tests.

# Why

Create usable tests while working on fixing different edge cases.


matsl:
@rswgnu Here is the updated tests for `hypb:in-string-p`. If you need other tests or modifications just comment in the PR and I'll try to update.

test/hypb-tests.el
```diff
@@ -85,62 +85,148 @@ See Emacs bug#74042 related to usage of texi2any."
...
+          (funcall mode)
+          (insert str)
+          (let ((pos 0)
+                (response-list '(nil nil nil nil t t t nil nil nil nil)))
+            (dolist (response response-list)
+              (setq pos (1+ pos))
+              (goto-char pos)
+              (if (not response)
+                  (progn
+                    (ert-info ((format "Test #%d: At pos %d, char '%c', expected not within \"%s\" in mode: %s"
```
matsl: @rswgnu I missed to update the "not within" part to "outside" before creating the PR. I'll wait with this until next update or review comment.

rswgnu: Please apply update when you have a chance.  Instead of 'not within', it should say, 'outside string text'.  When inside, 'inside string text'.  That way, we know that there is no claim that it is on a delimiter.  We're likely going to have to set up to test that as well after we cover the basics.
'

rswgnu (changes requested):

rswgnu:
@matsl, I suggest you move all the string tests in one "string-tests.el" file so we can work with them there easily.  I really need to see your changes here in one file to work with it easily before we merge.

rswgnu:
I have started looking at the string issues now and have a test function to add that shows when the in-string check of the Emacs (ppss) function differs from the result of hypb:in-string-p, making it easy to look at any descrepencies.

matsl:
Closed. Work has been moved over to #1028

PR #1033, opened 2026-08-06
Refactor install tests

matsl:
# What

Remove repo folders. Just store each config in a file for each repo or
install method.

* Makefile (install-all): Add install-melpa-releases target.

* install-test/melpa-releases-emacs: New repo melpa/releases.

* install-test/elpa-devel-emacs:
  install-test/elpa-emacs:
  install-test/elpaca-emacs:
  install-test/local-emacs:
  install-test/melpa-emacs:
  install-test/straight-emacs:
  install-test/tarball-emacs: Renamed and Removed Emacs 27 check.

* install-test/local-install-test.sh: Use the renamed files.

* install-test/tarball-install-local.sh: Renamed.

* install-test/MANIFEST: Use the renamed files.

# Why

Using separate folders for each repo or install method was not really
needed and created an unnecessary directory structure.

# Note

## use-package

The recent use-package definitions are not include in this refactoring. I prefer to do that in a separate PR to separate out the refactoring from the potential install changes. Maybe we also want to add more tests than just that Hyperbole was successfully installed!? 

## Further possibilities

Makefile still has targets for each install method. An alternative
could be to use a parameter. There would be less targets in the
Makefile that way. Today it looks like this:

    make install-elpa-devel

With using a parameter it would look like this:

    make install-test method=elpa-devel

### DIFF
--- a/install-test/elpa-devel-emacs
+++ b/install-test/elpa-devel-emacs
@@ -1,10 +1,8 @@
 ;; .emacs
 
-(when (< emacs-major-version 27)
-  (error "Hyperbole requires Emacs 27 or above; you are running version %d" emacs-major-version))
 (require 'package)
 (setq package-native-compile t)
-(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/"))
+(add-to-list 'package-archives '("gnu-devel" . "https://elpa.gnu.org/devel/"))
 (unless (package-installed-p 'hyperbole)
   (package-refresh-contents)
   (package-install 'hyperbole))
--- a/install-test/local-install-test.sh
+++ b/install-test/local-install-test.sh
@@ -6,15 +6,11 @@ set -e
 
 install_method="$1"
 
-app=/tmp/hypb-$$
+app=/tmp/hypb-$$/home/user
+export HOME=$app
+mkdir -p $app/.emacs.d
 
-mkdir -p $app
-cp -a * $app
-export HOME=$app/$install_method
-
-[ -e $app/$install_method/install-local.sh ] && cd $app/$install_method && ./install-local.sh
-
-cd $app/$install_method
+[ -e ${install_method}-install-local.sh ] && ./${install_method}-install-local.sh
 
 if [ -n "$2" ]
 then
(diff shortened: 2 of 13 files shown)


install-test/elpa-devel-emacs
```diff
@@ -1,7 +1,5 @@
 ;; .emacs
 
-(when (< emacs-major-version 27)
```
rswgnu, line 3: Why do you think this is unnecessary now?  What if someone runs his historical version of Emacs?

install-test/local-install-test.sh
```diff
@@ -6,28 +6,36 @@ set -e
...
+emacs --batch -l ~/.emacs.d/init.el \
       --eval '(load (expand-file-name "test/hy-test-dependencies.el" hyperb:dir))' \
       -l hypb-ert \
       --eval "(hypb-ert-require-libraries)" \
-      -f ert-run-tests-batch-and-exit
+      -f ert-run-tests-batch-and-exit || true
 
 ## Startup again interactive - check hyperbole is found
-emacs -nw --eval "(if (boundp 'hyperb:version) (kill-emacs 0) (kill-emacs 1))"
+if emacs -nw --eval "(if (boundp 'hyperb:version) (kill-emacs 0) (kill-emacs 1))"
```
rswgnu, line 36: Why do you need to start interactively to test if the hyperb:version variable exists?  Can't you test for this in batch mode or does batch skip your personal init?

install-test/melpa-releases-emacs
```diff
@@ -0,0 +1,12 @@
+;; .emacs
+
+(require 'package)
+(setq package-native-compile t)
+(add-to-list 'package-archives '("melpa-releases" . "https://releases.melpa.org/packages/"))
```
rswgnu, line 5: If we are starting to use melpa-releases rather than melpa-stable, then the 'devel' option should be melpa-snapshots rather than the older melpa.

rswgnu (changes requested):

install-test/elpa-devel-emacs
matsl: 1. This is our install test so checking the emacs version is not really needed since we won't run the install tests with an old Emacs.
2. We require Emacs 28.2 so if we should keep the check we ought to bump the version number but...
3. We declare in Package-Require that we require Emacs 28.2 which should be enough to inform users what is needed. (Users should not look in our install tests for how to setup Hyperbole.)

install-test/local-install-test.sh
matsl: This is how it was design when I first wrote it. Here I only added a message to standard out to make it more clear what the outcome was.

I can't point to any technical reason why an interactive Emacs would be a better test that Hyperbole is installed except that is how Hyperbole will be used. So it feels like a better tests.

install-test/melpa-releases-emacs
matsl: I just wanted to test how our Melpa release looked like so I added that as part of this refactoring. We have never tested "melpa stable" since it sort of was a discouraged repo from the start.

rswgnu (approved):

PR #1039, opened 2026-08-15
Build only master and with adviced package-install removed

matsl:


matsl:
This should not have occurred here. I'm testing in my own fork but it slipped through somehow!

PR #1046, opened 2026-08-29
Remove warnings

matsl:
# What

Remove warnings

* hui-mini.el (hui:menu-last-help-string): Defvar.
(hkey-help-hide): Declare function.

# Why

Warnings should be kept to a miminum. Sometimes they contain important
info like in this case with a function, hui:menu-help, also being
defined as an alias.

### DIFF
--- a/hui-mini.el
+++ b/hui-mini.el
@@ -3,7 +3,7 @@
 ;; Author:       Bob Weiner
 ;;
 ;; Orig-Date:    15-Oct-91 at 20:13:17
-;; Last-Mod:     29-Aug-26 at 18:07:05 by Mats Lidell
+;; Last-Mod:     30-Aug-26 at 10:55:42 by Mats Lidell
 ;;
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;;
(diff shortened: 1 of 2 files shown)


hui-mini.el
```diff
@@ -382,7 +384,6 @@ instead returns the one line help string for the key sequence."
 
 (defalias 'hui:menu-quit   #'hui:menu-enter)
 (defalias 'hui:menu-abort  #'hui:menu-enter)
-(defalias 'hui:menu-help   #'hui:menu-enter)
```
matsl, line 385: hui:menu-help is defined as function and an alias. This PR removes the alias. Which should be kept @rswgnu?

rswgnu: @matsl Remove this line from your PR and I will fix the conflict in an update I push myself.  I will rename the aliased function which is connected to the variable of the same name.

rswgnu (changes requested):

matsl:
Since the alias as it is then is wrong, why not merge this now and fix the new alias later?

rswgnu:
Because having the alias line there will remind me to change it.

hui-mini.el
matsl: Kept the line commented out and added a FIXME so it is not forgotten,

matsl:
I added a FIXME comment so it is not forgotten.

rswgnu (approved):

PR #1050, opened 2026-09-04
pr use package definitions in install test

matsl:
# What

Refactor install-test to only run under docker. Further let package install be
package manager dependent and let configuration use the same,
use-package declaration, for all tests.

* Makefile (install-melpa-snapshots install-melpa-releases): Use
melpas new releases and snapshots archives. For all install test
targets add debug parameter and mount man/use-package for use by the
install-test.sh script.
(install-local): Remove target.

* install-test/MANIFEST: Add elpaca-early-init and
melpa-snapshot-emacs. Remove melpa-emacs and local-emacs. Rename
local-install-test.sh to install-test.sh.

* install-test/elpaca-early-init: Early init required by elpaca.

* install-test/elpa-devel-emacs:
install-test/elpa-emacs:
install-test/elpaca-emacs:
install-test/melpa-releases-emacs:
install-test/melpa-snapshots-emacs:
install-test/straight-emacs:
install-test/tarball-emacs: Remove use-package declaration and only
keep package manager aware install.

* install-test/melpa-emacs:
install-test/local-emacs: Delete these old removed install tests.

* install-test/install-test.sh: Rename from
local-install-test.sh.  Add debug option.

* MANIFEST: Add man/use-package.

* man/use-package: New file, use-package declaration in man folder for
use in documentation and in install tests.

# Why

Running in docker isolates the install test from the local machine.
That makes it closer to a user install and avoids creating problems
on the local machine as well.

The current solution also reuses the same use-package declaration
for all tests from one file.


rswgnu (changes requested):
I am confused by all of the repetitive use-package calls. I think installation test should be run two different ways per package manager: 1. Raw installation without any use package; 2. Testing installation after running the reference use-package we provide, which should not have to be replicated.

Let's discuss all this when you have time. Just put something on the calendar and I'll likely be able to make it.

-- Bob

matsl:
> I am confused by all of the repetitive used package calls.

What repetitive calls? You mean that Emacs is started twice? The use-package definition is set in the user init.el file. Emacs is then started twice. First time does the install. The second time verifies that it still works. We had a regression at one time where Hyperbole worked in the session it was installed but not on the next run.

> I think installation test should be run two different ways per package manager: 1. Raw installation without any use package; 2. Testing installation after running the reference use-package we provide, which should not have to be replicated.

What do you mean by replicated?

The idea with the current install test is two fold:
1. Verify that the package at elpa, melpa etc can be installed.
2. Verify that the install instructions we give works. I'm thinking that the use-package definition is our new preferred way for users to install Hyperbole so that is why I have dropped the old way to install.

There can be more paths to install a package. So installing first without any use-package definition is an option. Many users will likely do that. I'm not sure the extra complexity in the test is worth it though. ensure:t in our use-package definition installs the package if it is not there. So will essentially be the same as an install followed by a configuration.

install-test/MANIFEST
```diff
@@ -1,10 +1,10 @@
 --- HYPERBOLE INSTALLATION TEST CONFIGURATIONS ---
 elpa-devel-emacs            - Elpa-devel package personal configuration
 elpa-emacs                  - Elpa package personal configuration
+elpaca-early-init           - Early init file for elpaca
 elpaca-emacs                - Elpaca package configuration (elpa-devel install)
-local-emacs                 - Straight git package from local repo and branch
-local-install-test.sh       - Run one or more test installations
-melpa-emacs                 - Melpa legacy package personal configuration
+install-test.sh             - Run one or more test installations
+melpa-snapshot-emacs        - Melpa snapshots package personal configuration
```
rswgnu: Fix name to: melpa-snapshots-emacs

MANIFEST
```diff
@@ -21,6 +21,7 @@ ChangeLog            - Summary of changes in recent GNU Hyperbole releases
 * --- DOCUMENTATION ---
 man/hy-package.el    - Hyperbole package.el installation and configuration instructions
```
rswgnu, line 22: Doesn't your additional 'use-package' file replace the configuration in man/hy-package.el; if so, update the description on that one.

rswgnu (approved):
A few minor fixes suggested but approved for when you make those, you can just merge.

MANIFEST
matsl: Yes, that is correct. My plan is to update the documentation to use the "new" use-package file but to do that in another PR. Not doing to much in each PR and partly unrelated. So man/hy-package.el is still used by the documentation so I leave it untouched here to be fixed later.

I would also want to discuss how the use-package declaration looks like and would want to do that before we update the docs.

matsl: @rswgnu OK?

rswgnu:
Yes, ok.-- BobOn Sep 14, 2026, at 4:07 PM, Mats Lidell ***@***.***> wrote:﻿



In MANIFEST:
> @@ -21,6 +21,7 @@ ChangeLog            - Summary of changes in recent GNU Hyperbole releases
 * --- DOCUMENTATION ---
 man/hy-package.el    - Hyperbole package.el installation and configuration instructions

@rswgnu OK?

PR #1059, opened 2026-09-16
pr update use package declaration use it in documentation

matsl:
# What

Update the use-package declaration and use it in the documentation.

- **Handle no debug exit gracefully with no error**
- **Update use-package declaration**
- **Use use-package declaration in documentation**

# Why

Refactoring the use-package declaration based on feedback from users.

# Note

This is a draft PR since the documentation is not updated. That will
be rearranged so that installation and configuration are more
separated.

Installation section will be made shorter to not go into details on
the different package managers. It should suffice to mention what
archives to set the package-archives to or what url the repository is
at more or less.

For the configuration the use-package declaration will be used.


rswgnu:
Looks reasonable assuming tested successfully.

man/hyperbole.texi
```diff
@@ -9186,36 +9188,36 @@ specifiable filters or from a list.
...
 
-@node Installation, Customization, Setup, Setup
-@section   Installation
+@node Installation and Configuration, Customization, Setup, Setup
+@section   Installation and Configuration
 
-There are multiple package managers you can use to install Hyperbole once
-you have GNU Emacs set up at your site.  Choose one based on your needs.
+Hyperbole can be installed using Emacs' package manager, from its
+development repository or manually from the tarball archive.
```
rswgnu: The above change looks like a regression to me.

man/hyperbole.texi
```diff
@@ -9229,83 +9231,78 @@ After installing Hyperbole, read the next section on Invocation.
...
+
+@noindent
+MELPA Releases:
+
+@code{(add-to-list 'package-archives '("melpa-releases" . "https://releases.melpa.org/packages/") t)}
+
+@noindent
+MELPA Snapshots:
 
-Now save the file and restart Emacs.  Hyperbole will then be downloaded
```
rswgnu: Snapshots is equiv to elpa-devel so it should come first and melpa releases should be in the stable release section.

man/hyperbole.texi
```diff
@@ -9229,83 +9231,78 @@ After installing Hyperbole, read the next section on Invocation.
...
+package, this is fine to use and update on a day-to-day basis.  But
+new features are tested on this branch and once in awhile it may break
+for a short time before a fix is pushed.
+
+If you prefer Hyperbole is also available from MELPA.
+
+For installing from GNU-devel ELPA, MELPA releases or MELPA snapshots
+add the respective archive to package-archives in your init
+file, see @pxref{Init File,,The Emacs Initialization File,emacs, the GNU Emacs
+Manual}.
```
rswgnu: Let's discuss this rewrite.

man/use-package
```diff
@@ -7,30 +7,36 @@
...
-   )
+  (:map hyperbole-mode-map
+        ;; Bind Smart Keys: {M-RET} is the Action Key; {C-u M-RET} is the
+        ;; Assist Key
+        (("M-RET" . hkey-either)
+         ;; Uncomment the next binding if you want to emulate Hyperbole mouse drag
+         ;; events from your keyboard.
+         ;; See "https://www.gnu.org/s/hyperbole/man/hyperbole.html#Keyboard-Drags".
+         ;; ("M-o"   . hkey-operate)
+         ))
```
rswgnu: Move the commented out binding up to the top of this sexp and add the trailing two closing parentheses to the last binding, so we don't have hanging parentheses (generally, considered bad Lisp style, although we sometimes do it for convenience).

man/use-package
```diff
@@ -39,4 +45,5 @@
   ;;   nil      - In Org mode, Org controls the key
   :custom
   (hsys-org-enable-smart-keys t)
+
   )
```
rswgnu: Move the closing paren up so it is not hanging.

rswgnu (changes requested):
Rebase off master and the trailing whitespace changes should disappear so we can consider further changes without those.  Also, see inline comments.

man/hyperbole.texi
matsl: Yes I understand that. I want to put less emphasis on choosing a packet manager. and more focus on that the user already has an Emacs capable of installing Hyperbole. So lowering the bar so to speak. Users who already use other package managers will not be confused by this.

man/hyperbole.texi
matsl: Yes, that is a bit confusing. I want ELPA to be the users first choice. What do you think of moving info about MELPA to a paragraph telling that it is possible to use MELPA as well if  user prefers that. And by that skip the bullet-list removing the mix of stable(release) and devel(snapshot) since that can be confusing?

matsl:
@rswgnu I have rebased so you don't have to configure github to ignore whitespace, changed the use-package to avoid hanging parens and added some comments to the PR for the discussion.

